PR #58, opened 2022-11-30
feat: Add staking/slashing mechanism

### LINKED ISSUE #51: Bring Staking/Slashing contracts into monorepo on feature branch
Staking/Slashing feature requires not only Staking/RewardPool contract implementation, but also Escrow/EscrowFactory contract changes.
Therefore, all the other components of Human Protocol needs to be updated accordingly. It's not limited to SDK, but also the fortune, escrow-dashboard, and other example apps.
This requires a lot of changes, so it's better to have feature branch for this functionality, so that we can make sure everything is working properly on that feature branch.

### LINKED ISSUE #62: Staking/Slashing and Slash on abuse

leric7:
Closes #51 .
Closes #62 .

- [x] Update core smart contracts.
- [x] Update Python SDK.
- [x] Update Node.js SDK.
- [x] Update Fortune.
- [x] Update subgraph if necessary.
- [x] Update ETH KVStore GUI if necessary.
- [x] Update Escrow Dashboard if necessary.


packages/core/test/Escrow.ts
```diff
@@ -2,68 +2,37 @@ import { anyValue } from '@nomicfoundation/hardhat-chai-matchers/withArgs';
...
-  launcher: Signer,
-  reputationOracle: Signer,
-  recordingOracle: Signer,
-  externalAddress: Signer,
-  restAccounts: Signer[],
-  trustedHandlers: string[];
-
-let token: HMToken, escrow: Escrow;
-
-async function deployEscrow() {
```
portuu3, line 30: Is there a reason to remove this function? It avoids duplicated code

packages/core/test/Escrow.ts
```diff
@@ -2,68 +2,37 @@ import { anyValue } from '@nomicfoundation/hardhat-chai-matchers/withArgs';
...
-  const Escrow = await ethers.getContractFactory('Escrow');
-  escrow = await Escrow.deploy(
-    token.address,
-    await owner.getAddress(),
-    100,
-    trustedHandlers
-  );
-}
-
-async function setupEscrow() {
```
portuu3, line 41: Is there a reason to remove this function? It avoid duplicated code

packages/core/test/Escrow.ts
```diff
@@ -2,68 +2,37 @@ import { anyValue } from '@nomicfoundation/hardhat-chai-matchers/withArgs';
...
-    );
-}
-
-async function addTrustedHandlers() {
-  await escrow
-    .connect(owner)
-    .addTrustedHandlers([await reputationOracle.getAddress()]);
-}
-
-async function fundEscrow() {
```
portuu3, line 60: Is there a reason to remove this function? It avoid duplicated code

packages/core/test/Escrow.ts
```diff
@@ -2,68 +2,37 @@ import { anyValue } from '@nomicfoundation/hardhat-chai-matchers/withArgs';
...
-      await reputationOracle.getAddress(),
-      await recordingOracle.getAddress(),
-      10,
-      10,
-      MOCK_URL,
-      MOCK_HASH
-    );
-}
-
-async function addTrustedHandlers() {
```
portuu3, line 54: Is there a reason to remove this function? It avoid duplicated code

packages/core/test/Escrow.ts
```diff
@@ -2,68 +2,37 @@ import { anyValue } from '@nomicfoundation/hardhat-chai-matchers/withArgs';
...
+    restAccounts: Signer[],
+    trustedHandlers: string[];
+
+  let token: HMToken, escrow: Escrow, staking: Staking, rewardPool: RewardPool;
+
+  const minimumStake = 2;
+  const lockPeriod = 2;
+  const rewardFee = 2;
+
+  beforeEach(async () => {
```
portuu3, line 35: it should be beforeAll, you don't need to deploy a new factory for each test

portuu3 (changes requested):
Tests can be ompitimized to avoid long runs for CI

packages/core/contracts/Staking.sol
```diff
@@ -0,0 +1,541 @@
...
+/**
+ * @title Staking contract
+ * @dev The Staking contract allows Operator, Exchange Oracle, Recording Oracle and Reputation Oracle to stake to Escrow.
+ */
+contract Staking is IStaking {
+    using SafeMath for uint256;
+    using Stakes for Stakes.Staker;
+
+    // Owner address
+    address public owner;
```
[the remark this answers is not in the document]
leric7: For now, we don't have any 3rd party contract dependencies. I don't think our contracts will be complicated, so we can simply keep using our own implementation of Ownable.

packages/core/test/Escrow.ts
leric7: Originally, there was deploy helper functions used for each `describe` within `before` function. Now, we use `beforeEach` instead of `beforeAll`, and I believe it's doing the same thing for testing. For both cases, we deploy all the contracts before each test case. I prefer using `beforeEach` while it makes more sense, and we don't need unnecessary helper functions to deploy. If this makes sense, we can resolve the comments above about the eliminated helper functions.

packages/core/contracts/Staking.sol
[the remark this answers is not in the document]
portuu3: should we then have a transferownership option?

packages/core/test/Escrow.ts
portuu3: using beforeAll we reduce the execution time of the tests, we have optimize this, since using a monorepo the number of tests grows a lot

packages/sdk/typescript/human-protocol-sdk/package.json
```diff
@@ -38,7 +38,7 @@
     ]
   },
   "dependencies": {
-    "@human-protocol/core": "^1.0.9",
+    "@human-protocol/core": "^1.0.11",
```
portuu3, line 41: why don't you use workspace?

packages/sdk/typescript/human-protocol-sdk/scripts/run-test.sh
```diff
@@ -1,9 +1,14 @@
 #!/bin/bash
+set -eux
```
portuu3, line 2: all this scripts does can be done from package.json, check fortune for an example

portuu3 (changes requested):

packages/sdk/typescript/human-protocol-sdk/package.json
leric7: It fails when we publish the package

packages/sdk/typescript/human-protocol-sdk/scripts/run-test.sh
leric7: Is there any benefit of having those complex commands directly in package.json? I prefer using bash script which is clean, and better to manage.

portuu3 (approved):

PR #59, opened 2022-12-01
Read bulk_payout logs to avoid double payouts

alidzm:
To avoid double payouts `bulk_payout` method has been reworked: read `bulk_payout` transaction receipt and check events of transaction to catch `Transfer` event and make sure that payout has been processed.


leric7 (dismissed):
LGTM

PR #60, opened 2022-12-02
feat: implement total supply on chain

m00n620:


portuu3:
Should these number be shown like this? Are they wei values?
image
image

portuu3 (approved):

PR #98, opened 2023-01-02
Reconfiguration of recording oracle to Deno based 

UncleTom29:


packages/examples/fortune/recording-oracle/src/contracts/EscrowAbi.json
```diff
@@ -0,0 +1,453 @@
+[
```
posix4e, line 1: Please do not checkin json of abi

posix4e (changes requested):

foufrix:
Hi, after trying out your repo, I can setup the escrow, but when I send the first message I get an axios error :
image

The two first with uncaught promise come from my extension, so you can neglect them

Those are the one important : 
image

posix4e (changes requested):
Stop checking in Escrow.json. That is not ok

packages/examples/fortune/recording-oracle/hardhat.config.js
```diff
@@ -1,4 +1,5 @@
-/** @type import('hardhat/config').HardhatUserConfig */
-module.exports = {
-  solidity: '0.6.2',
-};
+/** @type import('hardhat/config').HardhatUserConfig */
```
posix4e, line 1: why are you reformatting this file?

packages/examples/fortune/recording-oracle/LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
posix4e, line 1: Why are you checking in GPL? That's not ok

UncleTom29:
Sorry for that, it was used to test. I’ll remove it now
Moreover during deployment, there’s no way to mirror the NPM workspace init
in Deno. So there’s no way to import the ***@***.***/core” module. It
might cause the Deno app not to run properly.

UncleTom29:
@posix4e requested changes implemented

Issue #155, opened 2023-01-10
Fortune Reputation Oracle - Reputation contract


portuu3:
**Description**
Create a reputation contract

**Motivation**
We want to have a reputation system where the workers with higher reputation get higher payouts

**Implementation details** 
- [x] Function to increase/decrease workers reputation. The reputation starting value of reputation is 50 and cannot be less than 1 and more than 100. This function can only be called by addresses with an staking amount higher than a set value and a key value pair ("Role", "ReputationOracle") in KVStore.
- [x] Only contract owner can update the minimum stake necessary to become reputation oracle.
- [x] View to get the reputation value of a list of workers.
- [x] View to get the payout amount for each worker based on the total amount and the workers reputation.
- [x] Test for the new contract

PR #172, opened 2023-01-12
Feat/reputation contract

flopez7:
## Description

Create a reputation contract

## Summary of changes

- [x] Function to increase/decrease workers reputation. The reputation starting value of reputation is 50 and cannot be less than 1 and more than 100. This function can only be called by addresses with an staking amount higher than a set value and a key value pair ("Role", "ReputationOracle") in KVStore.
- [x]  Only contract owner can update the minimum stake necessary to become reputation oracle.
- [x]  View to get the reputation value of a list of workers.
- [x]  View to get the payout amount for each worker based on the total amount and the workers reputation.
- [x]  Test for the new contract

## Related issues

https://github.com/humanprotocol/human-protocol/issues/155

## Operational checklist


packages/core/contracts/Escrow.sol
```diff
@@ -213,11 +217,22 @@ contract Escrow is IEscrow {
...
+            if (status == EscrowStatuses.Pending) {
+                status = EscrowStatuses.Partial;
+                remainingFortunes = remainingFortunes.sub(_recipients.length);
+            }
+            if (
+                balance > 0 &&
+                status == EscrowStatuses.Partial &&
+                remainingFortunes == 0
+            ) {
+                _safeTransfer(canceler, balance);
```
portuu3, line 230: this should be launcher

packages/apps/escrow-dashboard/src/components/Token/TokenView.tsx
```diff
@@ -1,14 +1,11 @@
-import { Grid, Typography } from '@mui/material';
+import { Grid } from '@mui/material';
 import * as React from 'react';
 import { CardTextBlock } from 'src/components/Cards';
-import BitfinexIcon from 'src/components/Icons/BitfinexIcon';
```
portuu3, line 4: Why has this been modified?

portuu3 (changes requested):

packages/core/contracts/Escrow.sol
flopez7: Launcher address contains the escrow factory address and canceller is the owner of the escrow

packages/apps/escrow-dashboard/src/components/Token/TokenView.tsx
flopez7: I got an error from Lint because this was not used and forced me to delete to pass the tests

portuu3 (approved):

Issue #224, opened 2023-02-08
Fortune Recording Oracle - Fix bug with the substitution of already recorded answer


eugenvoronov:
**Description**
When sending fortune, a bug occurs that allows to user repeatedly overwrite an already existing fortune. It can be reproduce if the first answer is an answer containing curses or a non-unique answer, and the second and subsequent ones are valid.

PR #225, opened 2023-02-08
Fixed bug with the substitution of already recorded answer

eugenvoronov:
## Description
When sending fortune, a bug occurs that allows to user repeatedly overwrite an already existing fortune. It can be reproduce if the first answer is an answer containing curses or a non-unique answer, and the second and subsequent ones are valid.

## Summary of changes
- Added new downgrade method for worker;
- Fixed logic for adding fortune in case of an error;
- Possible types of errors when sending a fortune have been merged into one message.

## How test the changes
1. Send a fortune contains curses it will be recorder with `false` score; 
2. Send a valid fortune. It will be recorded with valid answer;
3. Send an another valid fortune. It will be rejected. 

## Related issues
[Fortune Recording Oracle - Fixed bug with the substitution of already recorded answer#224](https://github.com/humanprotocol/human-protocol/issues/224)


packages/examples/fortune/recording-oracle/src/services/recordingOracle.ts
```diff
@@ -103,54 +138,51 @@ export async function processFortunes(
...
 
   const fortuneResults = {
     escrowAddress: fortune.escrowAddress,
     chainId: escrow.chainId,
     fortunes: escrow.fortunes,
   };
+
   const fortuneResultsUrl = await saveFortuneResults(plugins, fortuneResults);
-  // console.log('Fortune Results Url: ', fortuneResultsUrl);
+  console.log('Fortune Results Url: ', fortuneResultsUrl);
```
flopez7, line 185: Remove this log

packages/examples/fortune/recording-oracle/src/plugins/storage.ts
```diff
@@ -55,10 +55,16 @@ export class Storage {
   ): IEscrowStorage {
     const escrow = store.get(escrowAddress);
 
-    escrow.fortunes[workerAddress] = {
+    console.log(escrow.fortunes);
```
flopez7, line 58: Remove this log

flopez7 (changes requested):

flopez7 (approved):

PR #371, opened 2023-04-05
Certik audit v2

### LINKED ISSUE #365: Core smart contracts - Resolve certik issues

leric7:
## Description

Certik Audit Report Fixes

## Summary of changes

- Contract Changes
  - Removed `solutionsRemaining`, `bulkPaid`, and `finalAmounts` storage variables, as they are not used.
  - Refactored `bulkPayout` function.
  - Fixed too many digit which is unreadable in `HMToken.sol`.
  - Added `constructor` to upgradeable contracts which calls `disableIntializers()` in it.
- Monorepo Changes
  - All of the `@human-protocol/core` dependencies are updated from `workspace:*` to `*`, except for `escrow-dashboard`. This prevents us to update `yarn.lock` each time we upgrade the core smart contract. `core` package will be symlinked, so that all the others are supposed to use latest version of `core` package.
  - Vercel config for `faucet-server`, `job-launcher-server`, `recording-oracle`, and `reputation-oracle` is updated. Removed unnecessary `buildCommand`, `installCommand`, and `outputDirectory`.
  - Fixed Python/Node SDK, and fortune according to smart contract change.
## Related issues

Closes #365 

## Operational checklist


packages/core/contracts/Escrow.sol
```diff
@@ -228,63 +223,46 @@ contract Escrow is IEscrow, ReentrancyGuard {
         require(aggregatedBulkAmount < BULK_MAX_VALUE, 'Bulk value too high');
 
         if (balance < aggregatedBulkAmount) {
-            return bulkPaid;
+            return false;
```
portuu3, line 226: use a require instead, otherwise we would be paying for a transaction that does nothing

packages/core/contracts/Escrow.sol
```diff
@@ -228,63 +223,46 @@ contract Escrow is IEscrow, ReentrancyGuard {
...
+        } else {
+            status = EscrowStatuses.Partial;
         }
+
         emit BulkTransfer(
             _txId,
             _recipients,
-            _amountsPaid,
+            finalAmounts,
             status == EscrowStatuses.Partial
```
portuu3, line 257: Why dont we just emit the status here instead of a boolean?

packages/core/contracts/Escrow.sol
```diff
@@ -228,63 +223,46 @@ contract Escrow is IEscrow, ReentrancyGuard {
...
-                status == EscrowStatuses.Partial &&
-                remainingSolutions == 0
-            ) {
-                _safeTransfer(canceler, balance);
-                status = EscrowStatuses.Paid;
-            }
-            if (balance == 0 && status == EscrowStatuses.Partial) {
-                status = EscrowStatuses.Paid;
-            }
+        if (balance == 0) {
```
portuu3, line 247: We might have a problem with residual balance from operations, so in this case the balance will never be 0, so the status will never turn to be paid. We need some mechanism to tell the contract that, with this call to bulkpayout the escrow status should be changed to Paid and the residual amount should be sent to the escrow owner

portuu3 (changes requested):

packages/core/contracts/Escrow.sol
leric7: I think it's a change to bulk payout logic then, and it's not something needs to be handled within this PR. We can implement that mechanism later.

packages/core/contracts/Escrow.sol
leric7: boolean field is more descriptive as an argument of this event emission.

packages/core/contracts/Escrow.sol
portuu3: Agree

portuu3 (changes requested):

packages/apps/faucet-server/vercel.json
```diff
@@ -1,7 +1,4 @@
 {
-  "buildCommand": "yarn workspace @human-protocol/faucet-server build",
-  "installCommand": "yarn install",
```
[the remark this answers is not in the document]
leric7: I do believe these two commands are actually ignored, when we use `builds` in the config.

.github/workflows/cd-core.yaml
```diff
@@ -12,9 +12,6 @@ jobs:
       - uses: actions/checkout@v3
       - run: yarn --ignore-scripts
         name: Install dependencies
-      - run: yarn compile
```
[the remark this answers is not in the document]
leric7: Added all those abis/artifacts/typechain-types in the package, so no need to compile.

portuu3 (approved):

PR #384, opened 2023-04-13
Feat/sdk storage minio

flopez7:
## Description
Change S3 client from boto3 to minio

## Summary of changes
Change S3 client from boto3 to minio

## How test the changes
Run only SDK storage tests


packages/sdk/python/human-protocol-sdk/Pipfile
```diff
@@ -14,6 +14,7 @@ black = "*"
 pylint = "*"
 pytest = "*"
 web3 = "==5.24.0"
+minio = "*"
```
Dzeranov, line 18: Need to remove `boto3` module if we are not going to use it.

PR #429, opened 2023-04-26
fix : network switch

mrhouzlane:
You can change the network later after been connected , but any network you select on **Select network** is not updated


portuu3:
@mrhouzlane please can you do it again using develop as base for your branch? exchange oracle client is in a different folder now

portuu3:
this has also been solver here https://github.com/humanprotocol/human-protocol/pull/426

PR #555, opened 2023-06-09
[CVAT][Exchange Oracle] Mock tests

flopez7:
## Description

Add mocks to tests that did not have them previously

## Summary of changes

Add mocks and modify api to return error in case of exception

## Operational checklist

### DIFF
--- a/packages/examples/cvat/exchange-oracle/src/modules/api.py
+++ b/packages/examples/cvat/exchange-oracle/src/modules/api.py
@@ -1,4 +1,4 @@
-from fastapi import APIRouter, Header
+from fastapi import APIRouter, HTTPException, Header
 from typing import Union
 
 from .api_schema import JLWebhook, JLWebhookResponse, CvatWebhook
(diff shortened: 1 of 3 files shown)


packages/examples/cvat/exchange-oracle/src/modules/api.py
```diff
@@ -22,13 +22,16 @@ def jl_webhook(
...
 
-    with SessionLocal.begin() as session:
-        webhook_id = create_webhook(session, jl_webhook, human_signature)
+        with SessionLocal.begin() as session:
+            webhook_id = create_webhook(session, jl_webhook, human_signature)
 
-    return JLWebhookResponse(id=webhook_id)
+        return JLWebhookResponse(id=webhook_id)
+    except Exception as e:
+        raise HTTPException(status_code=400, detail=str(e))
```
Dzeranov, line 34: I believe you added this try-except block to return 400 in case something went wrong during escrow validation.
But catching `Exception` here every exception will cause a 400 response and expose inner details of the error outside (for example, if database will fail to create a new entity. In this case we need to send back 500 error without any details). I suggest you here use `ValueError` instead of `Exception`.

packages/examples/cvat/exchange-oracle/src/modules/api.py
```diff
@@ -37,8 +40,11 @@ def jl_webhook(
...
-        case EventTypes.update_job.value:
-            handle_job_event(cvat_webhook)
+    try:
+        match cvat_webhook.event:
+            case EventTypes.update_task.value:
+                handle_task_update_event(cvat_webhook)
+            case EventTypes.update_job.value:
+                handle_job_event(cvat_webhook)
+    except Exception as e:
+        raise HTTPException(status_code=400, detail=str(e))
```
Dzeranov, line 50: Same here.

Dzeranov (changes requested):

Dzeranov (approved):

PR #570, opened 2023-06-15
fix: [Core] update escrow function permissions

### LINKED ISSUE #558: Check the permission of core contract functions.

leric7:
## Description

Escrow functions `bulkPayout`, `storeResult`, and `complete` needs more restricted permission.
And reputation/recording oracle should not be considered super user, in general.

## Summary of changes

- Added two more modifiers. `trustedOrReputationOracle`, `trustedOrRecordingOracle`.
- Changed modifier of three functions `bulkPayout`, `storeResult`, and `complete` of escrow contract.
- Removed `recordingOracle`, and `reputationOracle` from trusted handlers list.

## Related issues

Closes #558 

## Operational checklist

### DIFF
--- a/packages/core/contracts/Escrow.sol
+++ b/packages/core/contracts/Escrow.sol
@@ -47,6 +47,9 @@ contract Escrow is IEscrow, ReentrancyGuard {
     string public manifestUrl;
     string public manifestHash;
 
+    string public intermediateResultsUrl;
+    string public intermediateResultsHash;
+
     string public finalResultsUrl;
     string public finalResultsHash;
 
@@ -171,7 +174,12 @@ contract Escrow is IEscrow, ReentrancyGuard {
                 status == EscrowStatuses.Partial,
             'Escrow not in Pending or Partial status state'
         );
-        _storeResult(_url, _hash);
+        require(bytes(_url).length != 0, "URL can't be empty");
+        require(bytes(_hash).length != 0, "Hash can't be empty");
+
+        intermediateResultsUrl = _url;
+        intermediateResultsHash = _hash;
+
         emit IntermediateStorage(_url, _hash);
     }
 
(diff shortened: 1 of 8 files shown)


packages/core/contracts/Escrow.sol
```diff
@@ -47,6 +47,9 @@ contract Escrow is IEscrow, ReentrancyGuard {
     string public manifestUrl;
     string public manifestHash;
 
+    string public intermediateResultsUrl;
+    string public intermediateResultsHash;
```
portuu3, line 51: I wonder if hash of the files is actually necessary, what do you think?

packages/core/contracts/Escrow.sol
```diff
@@ -297,20 +307,27 @@ contract Escrow is IEscrow, ReentrancyGuard {
...
 
+    modifier trustedOrReputationOracle() {
+        require(
+            areTrustedHandlers[msg.sender] || msg.sender == reputationOracle,
+            'Address calling not trusted'
+        );
+        _;
+    }
+
+    modifier trustedOrRecordingOracle() {
```
portuu3, line 323: This modifier might not be worth since it is used only once. Check the weight of the contract, if it is lighter by removing it, we should consider to remove it

packages/examples/fortune/reputation-oracle/src/services/escrow.test.ts
portuu3, line 1: why this change?

portuu3 (changes requested):
We should update sdk and add functions if needed, like getIntermediateResultsUrl (also update docs)

packages/examples/fortune/reputation-oracle/src/services/escrow.test.ts
leric7: There is a `bulkPayout` function call, and recording oracle now can't call this function.

leric7:
@portuu3 , I resolved your comments, and added sdk functions. I don't know what you meant by updating doc, but if you were referring to updating doc for core contracts, I have another ticket for that. #560

packages/core/contracts/Escrow.sol
leric7: Agreed. Dropped `intermediateResultsHash` param. However, want to keep both function params of `storeResults` for backwards compatibility.

leric7:
> > @portuu3 , I resolved your comments, and added sdk functions. I don't know what you meant by updating doc, but if you were referring to updating doc for core contracts, I have another ticket for that. #560
> 
> I meant this doc https://www.notion.so/human-protocol/SDK-d312448eb4224d1886435e27897189e3

I updated.

portuu3 (approved):

PR #571, opened 2023-06-15
fix package dependencies and clean dead code

portuu3:
## Description
The branch had dead code and the app and tests did not work

## Summary of changes

- Clean dead code
- Fix depenencies
- Make app start
- Make test work


packages/apps/job-launcher/server/src/common/decorators/role.ts
```diff
@@ -1,6 +0,0 @@
-import { SetMetadata } from '@nestjs/common';
```
eugenvoronov, line 1: We need to return the role decorator since it is used in guards.

PR #599, opened 2023-06-24
Check env variables in job launcher

eugenvoronov:
## Description
Clean everything around env variables

- [x] Remove unused env variables from env example
- [x] Use env validator like [here](https://github.com/humanprotocol/human-protocol/blob/develop/packages/apps/fortune/exchange-oracle/server/src/common/config/env.ts)
- [x] Use constants for the env variable names


portuu3 (commented):
why there are so many changes in package.json?

eugenvoronov:
Updated Nest. Many libraries were not compatible with each other, all versions were updated. @portuu3

portuu3 (approved):

Issue #601, opened 2023-06-26
Job Launcher - Error logging


portuu3:
**Describe the bug**
We are not logging error information when we catch an exception in:
- [cron.job.ts](https://github.com/humanprotocol/human-protocol/blob/5d4e783b99bd7e5d6284d3c6d2647c7197c1254d/packages/apps/job-launcher/server/src/modules/job/job.cron.ts#L41)
- [job.repository.ts](https://github.com/humanprotocol/human-protocol/blob/5d4e783b99bd7e5d6284d3c6d2647c7197c1254d/packages/apps/job-launcher/server/src/modules/job/job.repository.ts#L70)
- [job.service.ts](https://github.com/humanprotocol/human-protocol/blob/5d4e783b99bd7e5d6284d3c6d2647c7197c1254d/packages/apps/job-launcher/server/src/modules/job/job.service.ts#L295)
- [payment.service.ts (1)](https://github.com/humanprotocol/human-protocol/blob/5d4e783b99bd7e5d6284d3c6d2647c7197c1254d/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts#L72)
- [payment.service.ts (2)](https://github.com/humanprotocol/human-protocol/blob/5d4e783b99bd7e5d6284d3c6d2647c7197c1254d/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts#L100)

**Expected behavior**
We should be logging information that tell us what was the error

PR #626, opened 2023-07-04
Updated error logging and error handling

eugenvoronov:
## Description
Now we are logging errors that were caught at runtime. Some methods were additionally wrapped in `try...catch` blocks, where there is a call to party functions and there is no certainty in the correct error handling.

Leave error handling in try...catch blocks only at entry points:

**Calling as API methods (Validation pipe handling errors occurs during API call):**
- createFortuneJob
- createCvatJob
- confirmFiatPayment
- createFiatPayment
- createCryptoPayment

**Calling as Cron methods - Using try...catch block as entry point method:**
- launchJob

**Service methods - Errors generated by these methods are handled at the top level of the methods that call them:**
- saveManifest
- getManifest
- sendWebhook
- getUserBalance
- getPayment
- createCustomer

## Summary of changes
- Added additional `try...catch`
- Added error logging where there was none

## Related issues
[Job Launcher - Error logging#601](https://github.com/humanprotocol/human-protocol/issues/601)


packages/apps/job-launcher/server/src/modules/payment/currency.service.ts
```diff
@@ -1,25 +1,31 @@
...
+  private readonly logger = new Logger(HttpService.name);
   constructor(private readonly httpService: HttpService) {}
 
   public async getRate(tokenId: TokenId, currency: Currency): Promise<number> {
-    const { data } = await firstValueFrom(
-      await this.httpService.get(
-        `${COINGECKO_API_URL}?ids=${tokenId}&vs_currencies=${currency}`,
-      ),
-    );
+    try {
```
portuu3, line 14: move this try catch to the controller

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/payment/currency.service.ts
eugenvoronov: Can you explain? I think that we should follow the rule - thin controllers and keep all the logic and error handling at the service level. There is no particular point in moving the error trapping block to the controller level. This `getRate` method will be used in other services, I suggest leaving this `try...catch` block where it is or removing it, but it cannot be in the controller.

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/payment/currency.service.ts
eugenvoronov: Imagine a situation where error handling requires an if-else condition, and when we have a lot of cases (example: https://github.com/humanprotocol/human-protocol/blob/f3f2de2c38f512eb2f0ba122b236e23a8c4a2e37/packages/sdk/typescript/human-protocol-sdk/src/escrow.ts#L138), we will write error handling logic in the controller, which is not very good.

I propose to completely remove this block, and not to take it out anywhere, since if this method is called in the controller, the error will be caught by the validation pipe. If this method is called from another service, it can be wrapped in a try-catch.

portuu3 (approved):

Issue #650, opened 2023-07-07
[Job Launcher] - [Payments] - Correctly handle conversion to HMT using rate


eugenvoronov:
**Description**
When crediting funds to the user's balance, the currency in which the deposit was made is not taken into account in any way. At the moment, all payments received in any currency require conversion to HMT token with the same rate.

**Motivation**
Accurately charge the balance to the user

**Implementation details** 
- When crediting stripe payments, credit funds to the balance with conversion to HTM
- Cryptocurrency payments are accepted only in HTM

PR #654, opened 2023-07-10
[Job Launcher] Added logic for rate processing

eugenvoronov:
## Description
When depositing funds to the balance, the currency deposit rate to USD is taken into account.

## Summary of changes
- Added account of the rate when replenishing the balance
- Added check for a token symbol when replenishing with crypto
- Added `payment` constants
- Added `payment` interface
- Updated tests

## Related issues
[[Job Launcher] - [Payments] - Correctly handle conversion to HMT using rate#650](https://github.com/humanprotocol/human-protocol/issues/650)


packages/apps/job-launcher/server/src/modules/payment/payment.dto.ts
```diff
@@ -43,18 +42,16 @@ export class PaymentCryptoCreateDto {
 export class PaymentCreateDto {
   public transactionHash?: string;
   public amount?: string;
-  public currency?: Currency;
+  public currency?: string;
```
flopez7, line 45: Why did you remove the Currency type?

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -164,14 +193,23 @@ export class PaymentService {
...
+      TokenId.HMT,
       PaymentType.DEPOSIT,
-      BigNumber.from(amount),
+      amount,
+      rate,
+      transaction.transactionHash
     );
 
     return true;
+    return true;
```
flopez7, line 212: duplicated return

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -146,11 +149,18 @@ export class JobService {
       throw new NotFoundException(ErrorJob.NotCreated);
     }
 
+    const rate = await this.currencyService.getRate(
```
portuu3, line 152: getRate is always called to be passed to savePayment, I think we can move getRate call insice saveParyment function

portuu3 (changes requested):
Check also @flopez7 comments

packages/apps/job-launcher/server/src/modules/payment/payment.dto.ts
eugenvoronov: It was incorrect a bit, we can store payment in fiat and in crypto, that means currency represent as fiat currencies and tokens symbols at the same time. I don't actually want to use double types for property, for example `currency: Currency | TokenId`, at the same time not good idea to union Currency and TokenId to the same enum type. Probably it can be improved somehow.

packages/apps/job-launcher/server/src/modules/job/job.service.ts
eugenvoronov: Good idea, but some times we are using reverse rate as here. What do you think about this case?

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -105,12 +109,19 @@ export class PaymentService {
       throw new BadRequestException(ErrorPayment.NotSuccess);
     }
 
-    await this.savePayment(
-      userId,
-      PaymentSource.FIAT,
-      PaymentType.DEPOSIT,
-      BigNumber.from(paymentData.amount),
-    );
+      const rate = 1 / await this.currencyService.getRate(
```
eugenvoronov, line 112: Reverse rate

packages/apps/job-launcher/server/src/modules/job/job.service.ts
eugenvoronov: https://github.com/humanprotocol/human-protocol/pull/654/files#r1263866867

portuu3: would swap the params `from` and `to` work?

eugenvoronov: > would swap the params from and to work?

eugenvoronov: Actually not, because of coingecko api, it will return correct response when token id it's first parameter. For example: 
✅ https://api.coingecko.com/api/v3/simple/price?ids=human-protocol&vs_currencies=usd
❌ https://api.coingecko.com/api/v3/simple/price?ids=usd&vs_currencies=human-protocol

@portuu3

portuu3: Can we have an enumerable of currencies and another one of tokens, so when the getRate function is called, depending on whether 'from' is a currency or a token, we reverse it, would that make sense?

eugenvoronov: Yes, I will implement it

portuu3 (changes requested):

portuu3 (approved):

PR #734, opened 2023-07-29
[Recording Oracle] - Fixed lint errors

eugenvoronov:
## Description
Fixed lint errors.

## Summary of changes
`yarn lint --fix` was applied.

## How test the changes
`yarn lint`


portuu3 (changes requested):
We should be using the same prettier config for all projects, don't make sense to use single quotes in some and double quotes in others. Let's make them match please

portuu3:
@eugenvoronov I would create a new PR from develop, you have many conflicts here. If you do it have in mind this https://github.com/humanprotocol/human-protocol/pull/734#pullrequestreview-1554301084

PR #740, opened 2023-08-01
Job Launcher - Refactor auth

portuu3:
## Description
Refactor authentication of job launcher, clean code and fix detected issues

## Summary of changes

- Generate automatic migration
- Remove unused env variables
- Use env variables in docker
- Remove unnecessary roles guard
- Add refresh token inside jwt
- Remove duplicated networksMap
- Add auth to all controllers that need it

> Auth endpoints fixed:

  signup
  - Before returned a JWT. It didn't make sense, after sign up, the user must verify the email. --> Now returns void
  - The user was created as active --> Now pending
  
  email-verification
  - Before returned a JWT. It didn't make sense, user should sigin using email and pass. --> Now returns void
  
  signin
  - It was not checking that the user was active
  - jwtservice was not configured properly. You don't need to pass expiration time for every sign call.
  
  refresh
  - It didn't have authentication
  
  logout
  - It didn't have authentication

## How test the changes
unit tests
`yarn test`

manual testing
`yarn docker:db:up && yarn start`


leric7 (dismissed):

packages/apps/job-launcher/server/src/database/migrations/1690813746809-InitialMigration.ts
```diff
@@ -0,0 +1,178 @@
...
+            ALTER TABLE "hmt"."token"
+            ADD CONSTRAINT "FK_e50ca89d635960fda2ffeb17639" FOREIGN KEY ("user_id") REFERENCES "hmt"."user"("id") ON DELETE NO ACTION ON UPDATE NO ACTION
+        `);
+    await queryRunner.query(`
+            ALTER TABLE "hmt"."auth"
+            ADD CONSTRAINT "FK_9922406dc7d70e20423aeffadf3" FOREIGN KEY ("user_id") REFERENCES "hmt"."user"("id") ON DELETE NO ACTION ON UPDATE NO ACTION
+        `);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
```
eugenvoronov, line 128: Drop schema here await `queryRunner.dropSchema(NS);`

packages/apps/job-launcher/server/src/database/migrations/1690813746809-InitialMigration.ts
```diff
@@ -0,0 +1,178 @@
+import { NS } from 'src/common/constants';
```
eugenvoronov, line 1: Use relative path `import { NS } from "../../common/constants";`

packages/apps/job-launcher/server/src/common/interfaces/payment.ts
eugenvoronov, line 1: Need to be return, It used here `job-launcher/server/src/common/constants/payment.ts`.
I don't see this import there, but it need to be imported to `constants/payment.ts`.

packages/apps/job-launcher/server/src/modules/auth/auth.controller.ts
```diff
@@ -3,84 +3,83 @@ import {
   ClassSerializerInterceptor,
```
eugenvoronov, line 3: Can you explain, why did you remove `refreshToken` logic?

eugenvoronov (changes requested):

portuu3:
@CodiumAI-Agent /review

packages/apps/job-launcher/server/src/modules/auth/auth.controller.ts
portuu3: It has not been removed. The refresh token has been included inside the jwt. As far as I understood the refresh token was used to get a new jwt and to set the jwt expired (logout). For those two actions you need to authenticate the user through the jwt, so it can be included in it

packages/apps/job-launcher/server/src/modules/payment/payment.entity.ts
```diff
@@ -1,20 +1,21 @@
...
+import { PaymentSource, PaymentType } from '../../common/enums/payment';
 import { UserEntity } from '../user/user.entity';
 
-@Entity({ schema: NS, name: 'payment' })
+@Entity({ schema: NS, name: 'payments' })
+@Index(['chainId', 'transaction'], {
+  unique: true,
+  where: '(chain_Id IS NOT NULL)',
+})
+@Index(['transaction'], { unique: true, where: '(chain_Id IS NULL)' })
```
eugenvoronov, line 12: Take care, because some payment records don't have any transaction id, for example the we want to withdraw some amount from user balance.

eugenvoronov (changes requested):

PR #743, opened 2023-08-03
Job Launcher - Database and payment fixes

portuu3:
## Description
Fixes related to database and payments

## Summary of changes

- Add unique index [chainId, escrowAddress] to jobs table to avoid duplicates.
- Set retriescount default value in jobs table, before it failed because of null value.
- Add chainId to payments table, otherwise you don't know from which network is the transaction.
- Add unique index [chainId, transaction] to payments table to avoid duplicates. (crypto payments). Without this we might have transaction reuse hack.
- Add unique index [transaction(paymentId from stripe)] to payments table to avoid duplicates. (fiat payments). Without this we might have id reuse hack.
- CreateFiatPayment was failing because the params passed to stripe were wrong. There is no need to pass stripe customerId since this is just for recurrent payments.
- Remove stripeCustomerId from db.
- Remove unncessary function getPayment from stripe, it just contained a line of code with a call to stripe
- Add a check so that crypto payments contain chainid.
- Change some catch erros so it gives more information like the service name, function name and exception message

Tests to be fixed in different PR


packages/apps/job-launcher/server/src/common/constants/errors.ts
```diff
@@ -25,6 +25,7 @@ export enum ErrorEscrow {
 export enum ErrorUser {
   NotFound = 'User not found',
   AccountCannotBeRegistered = 'Account cannot be registered',
+  BalanceCouldnotBeRetreived = 'User balance could not be retrieved',
```
eugenvoronov, line 28: BalanceCouldnotBeRetreived -> BalanceCouldNotBeRetreived

eugenvoronov (changes requested):

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -78,21 +74,16 @@ export class PaymentService {
   }
 
   public async createFiatPayment(
-    customerId: string,
+    user: UserEntity,
```
eugenvoronov, line 77: Where are you using this `user` object here?

eugenvoronov (changes requested):

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
eugenvoronov: Let’s keep the user inside create payment intent I will use it for the next issue related with client secret

Issue #792, opened 2023-08-16
[Job Launcher] Implement signature verification


eugenvoronov:
**Description**
Implement signature verification mechanism between services. When the webhook endpoint will be called, the received payload must be signed and the signature verified.

**Motivation**
Process requests received from trusted services.

**Implementation details** 
- Implement verify signature mechanism based on [solution](https://github.com/humanprotocol/human-protocol/blob/develop/packages/examples/cvat/exchange-oracle/src/validators/signature.py).

PR #800, opened 2023-08-21
[Job Launcher] Implemented signature verification mechanism

eugenvoronov:
## Description
Implemented signature verification mechanism between services. When the webhook endpoint will be called, the received payload must be signed and the signature verified.

## Summary of changes
- Implemented verify signature mechanism
- Implemented unit tests

## How test the changes
`yarn test`

## Related issues
[[Job Launcher] Implement signature verification#792](https://github.com/humanprotocol/human-protocol/issues/792)


packages/apps/job-launcher/server/src/common/utils/signature.ts
```diff
@@ -0,0 +1,41 @@
...
+    throw new ConflictException(
+      ErrorSignature.InvalidSignature);
+  }
+}
+
+export async function signMessage(message: string, privateKey: string): Promise<string> {
+  const wallet = new ethers.Wallet(privateKey);
+
+  const messageBytes = ethers.utils.toUtf8Bytes(message);
+  const messageHash = ethers.utils.hashMessage(JSON.stringify(messageBytes));
```
flopez7, line 37: Why did you hash the message? I think it is not necessary

cc: @portuu3

eugenvoronov: Yes, it's generally recommended to hash the message before signing for several reasons:
1. Hashing ensures that the message has a fixed length, regardless of the original message's length. This is important because the size of the data being signed affects the computational complexity of the signing process. Hashing provides a consistent length for all messages.
2. If we dont hash the data before signing we cannot have one consistent signature algorithm, because we could only sign messages up to a certain size and if the size of the message gets too large you would need to hash.
@flopez7 @portuu3

eugenvoronov: https://github.com/humanprotocol/human-protocol/blob/96ca12a2ff2bd5e1225e9b03789b87849f50cc45/packages/examples/cvat/exchange-oracle/src/chain/web3.py#L58
This version of the algorithm also uses hashing.

eugenvoronov: https://web3py.readthedocs.io/en/stable/web3.eth.account.html#sign-a-message

flopez7: This function does not generate a hash as such. It is a previous step to the signature that must be done in the python library but in ethers it is not necessary. If you try to verify a signed message in python you will see that if you use the hash the verification will fail. Here is an example signed in Python:
account -> `0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266`
msg -> `message`
signature -> `0x63fde9fec5d1924c8837bae8f19c632291725fb94bb03fb3e8d89bf6de17f52014e402e5769d27989a73e889c9aa35c7ace790d2b239d8e1d9d07046ae2d44f51c`

@portuu3 @Dzeranov

flopez7 (dismissed):

Dzeranov (approved):
lgtm

packages/apps/job-launcher/server/src/common/utils/signature.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (signer.toLowerCase() !== address.toLowerCase()) {
+    throw new ConflictException(
+      ErrorSignature.SignatureNotVerified);
+  }
+
+  return true;
+}
+
+export async function signMessage(message: object | string, privateKey: string): Promise<string> {
+  if (typeof(message) !== "string") { message = JSON.stringify(message); }
```
flopez7, line 22: Add a test for this new functionality

packages/apps/job-launcher/server/src/common/utils/signature.ts
```diff
@@ -0,0 +1,39 @@
...
+  
+  const wallet = new ethers.Wallet(privateKey);
+  const signature = await wallet.signMessage(message);
+
+  return signature;
+}
+
+export function recoverSigner(message: object | string, signature: string): string {  
+  if (typeof(message) !== "string") { message = JSON.stringify(message); }
```
flopez7, line 31: Add a test for this new functionality

flopez7 (changes requested):

flopez7 (approved):

Dzeranov (approved):

Issue #789, opened 2023-08-16
[Job Launcher] Job cancellation


eugenvoronov:
**Description**
Create a job cancelation endpoint for the job requester.

**Motivation**
Job requester must be able to cancel the created job.

**Implementation details** 
- Add new endpoint to job controller.
- Add new logic to job server.
- Send request to the escrow using SDK to cancel.
- Send webhook to exchange oracle
- Add unit tests

PR #805, opened 2023-08-22
[Job Launcher] Cancel job

eugenvoronov:
## Description
Created a job cancelation endpoint for the job requester.

## Summary of changes
- Added new cancel job endpoint to `JobController`.
- Added new `cancelJob` method to `JobService`.
- Added SDK call to `cancel` method.
- Added exchange oracle webhook call.
- Added unit tests

## How test the changes
`PATCH /job/cancel/:id` 
`yarn test`

## Related issues
[[Job Launcher] Job cancellation#789](https://github.com/humanprotocol/human-protocol/issues/789)


packages/apps/job-launcher/server/src/common/enums/job.ts
```diff
@@ -4,6 +4,7 @@ export enum JobStatus {
   LAUNCHED = 'LAUNCHED',
   COMPLETED = 'COMPLETED',
```
portuu3, line 5: I think in database we are not going to have COMPLETED status, so we can remove this

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -241,6 +242,30 @@ export class JobService {
...
+    if (!jobEntity) {
+      this.logger.log(ErrorJob.NotFound, JobService.name);
+      throw new NotFoundException(ErrorJob.NotFound);
+    }
+    
+    if (jobEntity.status === JobStatus.LAUNCHED && jobEntity.escrowAddress) {
+      const signer = this.web3Service.getSigner(jobEntity.chainId);
+
+      const escrowClient = await EscrowClient.build(signer);
+      await escrowClient.cancel(jobEntity.escrowAddress)
```
portuu3, line 260: We cannot call the escrow contract at request time because the blockchain might take some time to process the transaction and this endpoint could timeout. We need a cron job, or include the logic in the exisitng one

portuu3 (changes requested):

packages/apps/job-launcher/server/src/common/enums/job.ts
eugenvoronov: For the current implementation, I agree, but this status will be used when we start tracking the escrow with the status [EscrowStatuses.Paid](https://github.com/humanprotocol/human-protocol/blob/c7648cd7b48b0cc086c9de9bcc4ff3b877a99477/packages/core/contracts/interfaces/IEscrow.sol#L6), at least all completed jobs on the launcher side will be have [LAUNCHED](https://github.com/humanprotocol/human-protocol/blob/c7648cd7b48b0cc086c9de9bcc4ff3b877a99477/packages/apps/job-launcher/server/src/common/enums/job.ts#L1) status. Any ideas?

portuu3: I would not use status from database apart from the launching process. For the rest I would use subgraph data.

packages/apps/job-launcher/server/src/modules/job/job.service.ts
Dzeranov: Also we'll need to notify an exchange oracle that escrow is cancelled.

packages/apps/job-launcher/server/src/modules/job/job.controller.ts
```diff
@@ -40,4 +42,12 @@ export class JobController {
   public async getResult(@Request() req: any): Promise<any> {
     return this.jobService.getResult(req.user?.id);
   }
+
```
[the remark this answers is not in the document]
eugenvoronov: It actually depends on the REST spec for endpoints that make some changes, it's better to use PATCH. POST is recommended for endpoints that create new entities. We can simplify it, and use POST for update and create requests.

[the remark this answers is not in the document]
Dzeranov: @portuu3 as far as I can see, this is the endpoint to set a new status `TO_CANCEL`. So I don't understand what's the problem with the CVAT oracles here.
I mean, an actual webhook call to Exchange Oracle will be sent once escrow is cancelled.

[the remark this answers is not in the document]
portuu3: Forget it, I mixed with the failed webhook, is job requester who sends this from the frontend. I said nothing

packages/apps/job-launcher/server/src/modules/job/job.service.spec.ts
```diff
@@ -55,6 +55,7 @@ import { HMToken__factory } from '@human-protocol/core/typechain-types';
 import { RoutingProtocolService } from './routing-protocol.service';
 import { PaymentRepository } from '../payment/payment.repository';
 import { div, mul } from '../../common/utils/decimal';
+import { EventType } from '../../common/enums/webhook';
```
portuu3, line 58: seems like you didn't push this file

packages/apps/job-launcher/server/src/modules/job/job.dto.ts
```diff
@@ -9,9 +9,11 @@ import {
   IsDate,
   IsOptional,
   IsObject,
+  IsNumberString
 } from 'class-validator';
 import { ChainId } from '@human-protocol/sdk';
 import { JobRequestType, JobStatus } from '../../common/enums/job';
+import { EventType } from 'src/common/enums/webhook';
```
portuu3, line 16:
```suggestion
import { EventType } from '../../common/enums/webhook';
```

packages/apps/job-launcher/server/src/modules/job/job.service.spec.ts
```diff
@@ -667,6 +670,187 @@ describe('JobService', () => {
     });
   });
 
+  describe('requestToCancelJob', () => {
+    const jobId = 1;
+    const userId = 123;
+
+    it('should to cancel the job', async () => {
```
portuu3, line 677:
```suggestion
    it('should cancel the job', async () => {
```

packages/apps/job-launcher/server/src/modules/job/job.service.spec.ts
```diff
@@ -667,6 +670,187 @@ describe('JobService', () => {
...
+      const escrowAddress = '0xValidEscrowAddress';
```
portuu3, line 678:
```suggestion
      const escrowAddress = MOCK_ADDRESS;
```

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -261,18 +262,78 @@ export class JobService {
...
     }
 
     return jobEntity;
   }
 
+  public async requestToCancelJob(
+    userId: number,
+    id: number,
+  ): Promise<boolean> {
+    const jobEntity = await this.jobRepository.findOne({ id, userId });
```
portuu3, line 282: We need to check that the Escrow is in a status that allows to cancel it:

- balance >0
- status != Complete
- status != Paid

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -261,18 +262,78 @@ export class JobService {
...
+    if (jobEntity.escrowAddress) {
+      const signer = this.web3Service.getSigner(jobEntity.chainId);
+
+      const escrowClient = await EscrowClient.build(signer);
+      await escrowClient.cancel(jobEntity.escrowAddress)
+    }
+
+    jobEntity.status = JobStatus.CANCELED;
+    await jobEntity.save();
+
```
portuu3, line 308: the code above can be moved inside the `if (jobEntity.escrowAddress)`, if the escrow doesn't have an address, the exchange oracle is doesn't know about this job. Also in the webhook you are passing the escrowAddress which is undefined

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -261,18 +262,78 @@ export class JobService {
...
+    jobEntity.retriesCount = 0;
+    await jobEntity.save();
+    
+    return true;
+  }
+
+  public async cancelJob(
+    jobEntity: JobEntity
+  ): Promise<boolean> {
+    if (jobEntity.escrowAddress) {
```
portuu3, line 299: If the escrow has been created yet we need to return the payment to the job requester

portuu3 (changes requested):

PR #815, opened 2023-08-22
Job Launcher - fix bugs of job launcher

portuu3:
## Description

Erros fixed after manual testing

## Summary of changes

1. Fix in getrates for cases where from and to are the same. For example for usd to usd coingecko return 0.9... instead of 1.
2. Don't send verification email to users that are not in pending status.
3. Fix getResult endpoint. Previous implementation didn't make sense.
4. Fix math round in create fiat payment. Cannot round "randomly". Use decimals library and Math.Ceil to round up.
5. Save payment id instead of payment secret when creating a fiat payment.
6. In createCryptoPayment and chainId to findOne call, since we might have the same transaction hash in different chains.
7. Create job endpoint is considering that the fundAmount is HMT, but it should be USD.

## How test the changes

yarn test

## Operational checklist

### DIFF
--- a/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
+++ b/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
@@ -131,6 +131,8 @@ export class PaymentService {
       userId,
       transaction: data.paymentId,
       status: PaymentStatus.PENDING,
+      amount: div(paymentData.amount_received, 100),
+      currency: paymentData.currency,
     });
 
     if (!paymentEntity) {
--- a/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
+++ b/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
@@ -32,7 +32,7 @@ import {
 import { Web3Service } from '../web3/web3.service';
 import { CoingeckoTokenId } from '../../common/constants/payment';
 import { getRate } from '../../common/utils';
-import { add, mul } from '../../common/utils/decimal';
+import { add, div, mul } from '../../common/utils/decimal';
 
 @Injectable()
 export class PaymentService {
@@ -70,8 +70,9 @@ export class PaymentService {
   ): Promise<string> {
     const { amount, currency } = dto;
 
+    const amountInCents = Math.ceil(mul(amount, 100));
     const params: Stripe.PaymentIntentCreateParams = {
-      amount: mul(amount, 100),
+      amount: amountInCents,
       currency: currency,
     };
 
@@ -97,13 +98,13 @@ export class PaymentService {
       throw new BadRequestException(ErrorPayment.TransactionAlreadyExists);
     }
 
-    const rate = await getRate(Currency.USD, currency);
+    const rate = await getRate(currency, Currency.USD);
 
     await this.paymentRepository.create({
       userId,
       source: PaymentSource.FIAT,
       type: PaymentType.DEPOSIT,
-      amount,
+      amount: div(amountInCents, 100),
       currency,
       rate,
       transaction: paymentIntent.id,
@@ -228,14 +229,14 @@ export class PaymentService {
       throw new BadRequestException(ErrorPayment.TransactionAlreadyExists);
     }
 
-    const rate = await getRate(Currency.USD, tokenId);
+    const rate = await getRate(tokenId, Currency.USD);
 
     await this.paymentRepository.create({
       userId,
       source: PaymentSource.CRYPTO,
       type: PaymentType.DEPOSIT,
-      amount: mul(amount, rate),
-      currency: TokenId.HMT,
+      amount: amount,
+      currency: tokenId,
       rate,
       chainId: dto.chainId,
       transaction: dto.transactionHash,
@@ -252,7 +253,7 @@ export class PaymentService {
     });
 
     const totalAmount = paymentEntities.reduce((total, payment) => {
-      return add(total, payment.amount);
+      return add(total, mul(payment.amount, payment.rate));
     }, 0);
 
     return totalAmount;
--- a/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
+++ b/packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
@@ -70,10 +70,8 @@ export class PaymentService {
   ): Promise<string> {
     const { amount, currency } = dto;
 
-    const amountInCents = Math.round(amount * 100);
-    
     const params: Stripe.PaymentIntentCreateParams = {
-      amount: amountInCents,
+      amount: mul(amount, 100),
       currency: currency,
     };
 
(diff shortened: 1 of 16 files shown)


packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -219,6 +217,7 @@ export class PaymentService {
 
     const paymentEntity = await this.paymentRepository.findOne({
       transaction: transaction.transactionHash,
+      chainId: dto.chainId,
```
eugenvoronov, line 220: In createCryptoPayment and chainId to findOne call, since we might have the same transaction hash in different chains.
- No, it is not possible to have the same transaction hash in different chains. The transaction hash is generated using the transaction data and the specific rules of a blockchain network’s consensus mechanism, cryptographic hashing algorithms, and other factors. As a result, the same transaction data will always produce the same hash within a specific blockchain network.
- Since, in a hypothetical scenario it is possible for two transactions on different chains to have the same hash, but only if they were created with the same data at exactly the same time. However, this situation will be extremely rare and is a consequence of collisions in cryptographic hash functions, which are supposed to generate unique hashes for different data.

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -70,10 +70,8 @@ export class PaymentService {
   ): Promise<string> {
     const { amount, currency } = dto;
 
-    const amountInCents = Math.round(amount * 100);
-    
     const params: Stripe.PaymentIntentCreateParams = {
-      amount: amountInCents,
+      amount: mul(amount, 100),
```
eugenvoronov, line 74: it's not working.
Here will be Invalid integer error, I have described this problems already. It happens because Stripe’s API usually requires integer values for amounts and quantities. For example, an amount of $56.99 would be represented as 5699 in cents.

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -253,7 +252,7 @@ export class PaymentService {
     });
 
     const totalAmount = paymentEntities.reduce((total, payment) => {
-      return add(total, mul(payment.amount, payment.rate));
+      return add(total, payment.amount);
```
eugenvoronov, line 255: In getUserBalance we were multiplying the amount in dollar with the token rate, so we were mixing dollar amounts with token amounts.
- Be sure to record payment amounts in the currency of the transaction without additional multiplication by the exchange rate.
- This is incorrect logic, since we need to know the resulting balance of the user, given the fact that we are working with unchanged transaction data.

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -229,13 +228,13 @@ export class PaymentService {
...
 
-    const rate = await getRate(Currency.USD, TokenId.HMT);
+    const rate = await getRate(Currency.USD, tokenId);
 
     await this.paymentRepository.create({
       userId,
       source: PaymentSource.CRYPTO,
       type: PaymentType.DEPOSIT,
-      amount,
+      amount: mul(amount, rate),
```
eugenvoronov, line 237: In createCryptoPayment we were saving in paymentEntity.amount the amount in tokens and this field only should save USD.
- When recording new payments in the database, it's vital to retain the original account value prior to any currency conversions. This ensures the integrity and precision of the recorded data.

eugenvoronov (changes requested):

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
portuu3: Can you show an example where it does not work?
Request 
{
  "amount": 56.99,
  "currency": "usd"
}
Stripe
image
Database
image

eugenvoronov: I will share

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
portuu3: There something confusing here:

- As far as I understood amount field in payments table always stores USD values, since the balance we show to users is USD this would make sense to me. Might have missunderstood this.
- If I was wrong with 1., there is something wrong in create fortune job endpoint. The fundAmount value should be USD, and following your logic we store it as hmt (look screenshots taken with the code before this pr, payment id=4)

image
image

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
portuu3: related to this https://github.com/humanprotocol/human-protocol/pull/815#discussion_r1302525476

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
portuu3: If there is a minimal chance to happen we need to have it mind

eugenvoronov: Agree, just to be informed

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
eugenvoronov: Agree with screenshot, because this topic related with user balance. If the user what to add any fiat to his account he can do it: usd, eur and etc. And we need to record it as original data with original topup currency and rate relative to usd.
The same for crypto deposit. 
Since we are converting all amounts to usd and showing to the user as a balance. If the user want to deposit escrow, we are withdrawing an amount in the currency in which the escrow was issued, in our case it is hmt, and we store these values ​​in the database in relation to the currency used by the user, in our case it is usd.

eugenvoronov: let's make this logic clear.

eugenvoronov: the amounts are saved in the original currency in which the user made, for this reason we save the rate.

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
eugenvoronov: @portuu3 
```{
  "amount": 0.600002022002020,
  "currency": "eur"
}
{
  "statusCode": 400,
  "message": "Invalid integer: 60.000202200202"
}

{
  "amount": 56.0099,
  "currency": "usd"
}
{
  "statusCode": 400,
  "message": "Invalid integer: 5600.99"
}

{
  "amount": 4.199,
  "currency": "usd"
}
{
  "statusCode": 400,
  "message": "Invalid integer: 419.9"
}```

eugenvoronov: This error occurs when passing to the stripe the amount of the containing precision more than 2 decimal places.

eugenvoronov (approved):

Issue #875, opened 2023-09-07
[Reputation Oracle] [CVAT] Recording Oracle incoming webhook processing


eugenvoronov:
**Description**
Change behaviour of Recording Oracle incoming webhook processing

**Implementation details** 
1. Update `getIntermediateResults`, move `getIntermediateResultsUrl` logic out.
2. Download `resulting_annotations.zip` file from Recording Oracle bucket.
3. Upload `resulting_annotations.zip` file to Reputation Oracle bucket.
4. Parse `validation_meta.json` file and check results 
```
if quality >= min_quality then do payout eq job_bounty from manifest
```

PR #877, opened 2023-09-08
[Reputation Oracle] Enhanced webhook processing logic

eugenvoronov:
## Description
Enhanced webhook processing logic and changed behavior of Recording Oracle incoming webhook processing

## Summary of changes
- Added the ProcessingResultDto to represent the results of the processFortune and processImageLabelBinary methods consistently.
- Enhanced method annotations for clarity on functionalities, input parameters, return values, and possible exceptions.
- Moved out the logic from getIntermediateResultsUrl to streamline the process.
- The resulting_annotations.zip file is now downloaded from the Recording Oracle bucket and subsequently uploaded to the Reputation Oracle bucket.
- Implemented parsing for the validation_meta.json file.
- If the quality of results is greater than or equal to the `min_quality`, a payout equivalent to the `job_bounty` from the manifest is executed.
- Improved error messages to enhance the debugging process.

## How test the changes
In Progress

## Related issues
[[Reputation Oracle] [CVAT] Recording Oracle incoming webhook processing#875](https://github.com/humanprotocol/human-protocol/issues/875)


packages/apps/reputation-oracle/server/src/modules/webhook/webhook.service.ts
```diff
@@ -100,141 +110,151 @@ export class WebhookService {
...
       }
+  
+      await this.webhookRepository.updateOne({ id: webhookEntity.id }, {
+        resultsUrl: results.url,
+        checkPassed: results.checkPassed,
+        status: WebhookStatus.PAID,
+        retriesCount: 0,
+      });
+  
+      await escrowClient.bulkPayOut(webhookEntity.escrowAddress, results.recipients, results.amounts, results.url, results.hash);
```
eugenvoronov, line 151: Add double check for getBalance >= sum(amouns)

packages/apps/reputation-oracle/server/src/modules/webhook/webhook.service.ts
```diff
@@ -100,141 +110,151 @@ export class WebhookService {
...
-        hash,
-      );
+  /**
+   * Processes an IMAGE_LABEL_BINARY manifest type. It retrieves annotations, calculates payouts 
+   * for qualified annotators, and processes storage tasks.
+   * @param manifest The CVAT manifest data.
+   * @param intermediateResultsUrl The URL to retrieve intermediate results.
+   * @returns {Promise<ProcessingResultDto>} Returns the processing results including recipients, amounts, and storage data.
+   */
+  private async processImageLabelBinary(manifest: CvatManifestDto, intermediateResultsUrl: string): Promise<ProcessingResultDto> {
```
eugenvoronov, line 187: Rename processingCvat

packages/apps/reputation-oracle/server/src/common/utils/index.ts
```diff
@@ -0,0 +1,144 @@
...
+
+                return {
+                    key,
+                    url: `${clientParams.useSSL ? 'https' : 'http'}://${
+                        clientParams.endPoint
+                    }${clientParams.port ? `:${clientParams.port}` : ''}/${bucket}/${key}`,
+                    hash: fileHash,
+                };
+            } catch (e) {
+                throw new Error('Error uploading file');
```
simulacrum6, line 140: Iiuc, there is no way to know which file caused the problem, right? Could we include additional information in the error message that allows to identify which file failed to upload?

```suggestion
                let message = '';
                if (fileInput.type === 'path') {
                    message = `Error uploading file from path: ${fileInput.path}`);
                else {
                    message = `Error upload in-memory file. SHA1 hash: ${fileHash}`
                }
                throw new Error(message);
```

packages/apps/reputation-oracle/server/src/common/constants/index.ts
```diff
@@ -1,7 +1,14 @@
+import { JobRequestType } from "../enums";
+
 export const NS = 'hmt';
 export const RETRIES_COUNT_THRESHOLD = 3;
 export const INITIAL_REPUTATION = 0;
 
+export const CVAT_RESULTS_ANNOTATIONS_FILENAME = 'resulting_annotations.zip'
+export const CVAT_VALIDATION_META_FILENAME = 'validation_meta.json'
```
leric7, line 8: Are these file names always supposed to be `resulting_annotations.zip`, and `validation_meta.json`?

packages/apps/reputation-oracle/server/src/common/utils/index.ts
```diff
@@ -0,0 +1,161 @@
...
+        data.pipe(stream);
+
+        await new Promise((resolve, reject) => {
+            client.putObject(destBucket, key, stream, (err, etag) => {
+                if (err) reject(err);
+                else resolve(etag);
+            });
+        });
+
+        console.log(`File from ${url} copied to ${destBucket}/${key}`);
```
leric7, line 45: Remove console.log, and use logger instead.

packages/apps/reputation-oracle/server/src/common/utils/index.ts
```diff
@@ -0,0 +1,161 @@
...
+   * @param {StorageCredentials} credentials - Optional. Cloud storage access data. If credentials is not provided - use an anonymous access to the bucket
+   * @returns {Promise<UploadFile>} - Uploaded file with key/hash
+   */
+export async function uploadFiles(
+    files: InputFile[],
+    bucket: string,
+    clientParams: StorageParams,
+    credentials?: StorageCredentials,
+): Promise<UploadFile[]> {
+    const client: Minio.Client = new Minio.Client({
```
leric7, line 105: I think creating Minio client everytime we need to interact with cloud storage is not good from connection perspective. Doesn't it really help, if we have singleton of StorageClient, and then use that one to upload/download files? I think that way we can better manage the connection to storage cloud. I'm not entirely sure, though.

leric7 (changes requested):

packages/apps/reputation-oracle/server/src/common/constants/index.ts
eugenvoronov: For now yes, a bit further it will be changed @leric7

packages/apps/reputation-oracle/server/src/common/utils/index.ts
eugenvoronov: Agree

packages/apps/reputation-oracle/server/src/common/utils/index.ts
eugenvoronov: This implantation was added as a temporary solution until SDK Storage is improved. I agree with you, it is better to have one connection somewhere at the global level when initializing the application.

packages/apps/reputation-oracle/server/src/common/utils/index.ts
```diff
@@ -0,0 +1,160 @@
...
+/**
+   * **Upload file to cloud storage**
+   *
+   * @param {File[]} files - Files to upload
+   * @param {string} bucket - Bucket name
+   * @param {StorageParams} params - Cloud storage params
+   * @param {StorageCredentials} credentials - Optional. Cloud storage access data. If credentials is not provided - use an anonymous access to the bucket
+   * @returns {Promise<UploadFile>} - Uploaded file with key/hash
+   */
+export async function uploadFiles(
```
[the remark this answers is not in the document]
flopez7: Sure, remove it

packages/apps/reputation-oracle/server/src/common/utils/index.ts
simulacrum6: If creating the client is a problem, would passing a client instance be a solution?

packages/apps/reputation-oracle/server/src/modules/webhook/webhook-incoming.entity.ts
```diff
@@ -2,7 +2,7 @@ import { Column, Entity } from 'typeorm';
 
 import { NS } from '../../common/constants';
 import { BaseEntity } from '../../database/base.entity';
-import { WebhookStatus } from '../../common/enums';
+import { EventType, WebhookStatus } from '../../common/enums';
```
flopez7, line 5: EventType should be added to the WebhookIncomingEntity?

flopez7 (changes requested):

packages/apps/reputation-oracle/server/src/modules/webhook/webhook-incoming.entity.ts
eugenvoronov: At the moment I don't think we need it because it's just one type of event and we don't plan to have more.

flopez7 (approved):

leric7 (approved):

PR #898, opened 2023-09-18
[Job Launcher] - fix fiat pay form UI bug

m00n620:
## Description

Currently there's a bug on fiat pay form to disable Stripe card elements.

## Summary of changes

Fixed the condition to disable card elements, especially for checking if the checkbox is selected or not.


portuu3 (changes requested):
When I choose pay with balance and the fields get disabled, Name on card is still enabled

portuu3 (approved):

PR #913, opened 2023-09-20
[SDK][Typescript][Python] Add network array to getEscrows filter

flopez7:
## Description

Add network array to getEscrows filter to retrieve information from all networks

## How test the changes

`yarn test` for Typescript and `make run-test` for Python

## Operational checklist


leric7:
We need to update SDK doc on gitbook as well.

leric7 (changes requested):
We have /examples directory for both Python/Typescript SDK. Let's update example, as well. This way, you can make sure your updates are working end-to-end. In addition, we need to update SDK docs on GitBook.

flopez7:
> We have /examples directory for both Python/Typescript SDK. Let's update example, as well. This way, you can make sure your updates are working end-to-end. In addition, we need to update SDK docs on GitBook.

@leric7 I have it in mind, I was waiting for the approval to update the documentation only once.

portuu3 (approved):

Issue #843, opened 2023-08-29
[Job Launcher] Refund in cancellation


portuu3:
**Description**
Job Launcher needs to refund the remaining balance in the contract we job requester asks to cancell an escrow.

**Implementation details** 
Create new payment type called REFUND

1. Cancel the escrow.
2. From logs get the amount received back from the escrow contract.
3. Create a REFUND payment for the Job Requester.

PR #928, opened 2023-09-25
[WIP] [Job Launcher] Added refund payment logic

eugenvoronov:
## Description
Job Launcher needs to refund the remaining balance in the contract we job requester asks to cancel an escrow.

## Summary of changes
1. Canceled the escrow.
2. Created new payment type called REFUND.
3. Created new job type called TO_REFUND.
4. From logs got the amount received back from the escrow contract.
5. Created a REFUND payment for the Job Requester.
6. Updated job cancelation logic.
7. Added payment refund logic.
8. Track amount of confirmations.

## How test the changes
1. `POST /job/fortune`
2. `GET /job/cron/launch`
3. `PATCH /job/cancel/:jobId`
4. `GET /job/cron/cancel`
5. `GET /job/cron/refund`

`yarn test`

## Related issues
[[Job Launcher] Refund in cancellation#843](https://github.com/humanprotocol/human-protocol/issues/843)


packages/apps/job-launcher/server/src/modules/job/job.entity.ts
```diff
@@ -28,6 +28,9 @@ export class JobEntity extends BaseEntity implements IJob {
   @Column({ type: 'varchar' })
   public manifestHash: string;
 
+  @Column({ type: 'varchar', nullable: true })
+  public latestTransactionHash: string;
```
flopez7, line 32: I think checking the cancelled status of the subgraph would be a good idea instead of saving the hash of the cancelled transaction in the database and checking it later. We can avoid having a nullable field in the database that is going to be mostly null.

cc: @portuu3, @leric7

packages/sdk/typescript/human-protocol-sdk/src/escrow.ts
flopez7, line 1: You should add the same in Python SDK

flopez7:
I think checking the cancelled status of the subgraph would be a good idea instead of saving the hash of the cancelled transaction in the database and checking it later. We can avoid having a nullable field in the database that is going to be mostly null.

cc: @portuu3, @leric7

packages/apps/job-launcher/server/src/database/migrations/1695365418366-AddLatestTransactionHashToJob.ts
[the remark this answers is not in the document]
eugenvoronov: To avoid conflicts, because It had deployed to vercel. Also Adrian have recommended this in latest PR.

eugenvoronov:
> I think checking the cancelled status of the subgraph would be a good idea instead of saving the hash of the cancelled transaction in the database and checking it later. We can avoid having a nullable field in the database that is going to be mostly null.
> 
> cc: @portuu3, @leric7

Are we actually storing the hash data in a subgraph?

packages/apps/job-launcher/server/src/common/constants/index.ts
```diff
@@ -24,4 +24,5 @@ export const SENDGRID_API_KEY_REGEX =
 
 export const HEADER_SIGNATURE_KEY = 'human-signature';
 
-export const CVAT_JOB_TYPES = [JobRequestType.IMAGE_BOXES, JobRequestType.IMAGE_POINTS]
\ No newline at end of file
+export const CVAT_JOB_TYPES = [JobRequestType.IMAGE_BOXES, JobRequestType.IMAGE_POINTS]
```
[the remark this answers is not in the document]
eugenvoronov: No longer, this type will not be used for now.

packages/apps/job-launcher/server/src/database/migrations/1695365418366-AddLatestTransactionHashToJob.ts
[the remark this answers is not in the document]
eugenvoronov: My opinion, we can use one file as long as we can do this, since I don’t see the point in rolling out long migrations to get rid of one temporary status. On the other hand, changes go into the main branch, and it is better to separate migrations, although sometimes this is difficult.

packages/sdk/typescript/human-protocol-sdk/src/escrow.ts
eugenvoronov: I wanted to make this a separate PR, but why not do it here

flopez7:
> > I think checking the cancelled status of the subgraph would be a good idea instead of saving the hash of the cancelled transaction in the database and checking it later. We can avoid having a nullable field in the database that is going to be mostly null.
> > cc: @portuu3, @leric7
> 
> Are we actually storing the hash data in a subgraph?

Not the hash itself, but we are detecting the Canceled event and updating the status.

eugenvoronov:
> > > I think checking the cancelled status of the subgraph would be a good idea instead of saving the hash of the cancelled transaction in the database and checking it later. We can avoid having a nullable field in the database that is going to be mostly null.
> > > cc: @portuu3, @leric7
> > 
> > 
> > Are we actually storing the hash data in a subgraph?
> 
> Not the hash itself, but we are detecting the Canceled event and updating the status.

Is it guarantee that transaction will be confirmed by blockchain? It's guarantee that an event was emitted

flopez7:
> > > > I think checking the cancelled status of the subgraph would be a good idea instead of saving the hash of the cancelled transaction in the database and checking it later. We can avoid having a nullable field in the database that is going to be mostly null.
> > > > cc: @portuu3, @leric7
> > > 
> > > 
> > > Are we actually storing the hash data in a subgraph?
> > 
> > 
> > Not the hash itself, but we are detecting the Canceled event and updating the status.
> 
> Is it guarantee that transaction will be confirmed by blockchain? It's guarantee that an event was emitted

As the job laucher is the one who will send the transaction, the transaction will not be modified or deleted, so we can be sure that when the event is emited (this means that the transaction has been processed) the escrow has been cancelled.  Even in the SDK we are executing the transaction and then we run `await tx.wait()` inside the cancel method, so we can be sure that the transaction has been executed with at least 1 confirmation

packages/apps/job-launcher/server/src/database/migrations/1695365418366-AddLatestTransactionHashToJob.ts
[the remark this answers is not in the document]
eugenvoronov: @portuu3

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -804,4 +716,160 @@ export class JobService {
...
+    
+    await this.paymentService.createRefundPayment({ refundAmount: refundAmount.toNumber(), userId: userId, jobId: id });
+  
+    jobEntity.status = JobStatus.CANCELED;
+    await jobEntity.save();
+
+    return true;
+  }
+
+  public async calculateRefundAmount(chainId: ChainId, tokenAddress: string, escrowAddress: string) {
```
eugenvoronov, line 816: @portuu3

eugenvoronov:
> Also there is a mistake in the amount refunded, we cannot refund the total amount funded initaily, because a partial payout might have happened since the escrow was launched, we need to refund the amount transfered back to the job launcher when the cancel function of the contract is called and we should use the transaction logs for this.

Please, check this logic here, https://github.com/humanprotocol/human-protocol/pull/928/files#r1345434162

packages/apps/job-launcher/server/src/modules/job/job.service.ts
eugenvoronov: Here we are working with logs got from EOA. I will try to check, can we get something from cancel method tx.

PR #959, opened 2023-09-29
[Exchange Oracle] Upload solutions file

### LINKED ISSUE #949: [Exchange Oracle] Upload solutions file
**Description**
Upload the workers' solutions to an S3 file and send the URL of the file in the Recording Oracle webhook. 

**Implementation details** 
When the first solution is received, the file will be created and then the following solutions will be added to the same file.
To upload the file we must use the minio package as our StorageClient will be deleted.

flopez7:
## Description

Upload the workers' solutions to S3 and send the URL of the file in the Recording Oracle webhook.

## How test the changes

`yarn test`

## Related issues
Close #949

## Operational checklist


packages/apps/fortune/exchange-oracle/server/src/modules/job/job.service.ts
```diff
@@ -131,23 +138,22 @@ export class JobService {
...
+      workerAddress,
+      signer.address,
+      solution,
+      this.s3Config.bucket,
+    );
 
     await this.httpService.post(recordingOracleWebhookUrl, {
       escrowAddress: escrowAddress,
       chainId: chainId,
       exchangeAddress: signer.address,
```
portuu3, line 154: I think you don't need exchangeAddress here, it is already in the file and when we add the security it should be in the signature

packages/apps/fortune/exchange-oracle/server/src/common/interfaces/job.ts
```diff
@@ -0,0 +1,5 @@
+export interface ISolution {
```
portuu3, line 1: Solution can like 
{
  workerAddress: string;
  solution: string;
}

And then have like:
export interface ISolutionsFile {
    exchangeAddress: string;
    solutions: ISolution[];
}

So the file content would the class ISolutionsFile

portuu3 (changes requested):

portuu3 (approved):

PR #961, opened 2023-09-30
[SDK] [Typescript] Updated cancel method 

eugenvoronov:
## Description
Returns the hash of the transaction as the result of the cancel method.
The transaction hash is required when we process a refund to the user. The idea is to wait for a transaction to receive a certain number of confirmations before we receive logs of the last transfer.

## Summary of changes
- Update `cancel` method to SDK
- Updated GitBook

## How test the changes
`yarn test`


packages/sdk/typescript/human-protocol-sdk/test/escrow.test.ts
```diff
@@ -1071,9 +1076,12 @@ describe('EscrowClient', () => {
     });
 
     test('should successfully cancel escrow', async () => {
+      mockTx.wait.mockResolvedValueOnce({ transactionHash: FAKE_HASH });
+
       const escrowAddress = ethers.constants.AddressZero;
 
       escrowClient.escrowFactoryContract.hasEscrow.mockReturnValue(true);
+      escrowClient.escrowContract.cancel.mockResolvedValueOnce(mockTx);
```
flopez7, line 1084: Check if the hash returned by the `cancel` method is the same than `FAKE_HASH`

flopez7 (dismissed):

packages/sdk/typescript/human-protocol-sdk/test/escrow.test.ts
eugenvoronov: Updated

portuu3 (approved):

Issue #919, opened 2023-09-21
[Job Launcher] [Refund payment] Track tx amount confirmation


eugenvoronov:
**Description**
Using the transaction hash when we process a refund to the user. The idea is to wait for a transaction to receive a certain number of confirmations before we receive logs of the last transfer.

**Motivation**
If we try to get a logs after cancel escrow method we won't be able to find any logs about release, and we won't be able to calculate right refund amount without it. For this reason we will use tx confirmation too ensure the logs are presented.

**Implementation details** 
- Update refund payment logic
- Update unit tests

eugenvoronov:
Waiting approve from Adrian

PR #1012, opened 2023-10-07
[Job Launcher] Cancel and refund

eugenvoronov:
## Description
Canceled escrow, received `txHash` and `refundedAmount` from SDK `cancel` method and create new refund payment.

## Summary of changes
- Updated refund payment logic
- Updated unit tests

## How test the changes
`POST /job/fortune`
`GET /job/cron/launch`
`PATCH /job/cancel/{id}`
`GET /job/cron/cancel`
`yarn test`

## Related issues
[[Job Launcher] Cancel escrow and refund payment#919](https://github.com/humanprotocol/human-protocol/issues/919)


packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -633,65 +634,62 @@ export class JobService {
...
-          },
-        );
-      } else {
-        await this.sendWebhook(
-          this.configService.get<string>(
-            ConfigNames.CVAT_EXCHANGE_ORACLE_WEBHOOK_URL,
-          )!,
-          {
-            escrowAddress,
+  public async notifyWebhook(jobEntity: JobEntity) {
```
portuu3, line 679: This notifyWebhook cannot contain the code to set the payment to FAILED. In any case I think we don't need to set the payment as FAILED since the payment was actuially successfull and we are creating the REFUND.
Also does not make sense to have a notify webhook specific for cancellation, lets just use the generic send webhook, which can be still more generic like this:
```
public async sendWebhook(
    webhookUrl: string,
    webhookData: any,
  )
```

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -260,4 +263,29 @@ export class PaymentService {
...
+        await this.paymentRepository.create({
+            userId: dto.userId,
+            jobId: dto.jobId,
+            source: PaymentSource.BALANCE,
+            type: PaymentType.REFUND,
+            amount: dto.refundAmount,
+            currency: TokenId.HMT,
+            rate: div(1, rate),
+            status: PaymentStatus.SUCCEEDED,
+        });
```
portuu3, line 280: If change the order you don't need to do div(1,rate) later
```suggestion
    const rate = await getRate(TokenId.HMT, Currency.USD);

    try {
        await this.paymentRepository.create({
            userId: dto.userId,
            jobId: dto.jobId,
            source: PaymentSource.BALANCE,
            type: PaymentType.REFUND,
            amount: dto.refundAmount,
            currency: TokenId.HMT,
            rate: rate,
            status: PaymentStatus.SUCCEEDED,
        });
```

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/job/job.service.ts
eugenvoronov: Haven't checked the content of this method, it should work differently. I'll fix it.

portuu3 (approved):

PR #1071, opened 2023-10-17
[SDK] [Python] use class instead of dict for get escrow method

### LINKED ISSUE #1070: Fix Python SDK to return class instead of dict

leric7:
## Description

We need to return `EscrowData` class instead of normal dict.

## Related issues

Closes #1070 

## Operational checklist


portuu3 (dismissed):

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/escrow.py
```diff
@@ -900,11 +900,39 @@ def get_escrows(
...
-            for escrow in escrows:
-                escrow["chain_id"] = chain_id
-            escrow_addresses.extend(escrows)
-        return escrow_addresses
+            escrows_raw = escrows_data["data"]["escrows"]
+
+            escrows.extend(
+                [
+                    EscrowData(
+                        id=escrow["id"],
```
Dzeranov, line 908: Do we have a guarantee that all the fields we are trying to access will be presented in a `escrow` dict?
Anyways, I think it's not safe to access fields in this way because it can cause a `KeyError`, better to use `.get()`.
https://docs.python.org/3/library/stdtypes.html#dict.get

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/escrow.py
```diff
@@ -940,4 +968,29 @@ def get_escrow(
             },
         )
 
-        return escrow["data"]["escrow"]
+        escrow_raw = escrow["data"]["escrow"]
+
+        return EscrowData(
+            id=escrow_raw["id"],
```
Dzeranov, line 974: Same here.

Dzeranov (changes requested):

Dzeranov (approved):

PR #1140, opened 2023-10-26
[Job Launcher] job cancel button on failed page, add snackbar provider

m00n620:
## Description

- Add `Cancel` button on job detail page for failed jobs.
- Add `SnackbarProvider` to open alert snacks.


portuu3 (changes requested):
Once cancelled the button should disapear, it does only after refreshing.
Also statuses PENDING and LAUNCHED can be cancelled

portuu3 (approved):

PR #1144, opened 2023-10-26
[CVAT][Oracles] Fix CVAT oracles tests

flopez7:
## Description

Dockerfile setup fix for testing CVAT Exchange and Recording oracles
Fix and rewrite testcases in both exchange and recording oracles.

## Summary of changes

Added ffmpeg, libsm6 and libxext6 libs to Dockerfile setup to fix deps issue on a startup
Forced docker image node version to 18
Rewrite all tests cases to match the new code

## Related issues
#1152 

## Operational checklist
- [x]  All new functionality is covered by tests


packages/examples/cvat/exchange-oracle/docker-compose.test.yml
```diff
@@ -14,20 +14,20 @@ services:
...
-      # Using a magic nubmer of 28 here because this is a block number when blockchain-node container is ready to use
-      test: if [ $(( $(wget -q --post-data='{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}' -O- http://blockchain-node:8545 | grep -o '"result":"[^"]*"' | awk -F'"' '{print $4}' ) )) -ge 28 ]; then exit 0; else exit 1; fi
-      interval: 5s
-      timeout: 5s
-      retries: 15
-    networks:
-      - test-network
-    ports:
-      - 8545:8545
+  # blockchain-node:
```
Dzeranov, line 17: I think we could just remove it.

packages/examples/cvat/exchange-oracle/docker-compose.test.yml
```diff
@@ -80,16 +80,17 @@ services:
...
+      STORAGE_SECRET_KEY: 'devdevdev'
+      STORAGE_RESULTS_BUCKET_NAME: 'results'
+      STORAGE_USE_SSL: False
+      ENABLE_CUSTOM_CLOUD_HOST: Yes
     depends_on:
       postgres:
         condition: service_started
-      blockchain-node:
-        condition: service_healthy
+      # blockchain-node:
```
Dzeranov, line 92: Same here.

packages/examples/cvat/exchange-oracle/tests/api/test_cvat_webhook_api.py
```diff
@@ -0,0 +1,168 @@
...
+        "/cvat-webhook",
+        headers={"X-Signature-256": signature},
+        json=data,
+    )
+
+    assert response.status_code == 200
+
+    # Create some entities in test DB
+    add_cvat_project_to_db(cvat_id=1)
+    add_cvat_task_to_db(cvat_id=1, cvat_project_id=1, status="annotation")
```
zhiltsov-max, line 34: Let's use pytest to run tests, and its fixtures to simplify setup code instead.

packages/examples/cvat/exchange-oracle/tests/api/test_cvat_webhook_api.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Payload for "create:job" event
+    data = {
+        "event": "create:job",
+        "job": {
+            "url": "http://localhost:8080/api/jobs/1",
```
zhiltsov-max, line 40: It will be better to extract this address into a constant.

packages/examples/cvat/exchange-oracle/tests/api/test_cvat_webhook_api.py
```diff
@@ -0,0 +1,168 @@
...
+from tests.utils.setup_cvat import (
+    add_asignment_to_db,
+    add_cvat_project_to_db,
+    add_cvat_task_to_db,
+    generate_cvat_signature,
+    get_cvat_job_from_db,
+)
+
+
+def test_incoming_webhook_200(client: TestClient) -> None:
```
zhiltsov-max, line 17: It will be better to split this long test into several smaller ones.

packages/examples/cvat/exchange-oracle/tests/api/test_exchange_api.py
```diff
@@ -0,0 +1,283 @@
...
+                "/tasks",
+                headers={"signature": "sample"},
+            )
+
+            assert response.status_code == 200
+            assert isinstance(response.json(), list)
+            assert len(response.json()) == 2
+            for task in response.json():
+                assert not task["assignment"]
+                assert set(task.keys()) == {
```
zhiltsov-max, line 108: In CVAT we use the DeepDiff library to compare such data in tests. Consider using it here too.

packages/examples/cvat/exchange-oracle/tests/api/test_exchange_api.py
```diff
@@ -0,0 +1,283 @@
...
+        assert user["wallet_address"] == user_address
+        assert user["cvat_email"] == cvat_email_2
+
+
+def test_register_400(client: TestClient) -> None:
+    with SessionLocal.begin() as session:
+        # User with different wallet address
+        user = User(
+            wallet_address="0x86e83d346041E8806e352681f3F14549C0d2BC61",
+            cvat_email="test@hmt.ai",
```
zhiltsov-max, line 197: It's not clear from this code that these values are different from what's is in the `wallet_address` and `cvat_email` variables. Maybe it would be better to emphasize it somehow.

packages/examples/cvat/exchange-oracle/tests/api/test_exchange_api.py
```diff
@@ -0,0 +1,283 @@
...
+        wallet_address=user_address,
+        cvat_email="test@hmt.ai",
+        cvat_id=1,
+    )
+    session.add(user)
+    session.commit()
+    with (
+        open("tests/utils/manifest.json") as data,
+        patch("src.services.exchange.get_escrow_manifest") as mock_get_manifest,
+        patch("src.services.exchange.cvat_api"),
```
zhiltsov-max, line 246: Probably, you should also check that CVAT is called here. A job assignee should be changed after this call.

packages/examples/cvat/exchange-oracle/tests/api/test_exchange_api.py
flopez7: In that data we were not comparing the data itself, we just wanted to make sure that the data returned were instances of the TaskReturned class. I tried that with the isInstance method but didn't get it. So I'm just checking that the object contains some specific keys.

zhiltsov-max (approved):

PR #1156, opened 2023-10-29
feature: better profanity filter

AnshMasand:
## Description

Implement a profanity detecting transformer for the Human Protocol recording oracle. 

## Summary of changes

We have changed the curses file to download the transformer locally the first time the endpoint is accessed.
It returns a boolean value if the model is confident that the sentence contains hate speech/profanity/offensive language. 

## How test the changes

N/A

## Related issues
We are running into issues while developing, more info to follow. 

## Operational checklist


packages/examples/fortune/recording-oracle/src/plugins/config.ts
simulacrum6, line 1: Why was this file changed?

packages/examples/fortune/recording-oracle/src/plugins/curses.ts
```diff
@@ -1,465 +1,14 @@
...
-    for (let i = 0; i < curses.length; i++) {
-      const expression = new RegExp(curses[i], 'gi');
-      if (expression.test(text)) {
-        return true;
-      }
-    }
+  async isProfane(text: string) {
+    const result = await (await pipe)(text);
+    const profanity = result[0].find((res: { label: string; }) => res.label === "Profanity_detected");
+    return (profanity && profanity.score > 0.5);
```
simulacrum6, line 11: The score check is not necessary, pipe returns only the most likely label anyways, so score will always be > 0.5, or am I overlooking something?

simulacrum6:
@eugenvoronov At some point, we might want to look into the performance overhead of using transformer models. It might not matter much for the Fortune example, but processing one item at a time is fairly inefficient with neural networks. Usually, it is better to process them in batches.

packages/examples/fortune/recording-oracle/src/plugins/config.ts
simulacrum6, line 1: Did you add your access token to the [HF_ACCESS_TOKEN](https://huggingface.co/docs/transformers.js/guides/private) environment variable, or did you make the model public?

packages/examples/fortune/recording-oracle/src/plugins/curses.ts
```diff
@@ -1,465 +1,14 @@
 import fp from 'fastify-plugin';
 import { FastifyPluginAsync } from 'fastify';
+import { pipeline } from '@xenova/transformers';
 
-const curses = [
```
simulacrum6, line 4: I think we agreed on having a two-stage approach, where both a wordlist and the model prediction is checked. You mentioned an extended / improved list of cursewords.

simulacrum6: Alright, misunderstanding then.

packages/examples/fortune/recording-oracle/src/plugins/config.ts
simulacrum6: @eugenvoronov

packages/examples/fortune/recording-oracle/src/plugins/curses.ts
simulacrum6: It's alright. In case we want to add different models, I suppose we'd define a common interface for them anyways.

eugenvoronov (approved):

Issue #123, opened 2022-10-18
Add Celo and become the official maintainer for this network integration feature


ansaqib:

leetdev:
Hi, I'll be updating this checklist here to keep track of my progress.

* [x]  Suggest and have us authorize your bridging strategy. After all we need to bridge HMT over to the new network. Usually, we need to use an erc20-compatible approach.
* [x]  Deploy the contracts on testnet
* [x]  Verify the contracts on the testnet block scanner:
  - [HMToken](https://alfajores.celoscan.io/address/0x2736b33455a872dc478e1e004106d04c35472468#code)
  - EscrowFactory [Implementation](https://alfajores.celoscan.io/address/0xc90b43a5d576d9d8026c48904dfbaed50c15fa08#code) [Proxy](https://alfajores.celoscan.io/address/0x86af9f6cd34b69db1b202223c6d6d109f2491569#code)
  - Staking [Implementation](https://alfajores.celoscan.io/address/0x34cD3Bd6B16c559f321799b516dE61E12017fFd1#code) [Proxy](https://alfajores.celoscan.io/address/0x003548Df34be8836cF0F9673403a1E40ba449a0F#code)
  - RewardPool [Implementation](https://alfajores.celoscan.io/address/0xb9344bAD98E3d26a4d83900922baf395a2Ec154c#code) [Proxy](https://alfajores.celoscan.io/address/0xA9545C2530BD5bdb464d5E274F59ACceAa73eD86#code)
  - [EthKVStore](https://alfajores.celoscan.io/address/0x938335006ea6F9Eb0e8020969cFF94404425e298#code)
* [x]  Make sure that everything is working on testnet
* [ ]  Add the new deployed contracts addresses to this [list](https://github.com/humanprotocol/human-protocol/blob/main/CONTRACTS_LIST.md)
* [ ]  Transfer ownership of testnet contracts
* [ ]  Launch a working app on the network. The easiest one to get going is our example [fortune](https://github.com/humanprotocol/human-protocol/tree/main/packages/examples/fortune). Make a video of the functionality of the app working on the new network and link to relevant scanner transactions.
* [ ]  Deploy the contracts on mainnet
* [ ]  Verify the contracts on the mainnet block scanner
* [ ]  Make sure that everything is working on mainnet
* [ ]  Add the new deployed contracts addresses to this [list](https://github.com/humanprotocol/human-protocol/blob/main/CONTRACTS_LIST.md)
[...]

PR #1175, opened 2023-10-31
Add Celo integration

leetdev:
## Description

Added Celo mainnet and Celo Alfajores testnet support to the relevant packages, deployed contracts, and added to the list.

## Summary of changes

See #123 for details. **Create new subgraphs: `celo`, and `celo-alfajores` before releasing.**

## Related issues

#123

## Operational checklist


ansaqib:
@foufrix - Request review

foufrix:
The code seems good to me.
Only thing is contracts are not verified: https://celoscan.io/address/0x19Ead835951493A763c96910255d5eeF147E914F#code

Possible to verify it using [hardhat tutorial and tools](https://hardhat.org/hardhat-runner/docs/guides/verifying)

foufrix (approved):

foufrix:
solve #123

leetdev:
> The code seems good to me. Only thing is contracts are not verified: https://celoscan.io/address/0x19Ead835951493A763c96910255d5eeF147E914F#code
That one is the bridged token created by Wormhole. All the other contracts are verified using hardhat, see links in #123.

foufrix:
Effectively my bad there is 2 explorers, all good on celoscan.io

leetdev:
Rebased to the latest state of the `develop` branch.

I will reiterate that the following subraphs need to be created:
- https://api.thegraph.com/subgraphs/name/humanprotocol/celo
- https://api.thegraph.com/subgraphs/name/humanprotocol/celo-alfajores

I also noticed that Escrow Dashboard now uses an API instead of querying the subgraphs directly, not sure if I have to do anything to make sure it supports the new networks.

@portuu3 - please review.

package.json
```diff
@@ -47,6 +47,7 @@
     "@typescript-eslint/parser": "^5.43.0",
     "@typescript-eslint/utils": "^5.42.1",
     "concurrently": "^7.5.0",
+    "cross-env": "^7.0.3",
```
portuu3, line 50: remove this

packages/sdk/typescript/subgraph/template.yaml
portuu3, line 1: I don't understand this change, can you explain?

portuu3 (changes requested):

packages/sdk/typescript/subgraph/template.yaml
leetdev: Legacy Escrow Factory contract doesn't exist on this newly deployed network, and leaving it out would throw an error on subgraph config generation unless I wrapped it in a conditional block. Is there another way this should be handled?

portuu3: if you check Avalanche for example, legacy contract don't exist either and it doesn't throw any error, so it should be ok also in your case

leetdev: If you're talking about the regular avalanche config, it does throw errors for me:

Screenshot 2023-11-08 144119

This is because of the missing addresses and abi files (the fields are empty strings in the generated `subgraph.yaml`). 

The Avalance config template used in the Github subgraph workflow works because it uses the `avalance-v2` config which includes all the contracts, including `LegacyEscrowFactory`. The same is true for all subgraphs currently being automatically deployed - they all use v2 configs. For the Celo ones to work, they would also need to have all of these configured, or the `graph codegen` step would throw an error. Since the Legacy Escrow contract doesn't exist on the Celo networks, I turned that block into a conditional one in order for the codegen to run without errors.

portuu3: you need to use avalanche-v2
`NETWORK=avalanche-v2 yarn generat`

leetdev: Yes, but `avalance-v2.json` has `LegacyEscrowFactory` defined. That's what I'm saying. Without it defined, codegen throws the above errors.

leetdev: Since this isn't getting anywhere, I went ahead and deployed the useless legacy contract so that we can move on. But this is something to consider for future integrations.

portuu3:
@leetdev since you were right about subgraph template, can you undo your last commit please? sorry for my mistake

portuu3:
I will create the subgraphs

portuu3:
subgraphs created

leetdev:
> subgraphs created

Ok. I think you can merge the PR now unless there is anything else you want me to do?

portuu3 (approved):

Issue #1120, opened 2023-10-24
Create HuFi Client


Faouzijedidi1:
**Description**

Create the new client for HuFi 

[Figma File ](https://www.figma.com/file/u81GKebVavmQW8zGHb5Jgo/Dashboards---Everything-Apps?type=design&node-id=5252-24018&mode=design&t=VhoBCrZWN6mFDyEH-0)

PR #1247, opened 2023-11-19
[HuFi] Add HuFi Client 

Faouzijedidi1:
## Description

Adding HuFi Client to the Monorepo

## Summary of changes

New HuFi Client that launches campaigns and tracks them. 

Tested manually

Tests not implemented

## Related issues

https://github.com/humanprotocol/human-protocol/issues/1120


portuu3 (changes requested):
We probably should move it to packages/apps/hufi/client since it is not only a job a launcher. Also change the description in package.json

portuu3:
Hufi is going to have its own repo

PR #1347, opened 2023-12-12
[Reputation Oracle ] Decrypt data only when is encrypted

flopez7:
## Description

When the data was not encrypted, an error occurred because an attempt was made to parse the data as a JSON but the data was already an object.

## Summary of changes

Create a new method in SDK to check if data is encrypted using PGP
Decrypt only when the downloaded data is a PGP encrypted string. Otherwise, return the data itself.

## How test the changes

`yarn test`


packages/sdk/typescript/human-protocol-sdk/src/encryption.ts
```diff
@@ -437,4 +437,44 @@ export class EncryptionUtils {
...
+   * const isEncrypted = await EncriptionUtils.isEncrypted(message);
+   *
+   * if (isEncrypted) {
+   *   console.log('The message is encrypted with OpenPGP.');
+   * } else {
+   *   console.log('The message is not encrypted with OpenPGP.');
+   * }
+   * ```
+   */
+  public static isEncrypted(message: string) {
```
portuu3, line 471:
```suggestion
  public static isEncrypted(message: string) : boolean {
```

portuu3 (approved):

Issue #1354, opened 2023-12-14
SDK - Gas improvements


portuu3:
**Description**
Add the option to pass gas limit and gas price when sending transactions.
Add estimate gas in python SDK to avoid sending transactions that will fail because of contract requirements

**Motivation**
Improve interaction with SDK allowing to choose gas settings and avoid spending tokens in transactions with errors

PR #1355, opened 2023-12-14
[SDK] [Typescript] [Python] Gas improvements

flopez7:
## Description

Add the option to pass gas limit and gas price when sending transactions.
Add estimate gas in python SDK to avoid sending transactions that will fail because of contract requirements

Also add an enum for roles in Typescript SDK

## Summary of changes

Remove old gas_limit and gasMultiplier logic
Add a new parameter to configure the transaction params for all the blockchain write methods 
Add roles enum in Typescript SDK

## How test the changes

- Python: `make run-test`
- Typescript: `yarn test`

## Related issues
#1354


packages/sdk/python/human-protocol-sdk/human_protocol_sdk/utils.py
```diff
@@ -219,8 +220,19 @@ def handle_transaction(
             "You must add construct_sign_and_send_raw_middleware middleware to Web3 instance"
         )
     try:
-        tx_hash = tx.transact({"gas": gas_limit or GAS_LIMIT})
+        if tx_options and tx_options.get("gas") is None:
+            tx_options["gas"] = tx.estimate_gas()
+        elif tx_options is None:
+            tx_options = {"gas": tx.estimate_gas()}
+        print(tx_options.get("gas"))
```
alidzm, line 227: redundant

alidzm (dismissed):

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/utils.py
flopez7: @alidzm Actually it is not redundant, it is possible that the user does not set any parameter as `tx_options` and in that case it would come as null and would enter in that condition

portuu3 (approved):

PR #1366, opened 2023-12-15
[Dashhboard] Fix admin build issue

m00n620:
## Description

Fixed the build issues on strapi cloud.

## Summary of changes

- Updated the package dependencies
- Generated strapi typings
- Prettify some files.


portuu3 (changes requested):
We should use simple quotes as we do in all other projects

portuu3 (approved):

Issue #1365, opened 2023-12-15
Job Launcher - Split launch jobs cron job


portuu3:
**Description**
Split the current cron job into 3 different cron jobs:
1. Create escrows
2. Setup escrows
3. Fund escrows

**Motivation**
Better error handling and clarity in the code

**Implementation details** 
- Create 3 new cron jobs
Add or modify existing statuses if necessary

PR #1383, opened 2023-12-20
[Job Launcher] Split cron jobs to launch and lock while running.

### LINKED ISSUE #1282: Job Launcher - Cron job improvements
**Description**
At this moment Job Launcher, launchs only 1 job per execution of the cron job, but it should launch all pending jobs.
This might take to a situation where the next execution of the cron job starts before ending the previous. For this reason we need to handle the status of the cron job using the database.

**Motivation**
Avoid long queues of jobs

**Implementation details** 
Create a table called cronJobs that contains the status and date of start and end of the execution.
When a cron jobs starts check the status, if it is running then skip the execution.
If a job fails, continue with the next in the list.
After X retries (env variable) if we still get error launching a job, set it failed.

leric7:
## Description

We need to split the cron job to launch the escrow, to make every step more obvious - creation, setting up, and funding.
In addition, we'll handle all of those job entities with relevant status in each cron job run, and we want to lock cron job to prevent cron jobs running before previous one is finished.

## Summary of changes
- Split cron job to three separate ones.
- Add cron job entity to database to lock each cron job run.
- Updated test cases.

## Related issues

Closes #1282, #1365.


packages/apps/job-launcher/server/src/common/interfaces/cron-job.ts
```diff
@@ -0,0 +1,8 @@
+import { CronJobType } from '../enums/cron-job';
+import { IBase } from './base';
+
+export interface ICronJob extends IBase {
+  cronJobType: CronJobType;
+  createdAt: Date;
```
portuu3, line 6:
```suggestion
  cronJobType: CronJobType;
```

portuu3, line 6:
```suggestion
```
IBase already has createdAt property

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -901,18 +913,80 @@ export class JobService {
...
 
   @Cron(CronExpression.EVERY_10_MINUTES)
-  public async launchCronJob() {
-    this.logger.log('Launch jobs START');
+  public async createEscrowCronJob() {
+    const isCronJobRunning = await this.cronJobService.isCronJobRunning(
+      CronJobType.CreateEscrow,
+    );
+
+    if (isCronJobRunning) {
```
portuu3, line 921: This might be moved into isCronJobRunning

leric7: Sorry, I'm not following.

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -901,18 +913,80 @@ export class JobService {
...
+      this.logger.log('Previous cron job is not completed yet');
+      return;
+    }
+
+    this.logger.log('Create escrow START');
+    const cronJob = await this.cronJobService.createCronJob(
```
portuu3, line 927: does it work if the cron job entity already exists?

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -901,18 +913,80 @@ export class JobService {
...
+      CronJobType.CreateEscrow,
+    );
+
     try {
-      // TODO: Add retry policy and process failure requests https://github.com/humanprotocol/human-protocol/issues/334
-      let jobEntity;
+      const jobEntities = await this.jobRepository.find(
+        {
+          status: In([JobStatus.PAID, JobStatus.CREATING]),
```
portuu3, line 934: I think the status filter here should be only PAID

packages/apps/job-launcher/server/src/modules/job/job.service.ts
leric7: Yes, that works. Unique index on cron job entity is `cron job type` + `created at`

leric7: I confirmed multiple cron job entities are created under `create-escrow` type.

portuu3 (commented):
Also webhook cron job in webhook.service.ts should be included

packages/apps/job-launcher/server/src/modules/job/job.service.ts
portuu3: I am talking about this line this.logger.log('Previous cron job is not completed yet');

packages/apps/job-launcher/server/src/modules/job/job.service.ts
portuu3: Why to create multiple? The way I see it, it should have only one entry in the database and we update its status so it works as a flag. For example, if there is no entity or last execution was completed means we can execute it, so when we start update the field updatedat with current date and set completedat to null, this means it is being run. I dont say your implementation is wrong but I think we would be filling a table for each execution, with information that is not actually useful

portuu3:
I think these should be right statuses:
```
  PENDING = 'PENDING',
  PAID = 'PAID',
  CREATED = 'CREATED’,
  SET_UP = ‘SET_UP’,
  LAUNCHED = 'LAUNCHED',
  COMPLETED = 'COMPLETED',
  FAILED = 'FAILED',
  TO_CANCEL = 'TO_CANCEL',
  CANCELED = 'CANCELED',
```
So for each cron job:
- Create: get PAID and set them to CREATED when the escrow has been created.
- Set up: get CREATED and set them to SET_UP when the escrow has been set up.
- Fund: get SET_UP and set them to LAUNCHED when the escrow has been funded and is ready to be processed.

packages/apps/job-launcher/server/src/common/enums/cron-job.ts
```diff
@@ -0,0 +1,7 @@
+export enum CronJobType {
+  CreateEscrow = 'create-escrow',
+  SetupEscrow = 'setup-escrow',
+  FundEscrow = 'fund-escrow',
+  CancelEscrow = 'cancel-escrow',
+  ProcessPending = 'process-pending',
```
portuu3, line 6:
```suggestion
  ProcessPendingWebhook = 'process-pending-webhook',
```

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -0,0 +1,62 @@
...
+
+      cronJob.startedAt = new Date();
+      cronJob.completedAt = null;
+      await cronJob.save();
+    }
+
+    return cronJob;
+  }
+
+  public async isCronJobRunning(cronJobType: CronJobType): Promise<boolean> {
```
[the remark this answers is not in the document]
leric7: removed check from startCronJob function.

portuu3 (approved):

Issue #1486, opened 2024-01-19
Job Launcher - Read JL fee from KVStore


portuu3:
**Description**
Read Job Launcher fee from KVStore instead of using an env variable

**Motivation**
Use public values from KVStore

PR #1513, opened 2024-01-23
[Job Launcher] Get Job Launcher fee from KVStore

flopez7:
## Description

Read Job Launcher fee from KVStore instead of using an env variable

## Summary of changes

Remove the previous implementation using an env variable and update it by getting from KVStore

## How test the changes

`yarn test`

## Related issues
#1486


packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -1460,14 +1458,17 @@ export class JobService {
   }
 
   private async getOracleFee(
-    oracleAddress: string,
     chainId: ChainId,
+    oracleAddress?: string,
```
portuu3, line 1462: Dont make this param optional, it is not intuitive at all how to use this method

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -1467,7 +1470,10 @@ export class JobService {
 
     const kvStoreClient = await KVStoreClient.build(signer);
 
-    const feeValue = await kvStoreClient.get(oracleAddress, KVStoreKeys.fee);
+    const feeValue = await kvStoreClient.get(
+      oracleAddress || signer.address,
```
portuu3, line 1474: remove signer.address and also `const signer = this.web3Service.getSigner(chainId);`

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/job/job.service.ts
flopez7: The line const `signer = this.web3Service.getSigner(chainId)` is being used to create the KVStore client

portuu3 (approved):

Issue #1532, opened 2024-01-25
Rework `service` - `repository` schema


Dzeranov:
**Description**
Need to rework `service` - `repository` schema.

**Motivation**
To follow nest documentation and clean up the code so it would be easier to modify and test it in the future.

**Implementation details** 
1. Need to use an in-build `Repository` module from `typeorm` as it suggested in the nest documentation: https://docs.nestjs.com/recipes/sql-typeorm 
However, for better error handling (for example, handling a duplicate error) base typeorm `Repository` might be extended. Example: https://orkhan.gitbook.io/typeorm/docs/custom-repository
Why: because repos errors should be handled on a service level and by default typeorm exceptions (such as duplicate error)
will be bubbled up to controller level.
2. On a service level we should operate with entities. For example (`/user.service.ts`)
Current:
```
public async create(dto: UserCreateDto): Promise<UserEntity> {
    const { email, password, ...rest } = dto;

    await this.checkEmail(email, 0); 

    return await this.userRepository.create({
      ...rest,
      email,
      password: bcrypt.hashSync(password, this.HASH_ROUNDS),
      type: UserType.REQUESTER,
      status: UserStatus.PENDING,
    });
  }
```
How it could be done:
```
1. Create a userEntity first. No need of checking whether user with the same email exists because we've already set up a email field as unique and in case of a duplication error, repository will simply throw an exception which should be properly handled.
2. Then, call this.usersRepository.insert(userEntity). No need of using .save() since it's just a plain insert and we don't need to update any relations.
3. Return a userEntity from step 1.
```

PR #1552, opened 2024-02-01
[Job Launcher] Rework Cron Job service - repository schema

flopez7:
## Description

Rework service - repository schema in Cron Job module

## Summary of changes

Remove all unnecessary methods from user repository and extend it from base Repository

## How test the changes

`yarn test`

## Related issues
#1532


Dzeranov (approved):
lgtm
@portuu3 please take a look.

portuu3 (dismissed):

packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -32,25 +32,21 @@ export class CronJobService {
   ) {}
 
   public async startCronJob(cronJobType: CronJobType): Promise<CronJobEntity> {
-    let cronJob = await this.cronJobRepository.findOne({
-      cronJobType,
-    });
+    let cronJob = await this.cronJobRepository.findOneByType(cronJobType);
```
eugenvoronov, line 35: Let's use immutable variables with `const`.

```
const cronJob = await this.cronJobRepository.findOneByType(cronJobType);

if (!cronJob) {
  return this.cronJobRepository.createUnique(cronJobType);
}

cronJob.startedAt = new Date();
cronJob.completedAt = null;

return this.cronJobRepository.updateOne(cronJob);
```

eugenvoronov: @portuu3

flopez7: We can't do it that way. With the example you have provided, you are not assigning any values to the startedAt and completedAt properties as the original logic did. Anyways, it could be done by adding to the entity properties the default values @eugenvoronov

eugenvoronov: But as I can see its totally the same logic as you did, no?
To use a default value its a good point @flopez7

packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -37,17 +37,17 @@ export class CronJobService {
...
   public async startCronJob(cronJobType: CronJobType): Promise<CronJobEntity> {
-    let cronJob = await this.cronJobRepository.findOneByType(cronJobType);
+    const cronJob = await this.cronJobRepository.findOneByType(cronJobType);
 
     if (!cronJob) {
-      cronJob = await this.cronJobRepository.createUnique(cronJobType);
+      const cronJobEntity = new CronJobEntity();
+      cronJobEntity.cronJobType = cronJobType;
+      return this.cronJobRepository.createUnique(cronJobEntity);
     } else {
```
eugenvoronov, line 46: Usually you don't need `else` statement, because `if (!cronJob)` false, you will skip `if` block.

eugenvoronov (approved):

PR #1558, opened 2024-02-01
[Job Launcher] Rework apikey, token and auth service - repository schemas

flopez7:
## Description

Rework service - repository schema in apikey, token and auth modules

## Summary of changes

Remove all unnecessary methods from repositories and extend it from base Repository

## How test the changes

`yarn test`

## Related issues
#1532


packages/apps/job-launcher/server/src/modules/auth/auth.service.ts
```diff
@@ -114,11 +118,11 @@ export class AuthService {
   }
 
   public async logout(user: UserEntity): Promise<void> {
-    await this.authRepository.delete({ userId: user.id });
+    await this.authRepository.deleteByUserId(user.id);
```
Dzeranov, line 121: We can remove this service method since it doesn't have any logic apart from deleting tokens from DB.

Dzeranov (changes requested):

packages/apps/job-launcher/server/src/modules/auth/auth.service.ts
portuu3: how would you do it then?

packages/apps/job-launcher/server/src/modules/auth/auth.service.ts
```diff
@@ -190,10 +196,10 @@ export class AuthService {
   }
 
   public async restorePassword(data: RestorePasswordDto): Promise<void> {
-    const tokenEntity = await this.tokenRepository.findOne({
-      uuid: data.token,
-      tokenType: TokenType.PASSWORD,
-    });
+    const tokenEntity = await this.tokenRepository.findOneByUuidAndTokenType(
```
portuu3, line 199: Should it use tokenService instead of tokenRepository?

packages/apps/job-launcher/server/src/modules/auth/auth.service.ts
eugenvoronov: I think what @Dzeranov means is to move `await this.authRepository.deleteByUserId(user.id);` directly to the methods that call the logout service method. @portuu3 @flopez7

packages/apps/job-launcher/server/src/modules/auth/auth.service.ts
eugenvoronov: I think we need to follow the logic: if a method that calls another service method can directly call a repository method, if that service method doesn't contain any business logic, just, for example, look up something in the DB and return it. @Dzeranov @portuu3

flopez7: You're right @eugenvoronov, that's the logic we're following to reduce the amount of code we have

packages/apps/job-launcher/server/src/modules/cron-job/cron-job.repository.ts
```diff
@@ -1,34 +1,44 @@
...
+export class CronJobRepository extends Repository<CronJobEntity> {
+  constructor(private dataSource: DataSource) {
+    super(CronJobEntity, dataSource.createEntityManager());
+  }
 
-  constructor(
-    @InjectRepository(CronJobEntity)
-    private readonly cronJobEntityRepository: Repository<CronJobEntity>,
-  ) {}
+  async createUnique(type: CronJobType): Promise<CronJobEntity> {
```
Dzeranov, line 13: Why don't we just use base repo here?

flopez7: For that repository, I made the change in its own PR https://github.com/humanprotocol/human-protocol/pull/1552

Dzeranov: Ok.

Dzeranov (approved):

Issue #1487, opened 2024-01-19
Job Launcher - Send sign in hcaptcha token for verification


portuu3:
**Description**
Add to sign in dto the hcaptcha token and verify it.

**Motivation**
Use hcaptcha properly

PR #1569, opened 2024-02-05
[Job Launcher] Send hcaptcha token for verification

flopez7:
## Description

Send hcaptcha token to server for verification

## Summary of changes

Add the hcaptcha token to server and client login dtos
Check the token in hcaptcha server
If the token is correct, send the proper tokens to the client

## How test the changes

`yarn test`

## Related issues
#1487


packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
portuu3, line 1: Uncomment

portuu3 (changes requested):

portuu3 (approved):

PR #1590, opened 2024-02-13
Human APP backend - 1st milestone

BartoszSolkaBD:
## Description

This PR adds the code from 1st milestone for Human App backend-for-frontend

## Summary of changes

- project configuration
- worker signup
- operator signup
- worker signin

## How test the changes

There are end to end tests that can be run by running `yarn run test:e2e`


packages/apps/human-app/server/src/integrations/reputation-oracle/reputation-oracle.gateway.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  private async handleRequestToReputationOracle<T>(
+    options: AxiosRequestConfig,
+  ): Promise<T> {
+    try {
+      const response = await lastValueFrom(this.httpService.request(options));
+      return response.data;
+    } catch (error) {
+      if (error.response) {
+        throw new HttpException(error.response.data, error.response.status);
```
Dzeranov, line 61: @BartoszSolkaBD any particular reasons to throw `HttpException` here? I mean, technically, it's a service level which shouldn't be responsible for that.
Instead, we could throw here an app error and use nest exception filters: https://docs.nestjs.com/exception-filters.
Please, let me know what do you think about that.

BartoszSolkaBD: For now, we don't see benefits from the suggested refactoring. The current solution is more readable in our opinion. Additionally, it's similar to how the error handling is implemented in the current implementation of the Reputation Oracle server

Dzeranov: Tbh I wouldn't recommend using a reputation oracle server implementation as a reference because it's going to be refactored quite heavily since it's just not following a lot of conventions.
Well, you could leave it as it but please keep in mind that during the 2nd milestone we would ask you to change it.

portuu3:
Is there a reason for doing 2 mappings of the DTO since the request is received until it is by passed to the Reputation Oracle? I think technically is good, but it is kind of redundant. Just wonder if it has been done for following best practices guidelines or for any other reason.

portuu3:
We need additional test cases apart from the positive ones currently in use. It's important to include tests for error handling as well.

packages/apps/human-app/server/src/integrations/reputation-oracle/reputation-oracle.gateway.ts
BartoszSolkaBD: okay, fine for me

BartoszSolkaBD:
@portuu3 
> We need additional test cases apart from the positive ones currently in use. It's important to include tests for error handling as well.

There are tests for errors in both unit and e2e layers. Can you be more specific about where you think it's not covered?

portuu3:
> @portuu3
> 
> > We need additional test cases apart from the positive ones currently in use. It's important to include tests for error handling as well.
> 
> There are tests for errors in both unit and e2e layers. Can you be more specific about where you think it's not covered?

You are right, I was looking at service level. Looks good to me

portuu3 (approved):

Issue #1684, opened 2024-03-11
Fortune Exchange Oracle - JWT authentication


portuu3:
**Description**
Verify JWTs used by users and generated by the Reputation Oracle.

**Motivation**
Auhtentication users for jobs discovery and assignment.
This will be used for E2E tests of Human App

**Implementation details** 
Get the public key for verification from KVStore using the key `jwt_public_key`
Verify that the user has passed KYC.

PR #1693, opened 2024-03-11
[Exchange oracle] JWT authentication 

flopez7:
## Description

Verify JWTs used by users and generated by the Reputation Oracle

## Summary of changes

Create a new auth guard and interceptor to verify the token
Obtain the private key from KVStore
Verify if the user has already passed KYC

## How test the changes

Upload the Reputation Oracle public key to S3 and set the url in KVStore
Generate a new valid token signed with Reputation Oracle private key and use it as bearer token

## Related issues
#1684


packages/apps/fortune/exchange-oracle/server/src/common/guards/strategy/jwt.http.ts
```diff
@@ -0,0 +1,56 @@
...
+          done(error);
+        }
+      },
+      passReqToCallback: true,
+    });
+  }
+
+  public async validate(
+    @Req() request: any,
+    payload: { email: string; address: string; kyc_status: string },
```
portuu3, line 45: You will also need to include reputation_network. Later when you get the jobs list, you will have to filter by reputation oracle

portuu3 (changes requested):

packages/apps/fortune/exchange-oracle/server/src/common/guards/strategy/jwt.http.ts
flopez7: Discussed on call

portuu3 (approved):

PR #1827, opened 2024-04-05
[Exchange Oracle] Add job endpoint authentication 

### LINKED ISSUE #1824: Fortune Exchange Oracle - Solve job endpoint authentication
**Description**
Modify authentication of solve job endpoint. 
It should use a web3 signature of the body as authorization header.
Use [this](https://github.com/humanprotocol/human-protocol/blob/8b3fd26618116d403421ccd27a3a751b5191cec4/packages/apps/fortune/exchange-oracle/server/src/common/guards/signature.auth.ts)

**Motivation**
Users will sign the messages to auth in the client

**Implementation details** 
Create a new guard

mrhouzlane:
## Description

- Change `SignatureAuthGuard` to include condition if role is worker. 

## Summary of changes

- Add worker condition and verify signature.

Closes #1824


packages/apps/fortune/exchange-oracle/server/src/modules/job/job.controller.ts
```diff
@@ -74,8 +82,10 @@ export class JobController {
     description: 'Unauthorized. Missing or invalid credentials.',
   })
   @Post('solve')
+  @UseGuards(SignatureAuthGuard)
```
flopez7, line 85: You have to specify here which role is to be allowed, e.g. in this case it is worker only.

```suggestion
  @UseGuards(
    new SignatureAuthGuard([Role.Worker]),
  )
```

flopez7 (changes requested):

packages/apps/fortune/exchange-oracle/server/src/common/guards/signature.auth.ts
```diff
@@ -20,6 +21,9 @@ export class SignatureAuthGuard implements CanActivate {
     const signature = request.headers[HEADER_SIGNATURE_KEY];
     const oracleAdresses: string[] = [];
     try {
+      if (this.role.includes(Role.Worker)) {
+        return verifySignature(data, signature, [data.senderAddress]);
```
flopez7, line 25: The DTO you're using in this endpoint doesn't contain the sender address. You should modify the DTO of the endpoint by requesting now the assignmentId. With that, you can check the workerAddress in the database for verifying and also you can get escrowAddress and chainId from there.

flopez7 (changes requested):

packages/apps/fortune/exchange-oracle/server/src/modules/job/job.service.ts
```diff
@@ -282,4 +282,30 @@ export class JobService {
...
+      throw new NotFoundException('Job not found for the assignment');
+    }
+
+    const { escrowAddress, chainId } = job;
+    return {
+      workerAddress,
+      escrowAddress,
+      chainId,
+    };
+  }
```
flopez7, line 310: Why have you created this method? It doesn't have any reference apart from the tests. 
Anyway, as you have the `assignmentId` in the `SolveJob` method, you should modify this one to, instead of searching the task by `escrowAddress` and `worker`, use the `assignmentId`

flopez7 (changes requested):

packages/apps/fortune/exchange-oracle/server/src/modules/job/job.service.ts
mrhouzlane: Fixed

portuu3 (approved):

Issue #1888, opened 2024-04-18
Fortune Exchange Oracle - RPC urls as environment variables


portuu3:
**Description**
Set the RPC urls as environment variables

**Motivation**
Improve security of our RPCs

**Implementation details** 
If the RPC for a specific network is not passed, this network should be disabled.

PR #1917, opened 2024-04-24
[Fortune Exchange Oracle] RPC urls as environment variables

flopez7:
## Description

Set the RPC urls as environment variables to improve security of our RPCs

## Summary of changes

Create a new `network-config.service.ts` to read .env variables 

## How test the changes

`yarn test`

## Related issues
#1888


leric7 (changes requested):
- I think it's better to manage RPC urls on Vercel, env files, etc if we use`RPC_URL` as prefix rather than suffix for env variables. I already merged couple PRs using this prefix approach in the codebase, so it's good to use same pattern across the entire codebase, as well.
- In addition, please update `.env.example` file by adding these new env variables.

leric7 (approved):

packages/apps/fortune/exchange-oracle/server/src/common/config/env-schema.ts
```diff
@@ -18,6 +18,14 @@ export const envValidator = Joi.object({
   // Web3
   WEB3_ENV: Joi.string(),
   WEB3_PRIVATE_KEY: Joi.string().required(),
+  RPC_URL_POLYGON: Joi.string(),
+  RPC_URL_BSC: Joi.string(),
+  RPC_URL_AMOY: Joi.string(),
+  RPC_URL_GOERLI: Joi.string(),
```
portuu3, line 24: Remove Goerli, it is deprecated

portuu3 (approved):

PR #1918, opened 2024-04-25
[Fortune Recording Oracle] RPC urls as environment variables

flopez7:
## Description

Set the RPC urls as environment variables to improve security of our RPCs

## Summary of changes

Create a new `network-config.service.ts` to read .env variables

## How test the changes

`yarn test`

## Related issues
#1889


leric7 (changes requested):
- I think it's better to manage RPC urls on Vercel, env files, etc if we use`RPC_URL` as prefix rather than suffix for env variables. I already merged couple PRs using this prefix approach in the codebase, so it's good to use same pattern across the entire codebase, as well.
- In addition, please update `.env.example` file by adding these new env variables.
- And please make sure you defined these env vars on Vercel.

leric7 (approved):

portuu3 (approved):

Issue #1940, opened 2024-04-30
[Job Launcher Server] Endpoint for escrow cancellation by escrow address


Dzeranov:
**Description**
Add an endpoint for escrow cancellation by escrow address and chain id.

**Motivation**
Currently, `/escrow/cancel/{id}` works only with the internal id of a job, which is not really handy in case someone is using API directly.

PR #1945, opened 2024-05-01
[Job Launcher] Escrow cancelation by chain id and escrow address

eugenvoronov:
## Description
Added an endpoint for escrow cancellation by escrow address and chain id. Previous implementation`/escrow/cancel/{id}` worked only with the internal id of a job, which is not really handy in case someone is using API directly.

## Summary of changes
- Added new endpoint `/job/cancel/:chain_id/:escrow_address`
- Updated unit tests

## How test the changes
PATCH `/job/cancel/:chain_id/:escrow_address`
`yarn test`

## Related issues
[Job Launcher Server] Endpoint for escrow cancellation by escrow address
[#1940](https://github.com/humanprotocol/human-protocol/issues/1940)


packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -952,10 +952,24 @@ export class JobService {
     return jobEntity;
   }
 
-  public async requestToCancelJob(userId: number, id: number): Promise<void> {
-    const jobEntity = await this.jobRepository.findOneByIdAndUserId(id, userId);
+  public async requestToCancelJob(
+    userId: number,
+    id: number,
```
portuu3, line 957: It is not a good practice to use id input para as chainid and jobid. I suggest to have two methods:
- requestToCancelJobById(userId: number, jobId: number)
- requestToCancelJobByAddress(userId: number, chainId: number, escrowAddress: string)

Then a third method being called from previous ones:
- requestToCancelJob(job: JobEntity)

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/job/job.service.ts
eugenvoronov: I will change it

portuu3 (approved):

Issue #1845, opened 2024-04-09
Job Launcher UI - Refresh token not used


portuu3:
**Describe the bug**
1. For some reason, client stores access token twice: as an access token and as a refresh token.
2. Access token is not being refreshed when it gets expired.

Client should handle access token expiration and issue a new one automatically using refresh token.

**To Reproduce**
1. Log in
2. Wait until the access token expires

1. Log in
2. Check local storage in browser - access and refresh tokens are the same.

**Expected behavior**
1. Refresh token (if it's valid) should be used for re-issuing access token after 401 error from server. In case refresh token is no longer valid, user must be logged out.

PR #1973, opened 2024-05-07
Fix: Job-Launcher Ui- Refresh token

MicDebBlocky:
## Description

The change fixes the saving of the refresh token in the localstorage and when beckend response is 401, it sends a request that retrieves and saves the new access token 

## Summary of changes

Added a new interceptor to axios that creates a query queue if we got a 401. Sends a query for a new token with the current refresh token. Refreshes the query queue with the new access token

## How test the changes
Make any request, wait 10 minutes for the access token to expire. Execute another request and check if everything works correctly, ie:
1. a request for a new access token has been executed
2. the requests that failed were executed again
3. we were not logged out of the application when we got a 401 error 

## Related issues
[#1845](https://github.com/humanprotocol/human-protocol/issues/1845)


MicDebBlocky:
#1845

flopez7 (approved):

Issue #1047, opened 2023-10-13
Reputation Oracle - Return high for its own reputation


portuu3:
**Description**
For now Reputation Oracle should return high when its reputation score is requested for Reputation oracle address.

**Motivation**
Since for now it is centralized we should return high

**Implementation details** 
Save a value above high threshold on database

PR #1977, opened 2024-05-08
[Reputation Oracle] Save a value above high threshold on database

eugenvoronov:
## Description
For now Reputation Oracle should return high when its reputation score is requested. Saves a value above high threshold on database

## Summary of changes
- Added a highLevel value to initial value in increaseReputation method for reputation entity.
- Added a highLevel value to initial value in decreaseReputation method for reputation entity.
- Updated unit tests.

## How test the changes
`yarn test`

## Related issues
Reputation Oracle - Return high for its own reputation
[#1047](https://github.com/humanprotocol/human-protocol/issues/1047)


portuu3 (changes requested):
This is not the functionality required, the idea is that only for Reputation Oracle address we return always high reputation in GET /reputation endpoint, but the rest the reputation is based on the real score

eugenvoronov:
```
    const reputationOracleAddress =
      await escrowClient.getReputationOracleAddress(escrowAddress);

    await this.increaseReputation(
      chainId,
      reputationOracleAddress,
      ReputationEntityType.REPUTATION_ORACLE,
    );
```

portuu3 (approved):

PR #2108, opened 2024-06-05
Fix for allowed statuses to cancel a job

portuu3:
## Description

Fix to allow cancel jobs in status CREATED, SET_UP, PARTIAL

## Summary of changes
- Change allowed statuses that can be cancelled
- Set some services/repositories instances private
- Tests

## How test the changes
yarn test

### DIFF
--- a/packages/apps/job-launcher/server/src/modules/webhook/webhook.module.ts
+++ b/packages/apps/job-launcher/server/src/modules/webhook/webhook.module.ts
@@ -1,4 +1,4 @@
-import { Logger, Module, forwardRef } from '@nestjs/common';
+import { Logger, Module } from '@nestjs/common';
 import { TypeOrmModule } from '@nestjs/typeorm';
 import { ConfigModule } from '@nestjs/config';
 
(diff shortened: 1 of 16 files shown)


packages/apps/job-launcher/server/src/modules/webhook/webhook.service.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
-import { HttpStatus, Inject, Injectable } from '@nestjs/common';
+import { HttpStatus, Inject, Injectable, forwardRef } from '@nestjs/common';
```
flopez7, line 2:
```suggestion
import { HttpStatus, Inject, Injectable } from '@nestjs/common';
```

packages/apps/job-launcher/server/src/modules/webhook/webhook.module.ts
```diff
@@ -1,4 +1,4 @@
-import { Logger, Module } from '@nestjs/common';
+import { Logger, Module, forwardRef } from '@nestjs/common';
```
flopez7, line 1:
```suggestion
import { Logger, Module } from '@nestjs/common';
```

flopez7 (changes requested):

flopez7 (approved):

Issue #2045, opened 2024-05-21
[Job Launcher] Add failed reason in job details


portuu3:
**Description**
Add failed reason in get job details response

**Motivation**
Give more information to the users

**Implementation details** 
The reason is stored in the database

PR #2145, opened 2024-06-14
Added failed reason

eugenvoronov:
## Description
Added failed reason in get job details response

## Summary of changes
- Added failed reason to all places where is it needed 
- Updated get job details endpoint
- Added new escrow errors 

## Related issues
[Job Launcher] Add failed reason in job details
[#2045](https://github.com/humanprotocol/human-protocol/issues/2045)

### DIFF
--- a/packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
+++ b/packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
@@ -1,11 +1,7 @@
 import { HttpStatus, Injectable, Logger } from '@nestjs/common';
-
+import { v4 as uuidv4 } from 'uuid';
 import { CronJobType } from '../../common/enums/cron-job';
-import {
-  ErrorCronJob,
-  ErrorEscrow,
-  ErrorJob,
-} from '../../common/constants/errors';
+import { ErrorCronJob, ErrorEscrow } from '../../common/constants/errors';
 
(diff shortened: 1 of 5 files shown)


packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -88,6 +92,7 @@ export class CronJobService {
         try {
           await this.jobService.createEscrow(jobEntity);
         } catch (err) {
+          jobEntity.failedReason = ErrorEscrow.NotCreated;
```
portuu3, line 95: Pass the failed reason into handleProcessJobFailure.
Also we should generate an errorId (GUID) and include it in the log and in the failed reason, so that later when we try to find out what happened we can identify the log

portuu3 (changes requested):

packages/apps/job-launcher/server/src/modules/cron-job/cron-job.service.ts
eugenvoronov: Error setting up escrow. Error ID: ab83a915-5856-49cd-8e0b-386f7998b4b5, Job ID: 44, Reason: Escrow has not been setup (Error ID: ab83a915-5856-49cd-8e0b-386f7998b4b5), Message: Manifest validation failed

portuu3 (approved):

Issue #2148, opened 2024-06-17
[Reputation Oracle] NDA


portuu3:
**Description**
Workers should sign an NDA to protect requesters datasets

**Motivation**
Protect sensitive data and intellectual property from requesters dataset, ensuring compliance with data privacy laws.

**Implementation details** 
1. Create NDAVersion Entity
Define an entity for NDA versions. This entity should include fields for the version identifier, document text, and creation date.
id: Primary key
version: String to represent the version of the NDA
documentText: Text of the NDA
createdAt: Timestamp of when the NDA version was created
updatedAt: Timestamp of when the NDA version was updated

2. Create NDA Entity
Define an entity for the NDA that users will sign. This entity should reference the NDAVersion entity.

id: Primary key
createdAt: Timestamp of when the NDA was signed
ipAddress: IP address of the user at the time of signing
userId: relationship with the User entity
ndaVersionId: relationship with the NDAVersion entity

3. Implement NDA Endpoints
Create these endpoints:
- `GET: ` `/nda`: for retrieving the last NDA version. If the user already signed it, return null. If it is not signed yet, return the nda version information. Use JWT auth
- `POST:` `/nda/sign`: for signing NDA. Use JWT auth

4. Add NDA information to JWT
Add wheter NDA has been signed to the JWT payload as:
`"NDA": "true" or "false"`
If the last NDA version is not the last signed by the user, set it to `false`

portuu3:
We will go different with a different approach: https://github.com/humanprotocol/human-protocol/issues/3105

PR #2160, opened 2024-06-19
Implemented NDA

eugenvoronov:
## Description
Implemented a feature requiring workers to sign a NDA to protect requesters' datasets. It ensures compliance with data privacy laws and safeguards sensitive data and intellectual property.

## Summary of changes
1. NDAVersion Entity
Defined an entity for NDA versions, including the following fields:
```
id: Primary key
version: String representing the version of the NDA
documentText: Text of the NDA
createdAt: Timestamp of when the NDA version was created
updatedAt: Timestamp of when the NDA version was updated
```

2. NDA Entity
Defined an entity for the NDA that users will sign, referencing the NDAVersion entity:
```
id: Primary key
createdAt: Timestamp of when the NDA was signed
ipAddress: IP address of the user at the time of signing
userId: Relationship with the User entity
ndaVersionId: Relationship with the NDAVersion entity
```

3. NDA Endpoints
Implemented the following endpoints:

`GET /nda`: Retrieves the latest NDA version. If the user has already signed it, returns null. If not, returns the NDA version information. Uses JWT authentication.
`POST /nda/sign`: Allows the user to sign the NDA. Uses JWT authentication.

4. Added NDA Information to JWT
Added the NDA signing status to the JWT payload:
```
"NDA": "true" or "false"
```
If the latest NDA version is not signed by the user, set it to "false".

## How test the changes
`yarn test`

## Related issues
[Reputation Oracle] NDA
[#2148](https://github.com/humanprotocol/human-protocol/issues/2148)


packages/apps/reputation-oracle/server/src/database/database.module.ts
```diff
@@ -18,6 +18,8 @@ import { ServerConfigService } from '../common/config/server-config.service';
 import { SiteKeyEntity } from '../modules/user/site-key.entity';
 import { CredentialValidationEntity } from '../modules/credentials/credential.entity';
 import { CredentialEntity } from '../modules/credentials/credential.entity';
+import { NDAEntity } from 'src/modules/nda/nda.entity';
```
flopez7, line 21:
```suggestion
import { NDAEntity } from '../../modules/nda/nda.entity';
```

packages/apps/reputation-oracle/server/src/database/database.module.ts
```diff
@@ -18,6 +18,8 @@ import { ServerConfigService } from '../common/config/server-config.service';
...
+import { NDAVersionEntity } from 'src/modules/nda/nda-version.entity';
```
flopez7, line 22:
```suggestion
import { NDAVersionEntity } from '../../modules/nda/nda-version.entity';
```

flopez7 (changes requested):

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.ts
```diff
@@ -0,0 +1,89 @@
+import { HttpStatus, Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { NDARepository } from './nda.repository';
+import { NDAEntity } from './nda.entity';
+import { UserEntity } from '../user/user.entity';
+import { NDAVersionRepository } from './nda-version.repository';
+import { ControlledError } from '../../common/errors/controlled';
+import { ErrorNda } from '../../common/constants/errors';
+import { NdaVersionDto } from './nda.dto';
+import { NDAVersionEntity } from './nda-version.entity';
```
portuu3, line 10:
```suggestion
```

packages/apps/reputation-oracle/server/src/modules/nda/nda.repository.ts
```diff
@@ -0,0 +1,29 @@
...
+import { UserEntity } from '../user/user.entity';
+import { NDAVersionEntity } from './nda-version.entity';
+
+@Injectable()
+export class NDARepository extends BaseRepository<NDAEntity> {
+  constructor(private dataSource: DataSource) {
+    super(NDAEntity, dataSource);
+  }
+
+  async getLastNDA(user: UserEntity): Promise<NDAEntity | null> {
```
portuu3, line 14: not referenced

portuu3 (changes requested):

eugenvoronov:
Resolved conflicts

packages/apps/reputation-oracle/server/src/common/enums/nda.ts
```diff
@@ -0,0 +1,4 @@
+export enum NdaStatus {
+  SIGNED = 'SIGNED',
+  REJECTED = 'REJECTED',
```
portuu3, line 3: Regenerate migrations after this
```suggestion
  SIGNED = 'signed',
  REJECTED = 'rejected',
```

portuu3 (changes requested):
Solve conflicts

portuu3:
> Shared some suggestions async
> 
> General thought is that we don't have to store the NDA text in the DB and have the extra code to maintain that. Base on how accurate we want to be with the fact that user accepted the NDA, we can have different options, like:
> 
> * simple have `nda_accepted` prop in user model if we just want to have a fact about the "NDA checkbox" and display the NDA by the configurable link on UI
> * or have NDA document as part of our public repo for versioning and store "signature" in DB as a fact of user acceptance for further proofs
> 
> Within current implementation we can only show simple text (at max formatted with the markdown) and it wouldn't be nice to not manage it at all, but let legal people to prepare a doc and share it in a user-friendly format

I agree that the text could be stored somewhere other than the database (such as in an S3 bucket) but I don't entirely agree with the two proposed options:

1. Using a nda_accepted flag

Issue: This approach doesn't allow us to track if a user has signed a previous NDA or when he did so.
Maintenance Problem: Whenever the NDA text is modified, we would need to update this flag to false for all users.

With our current implementation, we could simply insert a new row into the nda_versions table.

2. Storing the NDA document in the repository

Deployment Overhead: Any changes to the NDA text would require going through the entire deployment process (e.g., creating pull requests, code reviews, etc.).
Loss of Traceability: We would lose track of which version each user signed.
Performance Issue: On each login, we would have to compute the hash of the current document and verify the user's signature to ensure they signed the latest version.


I propose the following changes to the current implementation:

- In the nda_versions table, replace storing the full text of the document with its hash and add a column for the URL where the document is stored (I suggest an S3 bucket).
- In the nda_signatures table, add a signature column to store the user's signature, enabling us to verify the hash of the signed document at any time.

dnechay:
@portuu3 

1. wrt first approach
> This approach doesn't allow us to track if a user has signed a previous NDA or when he did so

If we need to know the exact version and timestamp:
-  in case we don't need history: like a single prop with compound value: `user.nda_signature: "<version>_<timestamp>"`  or store it as base64 decoded json (or just json in postgres)
- in case we need to know full history - in a separate table

> Whenever the NDA text is modified, we would need to update this flag to false for all users

In case you operate on new version of NDA - you need to invalidate the old signature no matter how you store it

2. wrt second approach
> Deployment Overhead

That's true, so we can deny this option in favor of some "live" update

> On each login, we would have to compute the hash of the current document and verify the user's signature to ensure they signed the latest version.

You don't have to do it on each login, it can be optimized

Let's not dive into the deep discussion of these two "example approaches", there are many different ways to implement this. Before we jump into discussion related the "best fit solution", let's clarify nest questions about requirements:
- Do we need to store NDA versions on our end? Do we need RepO DB for that or maybe some 3rd party (Google Docs, GitBook, whatever)
- Do we need to store the history of signatures by some reason? Or is it enough for our legal purpose to know that user signed specific version?
- What information about signature we need? Is it "checkbox"-like or we need some kind of proof (e.g. sign something using private key)? What kind of proof is enough?
- Do we need to invalidate the signature in case new NDA version is there? If so (I suppose so), what would be the business flow for that? Should it be immediate or we allow some time for the previous "check" to expire? (if we store info like `isNdaSigned` in jwt - user will be able to solve jobs till toke expires)

portuu3:
This pr won't me merged, we will use a different approach. https://github.com/humanprotocol/human-protocol/issues/3105

Issue #2201, opened 2024-06-27
[Reputation Oracle] Convert addresses to lower case for signatures


portuu3:
**Description**
Convert addresses from body to lower case in `prepare-signature`, `web3/signup` and `web3/signin` endpoints

**Motivation**
Avoid case sensitive issues

PR #2211, opened 2024-06-29
[Reputation Oracle] Converted addresses to lower case for signatures

eugenvoronov:
## Description
Converted addresses from body to lower case in `prepare-signature`, `web3/signup` and `web3/signin` endpoints

## Summary of changes
- Implemented `LowerCaseAddressPipe`

## How test the changes
`yarn test`

## Related issues
[Reputation Oracle] Convert addresses to lower case for signatures
[#2201](https://github.com/humanprotocol/human-protocol/issues/2201)

### DIFF
--- a/packages/apps/reputation-oracle/server/src/common/pipes/index.ts
+++ b/packages/apps/reputation-oracle/server/src/common/pipes/index.ts
@@ -1,1 +1,2 @@
 export * from './validation';
+export * from './lower-case-address';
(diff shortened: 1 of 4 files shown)


portuu3 (commented):
I think is just simpler to do a toLowerCase when required. Imagine a case where the key is 'wallet_address' instead of 'address'.

portuu3 (changes requested):
Remove pipe

Issue #2206, opened 2024-06-28
[Fortune Exchange Oracle] Lower case addresses in signatures


portuu3:
**Description**
In the signature verification and sign message functions if the message contains an address convert it to lower case.

**Motivation**
Avoid case sensitive errors

**Implementation details** 
The conversion should be before calling the functions `verifySignature `and `signMessage`

PR #2213, opened 2024-06-29
[Fortune] Lower case addresses in signatures

eugenvoronov:
## Description
In the signature verification and sign message functions if the message contains a an address convert it to lower case.

## Summary of changes
Implemented `toLowerCase`.

## How test the changes
`yarn test`

## Related issues
[Fortune Exchange Oracle] Lower case addresses in signatures
[#2206](https://github.com/humanprotocol/human-protocol/issues/2206)
[Fortune Recording Oracle] Lower case addresses in signatures
[#2207](https://github.com/humanprotocol/human-protocol/issues/2207)


packages/apps/fortune/exchange-oracle/server/src/common/utils/index.ts
```diff
@@ -0,0 +1,7 @@
+export function toLowerCase(input: string | undefined): string {
```
portuu3, line 1: This doesn't make sense, just use native toLowerCase

portuu3 (changes requested):

packages/apps/fortune/exchange-oracle/server/src/common/guards/signature.auth.ts
```diff
@@ -27,7 +27,13 @@ export class SignatureAuthGuard implements CanActivate {
     );
     if (!roles) throw new NotImplementedException(ErrorSignature.MissingRoles);
     const request = context.switchToHttp().getRequest();
-    const data = request.body;
+    const { escrow_address, ...restBody } = request.body;
```
portuu3, line 30: We should the lower case conversion in the signature verification (method verifySignature). If we do it here and we call verifySignature from somewhere else the issue will persist. So this changes here are not required

packages/apps/fortune/recording-oracle/src/modules/job/job.service.spec.ts
```diff
@@ -136,7 +136,7 @@ describe('JobService', () => {
       (EscrowClient.build as jest.Mock).mockResolvedValue(escrowClient);
 
       const jobSolution: WebhookDto = {
-        escrowAddress: MOCK_ADDRESS,
+        escrowAddress: MOCK_ADDRESS.toLowerCase(),
```
portuu3, line 139: If the implementation is correct we should not requiered to pass lowercase in the tests, in should be done internally.

portuu3 (changes requested):

Dzeranov:
@eugenvoronov @portuu3 
Why can't we just modify a bit the structure here?
As far as I can see, we're actually verifying a specific data structure:
```
{
  chain_id: ChainId,
  escrow_address: string,
  event_type: EventType,
  event_data?: EventData
}
```
So we actually don't need to apply any `.stringify()` before data comes to `.verifySignature()`.
In this case we can specify an exact type of data/message and make all the necessary case conversions inside e.g.

```
export function verifySignature(
  data: WebhookDto,
  signature: string,
  addresses: string[],
): boolean {
  // set escrow_address to lower case
  // stringify a message
  const signer = recoverSigner(message, signature);

  ...
  }

  return true;
}
```

portuu3 (dismissed):

portuu3 (approved):

PR #2240, opened 2024-07-05
[Dashboard Backend] Init

Dzeranov:
## Description

Dashboard backend init commit

## Summary of changes

Added dashboard backend server with Redis as a caching solution

### DIFF
--- a/packages/apps/dashboard/server/src/modules/redis/redis.service.ts
+++ b/packages/apps/dashboard/server/src/modules/redis/redis.service.ts
@@ -1,0 +1,29 @@
+import { Inject, Injectable, Logger } from '@nestjs/common';
+import { Cache, CACHE_MANAGER } from '@nestjs/cache-manager';
+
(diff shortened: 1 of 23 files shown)


packages/apps/dashboard/server/src/modules/redis/redis.service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  async getHmtPrice(): Promise<number> {
+    const cachedHmtPrice: number | undefined = await this.cacheManager.get(
+      this.redisConfigService.hmtPriceCacheKey,
+    );
+    return cachedHmtPrice;
+  }
+
+  async setHmtPrice(hmtPrice: number): Promise<void> {
+    this.logger.log(`Setting a new HMT price: ${hmtPrice}`);
```
portuu3, line 22: makes sense to log this for every update?
Maybe would make sense to save on Redis also the last date it was updated so that we can check when was the last update to make sure it is working fine.

portuu3 (approved):

PR #2272, opened 2024-07-11
[CVAT][Exchange/Recording Oracles] Add encryption support

Marishka17:
## Description

This PR enables decryption of downloaded escrow manifests that Job Launcher encryptes
## Summary of changes
- Python SDK:
    - Added `EncryptionUtils.is_encrypted` method to check whether data is encrypted or not
- CVAT Oracles:
   - Set Oracle `PGP_PUBLIC_KEY_URL` to `KVStore` when starting service if there is no such key URL in KVStore or the key/URL has changed
    - Check whether escrow manifest is encrypted and decrypt them using `PGP_PRIVATE_KEY`/`PGP_PASSPHRASE` if necessary
    - PR adds 3 new environment variables (per oracle) to handle encryption: `PGP_PRIVATE_KEY`, `PGP_PASSPHRASE`, `PGP_PUBLIC_KEY_URL`

## How test the changes
Added unit tests for Exchange/Recording Oracles and for python SDK

## Related issues


packages/examples/cvat/exchange-oracle/src/core/config.py
```diff
@@ -207,3 +254,27 @@ class Config:
...
+    @classmethod
+    def validate(cls) -> None:
+        for attr_or_method in cls.__dict__:
+            attr_or_method = getattr(cls, attr_or_method)
+            if inspect.isclass(attr_or_method) and issubclass(attr_or_method, _BaseConfig):
+                attr_or_method.validate()
+
+    @classmethod
+    def get_network_configs(
+        cls, only_configured: Optional[bool] = None
```
zhiltsov-max, line 268: I suggest that this parameter is set to true by default and accepts only true and false. False returns all, and true returns only configured. Currently, there is only 1 use case for the function, which needs this parameter set to true. I don't see any use cases for false and none distinction here at the moment.

zhiltsov-max (dismissed):

Dzeranov:
Lgtm.
Once we prepare a staging environment for these changes (public key generation, etc.) will approve and merge.

Dzeranov (approved):

PR #2395, opened 2024-08-16
[Exchange Oracle] Add ruff

Bobronium:
## Description
Adds [ruff](https://github.com/astral-sh/ruff) and integrates linting into the CI pipeline.

Ruff combines features from `flake8` (with plugins), `pylint`, `isort`, `black`, `autopep8`, `flynt`, etc., into one extremely fast and well-maintained package.

One of the main features is availability of automatic fixes for some violations. 

Relying on a tool like Ruff helps to reduce cognitive load when writing/reading the code by enforcing consistency.

## Summary of changes
I've adopted the following strategy for integrating Ruff:
- Reformatted the codebase.
- Configured Ruff to report all violations.
- Applied safe fixes each in its own commit.
- Applied unsafe fixes each in its own commit.
- Handled the most trivial violations manually.
- Configured Ruff to ignore the noisiest/questionable violations.
- Used `# noqa` for remaining violations where necessary.

Each commit with autofix has a comment explaining the reasoning behind the changes. 

## How to test the changes
The code should be semantically equivalent, with only stylistic and formatting changes. I've ensured that all tests pass.

## References
After this PR and https://github.com/humanprotocol/human-protocol/pull/2396 will get merged, I'll introduce a PR adding `.git-blame-ignore-revs` with revisions from merge commits, so these changes will be ignored by git blame for better history overview.


packages/examples/cvat/exchange-oracle/.pre-commit-config.yaml
```diff
@@ -1,11 +1,17 @@
...
-          language_version: python3.11
-  -   repo: https://github.com/PyCQA/isort
-      rev: 5.12.0
-      hooks:
-        - id: isort
-          language_version: python3.11
\ No newline at end of file
+        - id: lint
+          name: lint
+          entry: ruff check --fix --unsafe-fixes --show-fixes
```
zhiltsov-max, line 6: Why unsafe fixes are enabled?

packages/examples/cvat/exchange-oracle/pyproject.toml
```diff
@@ -28,19 +28,104 @@ pyinstrument = "^4.6.2"
...
+]
+ignore = [
+    "ANN101",  # method args annotations (mypy will take care of that)
+    "ANN001",  # |
+    "ANN202",  # |
+    "ANN201",  # |
+    "ANN401",  # |
+    "ANN102",  # |
+    "RUF001",  # Allow cyrillic letters in comments
+    "COM812",  # Trailing comas are handled by ruff format
```
zhiltsov-max, line 52: If it's handled by ruff formatting, why it's ignored?

packages/examples/cvat/exchange-oracle/pyproject.toml
```diff
@@ -28,19 +28,104 @@ pyinstrument = "^4.6.2"
...
+    "ISC001",  # Recommended to be disabled when using ruff format
+    "B904",    # Raise from: modern pythons preserve previous exceptions
+    "EM",      # Forbids using literal strings in exceptions. I think it's ok
```
zhiltsov-max, line 55:
```suggestion
    "EM",      # Forbids using literal strings in exceptions. I think it's ok
```
I think there should not be any "I" in the source code comments.

packages/examples/cvat/exchange-oracle/pyproject.toml
```diff
@@ -28,19 +28,104 @@ pyinstrument = "^4.6.2"
...
+    "TRY003",  # Forbids long messages in exceptions, again, it's ok
+    "G004",    # Logging statement uses f-string — it's ok, f-strings are fast
```
zhiltsov-max, line 57: The comment here looks strange. If "it's ok" and "f-strings are fast", why it's in the ignored list? It would be better to say that it's not a problem for the project, if logging is not optimized.

packages/examples/cvat/exchange-oracle/pyproject.toml
```diff
@@ -28,19 +28,104 @@ pyinstrument = "^4.6.2"
...
+    "TD001",   # |
+    "TD002",   # |
+    "TD003",   # |
+    "E711",    # Allow == None comparisons for sqlalchemy queries
+    "E712",    # Allow == True comparisons for sqlalchemy queries
+    "PERF203", # Noisy microoptimisation
+    # Want to resolve eventually, but not now:
+    "S101",    # Allow asserts (there are too many of them right now to fix)
+    "TRY401",  # Checks for excesive logging of exception objects
+    "E501",    # Too long lines
```
zhiltsov-max, line 70: There should be no such lines, as it's enforced by the code formatter. Except some comments, which are an exception in its configuration (e.g. urls).

packages/examples/cvat/exchange-oracle/pyproject.toml
```diff
@@ -28,19 +28,104 @@ pyinstrument = "^4.6.2"
...
+# alembic is not a package in a traditional sense, so putting __init__.py there doesn't make sense
+"alembic/*" = ["INP001"]
+
+[tool.ruff.lint.pep8-naming]
+classmethod-decorators = [
+    "pydantic.validator",
+]
+
+[tool.ruff.lint.pylint]
+max-args = 9  # in the name of stupidity and lazyness, please god forgive me
```
zhiltsov-max, line 122: Please remove meaningless comments.

packages/examples/cvat/exchange-oracle/src/crons/process_recording_oracle_webhooks.py
```diff
@@ -83,29 +81,22 @@ def handle_recording_oracle_event(webhook: Webhook, *, db_session: Session, logg
...
+                            f"ignoring (escrow_address={webhook.escrow_address}, project={project.cvat_id})"
                         )
                         return
 
                     new_status = ProjectStatuses.recorded
                     logger.info(
-                        "Changing project status to {} (escrow_address={}, project={})".format(
-                            new_status, webhook.escrow_address, project.cvat_id
-                        )
+                        f"Changing project status to {new_status} (escrow_address={webhook.escrow_address}, project={project.cvat_id})"
```
zhiltsov-max, line 91: It seems like the fixes actually introduce too long lines. I'd also say that there are cases where f-strings are harder to read and/or maintain (e.g. with long variable names or field access in substitutions, with many substitutions).

packages/examples/cvat/exchange-oracle/src/crons/process_recording_oracle_webhooks.py
```diff
@@ -83,29 +81,22 @@ def handle_recording_oracle_event(webhook: Webhook, *, db_session: Session, logg
...
                     )
                     cvat_db_service.update_project_status(db_session, project.id, new_status)
 
         case RecordingOracleEventTypes.task_rejected:
             event = RecordingOracleEvent_TaskRejected.parse_obj(webhook.event_data)
 
             rejected_jobs = cvat_db_service.get_jobs_by_cvat_id(db_session, event.rejected_job_ids)
-            rejected_project_cvat_ids = set(j.cvat_project_id for j in rejected_jobs)
+            rejected_project_cvat_ids = {j.cvat_project_id for j in rejected_jobs}
```
zhiltsov-max, line 99: I don't really like this specific change for set declarations. With `set` it's impossible to confuse the declaration with a dict, so the code intention is clear. With `{}` you need to actually parse the expression to understand what's inside the variable, and you don't have any information about the author intentions for the variable.

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -509,13 +509,13 @@ def delete_cloudstorage(cvat_id: int) -> None:
...
 
-def fetch_projects(assignee: str = "") -> List[models.ProjectRead]:
+def fetch_projects(assignee: str = "") -> list[models.ProjectRead]:
     logger = logging.getLogger("app")
     with get_api_client() as api_client:
         try:
             return get_paginated_collection(
                 api_client.projects_api.list_endpoint,
-                **(dict(assignee=assignee) if assignee else {}),
+                **({"assignee": assignee} if assignee else {}),
```
zhiltsov-max, line 518: While correct, the previous variant is useful in the sense that it makes changing from a dict to a class as simple as `dict` renaming the class name. With {} you actually need to replace all the quotes and : to make such a transition.

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -615,7 +615,7 @@ def remove_user_from_org(user_id: int):
     with get_api_client() as api_client:
         try:
             (page, _) = api_client.users_api.list(
-                filter='{"==":[{"var":"id"},"%s"]}' % (user_id,),
+                filter=f'{{"==":[{{"var":"id"}},"{user_id}"]}}',
```
zhiltsov-max, line 618: I can't say it's easier to read now after the change. The {} are a part of the syntax here in the filter expression, so it makes total sense not to use f-strings here.

packages/examples/cvat/exchange-oracle/src/db/errors.py
```diff
@@ -3,6 +3,5 @@
 if db_engine.driver != "psycopg2":
     raise NotImplementedError
 
-from psycopg2.errors import LockNotAvailable
```
zhiltsov-max, line 6: This import is supposed for db driver abstraction. I suggest to add `__all__ = [LockNotAvailable]` instead of removing.

packages/examples/cvat/exchange-oracle/src/endpoints/middleware.py
```diff
@@ -44,8 +45,7 @@ async def profile_request(request: Request, call_next: Callable):
         response_data = profiler.output(renderer=renderer)
         if profile_type == "html":
             return HTMLResponse(response_data)
-        else:
-            return StreamingResponse(iter(response_data))
+        return StreamingResponse(iter(response_data))
```
zhiltsov-max, line 48: There are several similar places where the same comment would fit. I'd say that while correct, the early return pattern is typically used either for an optimization or for secondary use-cases. In this case both branches have equal priority, so I'd prefer using just a normal if-else.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -37,6 +35,12 @@
 from src.utils.assignments import parse_manifest
 from src.utils.logging import NullLogger, get_function_logger
 
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+    from logging import Logger
+
+    from src.core.manifest import TaskManifest
```
zhiltsov-max, line 42: What is the point of using type checking here? I don't think these imports have some special requirements.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -1234,16 +1210,14 @@ def _draw_roi_point(
             (255, 255, 255),
             cv2.FILLED,
         )
-        roi_pixels = cv2.circle(
+        return cv2.circle(
```
zhiltsov-max, line 1213: I think in some cases this optimization makes it a little bit harder to understand what's being returned.

packages/examples/cvat/exchange-oracle/src/services/cloud/types.py
```diff
@@ -31,24 +30,22 @@ def from_str(cls, provider: str) -> CloudProviders:
...
 
         if (config.access_key or config.secret_key) and config.provider.lower() != "aws":
             raise ValueError(
                 "Invalid storage configuration. The access_key/secret_key pair"
                 f"cannot be specified with {config.provider} provider"
             )
-        elif (
-            bool(config.access_key) ^ bool(config.secret_key)
-        ) and config.provider.lower() == "aws":
+        if (bool(config.access_key) ^ bool(config.secret_key)) and config.provider.lower() == "aws":
```
zhiltsov-max, line 48: This transition makes it look like it is else, while it's not. I suggest to add 1 empty line between the blocks to visually separate them.

packages/examples/cvat/exchange-oracle/src/utils/annotations.py
```diff
@@ -234,7 +236,7 @@ def __init__(
                 dst_label_cat = dm.LabelCategories(attributes=deepcopy(src_label_cat.attributes))
 
                 for dst_label in dst_labels:
-                    assert isinstance(dst_label, str) or isinstance(dst_label, tuple)
+                    assert isinstance(dst_label, str | tuple)
```
zhiltsov-max, line 239:
```suggestion
                    assert isinstance(dst_label, (str, tuple))
```

packages/examples/cvat/exchange-oracle/tests/api/test_exchange_api.py
```diff
@@ -38,7 +38,9 @@ def test_empty_list_tasks_200_without_address(client: TestClient) -> None:
 
 
 def test_list_tasks_200_with_address(client: TestClient) -> None:
-    with (SessionLocal.begin() as session,):
+    with (
+        SessionLocal.begin() as session,
+    ):
```
zhiltsov-max, line 43:
```suggestion
    with SessionLocal.begin() as session:
```

packages/examples/cvat/exchange-oracle/pyproject.toml
Bobronium: > There should be no such lines, as it's enforced by the code formatter.

Code formatter cannot do anything with comments and strings without changing AST, and it's something that it doesn't do, AFAIK.

> Except some comments, which are an exception in its configuration (e.g. urls).

I'm not exactly sure I follow you here regarding configuration and urls. If you imply there is/should be some way of configuring exceptions in formatter, I believe I've touched upon how this works above.

Since there are quite many exceptions where formatter is unable to step in without potentially changing semantics, I decided to ignore this rule for now. 

<details>
<summary>Existing violations:</summary>

```bash
 $ ruff check . --select E501 --output-format concise
src/crons/process_recording_oracle_webhooks.py:84:101: E501 Line too long (124 > 100)
src/crons/process_recording_oracle_webhooks.py:85:101: E501 Line too long (108 > 100)
src/crons/process_recording_oracle_webhooks.py:91:101: E501 Line too long (135 > 100)
src/crons/process_recording_oracle_webhooks.py:110:101: E501 Line too long (124 > 100)
src/crons/process_recording_oracle_webhooks.py:111:101: E501 Line too long (111 > 100)
src/crons/process_recording_oracle_webhooks.py:130:101: E501 Line too long (135 > 100)
src/crons/state_trackers.py:127:101: E501 Line too long (146 > 100)
src/crons/state_trackers.py:152:101: E501 Line too long (154 > 100)
src/handlers/completed_escrows.py:101:101: E501 Line too long (123 > 100)
src/handlers/job_creation.py:238:101: E501 Line too long (120 > 100)
src/handlers/job_creation.py:506:101: E501 Line too long (124 > 100)
src/handlers/job_creation.py:524:101: E501 Line too long (112 > 100)
src/handlers/job_creation.py:551:101: E501 Line too long (124 > 100)
src/handlers/job_creation.py:559:101: E501 Line too long (145 > 100)
src/handlers/job_creation.py:668:101: E501 Line too long (112 > 100)
src/handlers/job_creation.py:704:101: E501 Line too long (123 > 100)
src/handlers/job_creation.py:711:101: E501 Line too long (115 > 100)
src/handlers/job_creation.py:833:101: E501 Line too long (121 > 100)
src/handlers/job_creation.py:941:101: E501 Line too long (145 > 100)
src/handlers/job_creation.py:963:101: E501 Line too long (126 > 100)
src/handlers/job_creation.py:1020:101: E501 Line too long (111 > 100)
src/handlers/job_creation.py:1222:101: E501 Line too long (103 > 100)
src/handlers/job_creation.py:1564:101: E501 Line too long (112 > 100)
src/handlers/job_creation.py:1607:101: E501 Line too long (132 > 100)
src/handlers/job_creation.py:1693:101: E501 Line too long (111 > 100)
src/handlers/job_creation.py:1718:101: E501 Line too long (126 > 100)
src/handlers/job_creation.py:1726:101: E501 Line too long (139 > 100)
src/handlers/job_creation.py:1888:101: E501 Line too long (133 > 100)
src/handlers/job_creation.py:2008:101: E501 Line too long (152 > 100)
src/handlers/job_creation.py:2031:101: E501 Line too long (127 > 100)
src/handlers/job_creation.py:2409:101: E501 Line too long (105 > 100)
src/handlers/job_creation.py:2410:101: E501 Line too long (107 > 100)
tests/integration/chain/test_kvstore.py:162:101: E501 Line too long (114 > 100)
tests/integration/chain/test_kvstore.py:172:101: E501 Line too long (109 > 100)
tests/integration/cron/test_process_job_launcher_webhooks.py:303:101: E501 Line too long (105 > 100)
tests/integration/cron/test_process_recording_oracle_webhooks.py:196:101: E501 Line too long (103 > 100)
tests/integration/services/test_webhook.py:86:101: E501 Line too long (104 > 100)
tests/integration/services/test_webhook.py:161:101: E501 Line too long (104 > 100)
tests/utils/constants.py:26:101: E501 Line too long (146 > 100)
tests/utils/constants.py:35:101: E501 Line too long (159 > 100)
Found 40 errors.
```
</details>

zhiltsov-max: Ok, but there are cases where the lines (strings, actually) became long because of changes in this PR.

packages/examples/cvat/exchange-oracle/src/crons/process_recording_oracle_webhooks.py
Bobronium: Ah, I see. Ruff `UP032` seem to ignore the line-length requirement since https://github.com/astral-sh/ruff/pull/10263 was introduced.

From my experience, f-string are almost always more readable than stings with `.format()`, when used correctly, 

Yet I tend to stick with `.format()` here just to preserve line-length compliance without changing everything by hand.

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
Bobronium: I see your point! Fortunately, `ruff` can take care of this case:

as per `PIE804`, this code
```py
MyClass(**{"aws_access_key_id": access_key})
```

automatically gets converted to
```py
MyClass(aws_access_key_id=access_key)
```

So developer won't have to deal with replacing quotes or braces by hand.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Bobronium: It's considered a good practice to keep imports used only in annotations in `if TYPE_CHECKING:` block, to clearly signify what code is actually used in runtime, not to mention reducing startup overhead and avoiding import cycles.

I wouldn't want to track these imports myself, but fortunately, `ruff` can take care of this automatically: by either moving imports unused in runtime to `if TYPE_CHECKING` block, or by moving import from this block, if import becomes used in runtime.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Bobronium: I'd argue that is should be signified by the function name and return annotation.

In this example, I don't see how removing intermediate variable is confusing.

Otherwise, to me it would make more sense to rename `_draw_roi_point` -> `_get_roi_pixels`. Though current name is clear enough, IMO.

packages/examples/cvat/exchange-oracle/src/services/cloud/types.py
Bobronium: Thanks for the suggestion. I've added an extra line to improve the visual separation.

> One more downside I can see here is that now once you decide to change logic in this condition, you might need to change it back to `elif` everywhere.

While that’s a valid point, I believe it's a small price to pay for the gain in clarity and readability.

Initially, I had the same reservations about following this rule of avoiding unnecessary `elif` statements, but after working with it for a while, I've found it to be more beneficial. 

Using unnecessary constructs like `elif` in combination with early `return` or `raise` can lead to confusion. It may prompt the reader to search for a scenario where this construct would make a difference, only to find that there isn't one. This wastes time and mental energy, which could be better spent understanding the actual logic of the code.

packages/examples/cvat/exchange-oracle/src/utils/annotations.py
Bobronium: Sorry, I don't see clear basis for this suggestion. This was not changed by accident, if that is what you're implying.

Since Python 3.10, UnionType can be used in `isinstance`/`issubclass` calls and this syntax is more concise and readable, IMO.

packages/examples/cvat/exchange-oracle/.pre-commit-config.yaml
Bobronium: Here's my reasoning behind enabling unsafe fixes:

I find them quite useful and relatively safe and don't see any reason for disabling the.

Only one change made via `unsafe-fixes` actually made unintended changes which I had to fix by marking this rule as `unfixable` globally.

We can always disable it if it will cause more troubles than use, but unless we'll try, we'll never know, but will be guaranteed to miss out on useful fixes.

packages/examples/cvat/exchange-oracle/pyproject.toml
Bobronium: Good point 😄. I've fixed existing violations and re-enabled the rule.

packages/examples/cvat/exchange-oracle/src/crons/process_recording_oracle_webhooks.py
Bobronium: Disabled `UP032` for now. Hope `ruff` will get a flag to restore previous behaviour.

packages/examples/cvat/exchange-oracle/tests/api/test_exchange_api.py
Bobronium: Fixed here and in other instance, thanks!

packages/examples/cvat/exchange-oracle/pyproject.toml
Bobronium: You're right, it was confusing indeed. Changed the wording completely.

packages/examples/cvat/exchange-oracle/pyproject.toml
Bobronium: Added other conflicting rules to that list and added clarification as well. 

See https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules.

packages/examples/cvat/exchange-oracle/.pre-commit-config.yaml
zhiltsov-max: I can see a problem in this, because if these fixes are applied, e.g. on commit automatically, or you're just applying it blindly, you can silently end up with a broken code. So, you can't "just format" the code, you must check all the changes.

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
zhiltsov-max: Nice, great feature.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
zhiltsov-max: From my knowledge, `TYPE_CHECKING` was added to support forward references and extra code for annotations specifically, then there such support was added in `from future import __annotations__`, now it makes sense to use it only for cyclic references between different files and heavy components (e.g. libraries with some initialization on import - hi, `tensorflow`), that are only used for annotations. It's hard to believe it's a considered a good practice to check all the code and search for imports that are only used for annotations.

packages/examples/cvat/exchange-oracle/src/endpoints/middleware.py
zhiltsov-max: I have no objections against someone writing their code this way, but I'm against enforcing it via automatic formatting.

packages/examples/cvat/exchange-oracle/src/utils/annotations.py
zhiltsov-max: I believe it's allowed, but why this specific place should use a different style from all the other isinstance checks?

packages/examples/cvat/exchange-oracle/.pre-commit-config.yaml
Bobronium: There was a confusion on whether `--unsafe-fixes` are committed automatically. After a discussion on a call, we decided to leave them on since commit is prevented in case of any changes present.

packages/examples/cvat/exchange-oracle/src/endpoints/middleware.py
Bobronium: Although, the point was exactly in enforcing one consistent style throughout the project and, so there would not be multiple ways of doing the same thing, I've disabled the rule and reverted the changes for now so we can merge this.

packages/examples/cvat/exchange-oracle/src/utils/annotations.py
Bobronium: It's the only place where multiple types are used with `isinstance`, hence the only place that has these changes.

packages/examples/cvat/exchange-oracle/src/crons/process_recording_oracle_webhooks.py
Bobronium: Reverted the changes and disabled the rule for now.

zhiltsov-max (approved):

PR #2408, opened 2024-08-19
feat: add events and cancel method for erc20 token support

### LINKED ISSUE #2381: [Smart contracts] Support ERC20 token tracking of escrow
**Description**
Add fund event/payout event to keep track of ERC20 tokens used as escrow balance.
Add cancel function with ERC20 token argument, to support withdrawing ERC20 tokens sent to the escrow.

**Motivation**
Now the subgraph can only update HMT balance information for the escrow. However, our escrows are designed to support any ERC20 token for the payments. It's not possible to setup transfer event handler for every single ERC20 token, so is better setting up our own events to monitor the ins/outs of the ERC20 token to the campaign.
In addition, we need the ability to withdraw ERC20 tokens sent to the escrow. There are some edge cases that user can send ERC20 tokens multiple times for deposit / send wrong ERC20 tokens, and we need to return the funds back to the user.

**Implementation details** 

- Add a new event to escrow contract and in setup function trigger it checking the balance of the escrow.
- Add a new event to bulk payout function, with the paid amount.
- Refactor the balance check logic relying on getBalance of ERC20 token.
- Add a new cancel function with custom ERC20 token argument.
- Update subgraph.

leric7:
## Description

- Added `Fund` and `BulkPayout` events to track balance of the escrow.
- Added `withdraw` function with custom token address.
- Added `withdraw` function to the Python/Typescript SDK.
- Updated `balance` track logic on subgraph.

## Related issues

Closes #2381


packages/core/contracts/Escrow.sol
```diff
@@ -150,11 +164,16 @@ contract Escrow is IEscrow, ReentrancyGuard {
...
         status = EscrowStatuses.Pending;
+
+        remainingFunds = getBalance();
+        require(remainingFunds > 0, 'Escrow balance is zero');
+
         emit Pending(manifestUrl, manifestHash);
+        emit Fund(remainingFunds);
     }
 
     function abort() external override trusted notComplete notPaid {
```
portuu3, line 175: Is it actually worth to keep abort? When would this be called since we already have cancel function?

packages/core/contracts/Escrow.sol
```diff
@@ -268,24 +305,26 @@ contract Escrow is IEscrow, ReentrancyGuard {
 
         for (uint256 i = 0; i < _recipients.length; ++i) {
             if (finalAmounts[i] > 0) {
-                _safeTransfer(_recipients[i], finalAmounts[i]);
+                _safeTransfer(token, _recipients[i], finalAmounts[i]);
+                emit Payout(_recipients[i], finalAmounts[i]);
```
portuu3, line 309: I think this payout event is not required, since we already have BulkTransfer which already contains this info

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/escrow/escrow_client.py
```diff
@@ -348,6 +362,7 @@ def create_and_setup_escrow(
         trusted_handlers: List[str],
         job_requester_id: str,
         escrow_config: EscrowConfig,
+        amount: Decimal,
```
portuu3, line 365: We should rename this function or even consider removing it since it involves a transfer.

packages/sdk/typescript/human-protocol-sdk/src/escrow.ts
```diff
@@ -949,7 +1059,13 @@ export class EscrowClient extends BaseEthersClient {
     try {
       const escrowContract = this.getEscrowContract(escrowAddress);
 
-      return escrowContract.getBalance();
+      try {
+        return await escrowContract.remainingFunds();
```
portuu3, line 1063: We should probably have 2 functions:
- getTotalBalance
- getRemainingFunds

packages/core/contracts/Escrow.sol
leric7: We still need `Payout` event for oracle payments to keep track of the escrow balance. Since we can't drop this event for that, I'd better use this event to make it easier to handle balance on subgraph.

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/escrow/escrow_client.py
leric7: I don't know how often we use this function. If you think we can remove this, I'll delete this one.

packages/sdk/typescript/human-protocol-sdk/src/escrow.ts
leric7: What's the difference between these two? I think we just need balance information. All the extra funds are just available for withdraw.

portuu3 (changes requested):

portuu3 (approved):

Issue #2415, opened 2024-08-20
[Dashboard][Server] Add reputation scores for the leaderboard


Dzeranov:
**Description**
Need to add reputation scores for the leaderboard endpoints:
`/details/leaders` and `/details/leaders/all`

**Motivation**
Reputation score is not on chain and should be retrieved from a reputation oracle `/reputation` endpoint.

PR #2416, opened 2024-08-20
[Dashboard][Server] Add reputation scores for the leaderboard

flopez7:
## Description

Need to add reputation scores for the leaderboard endpoints:
`/details/leaders` and `/details/leaders/all`

## Summary of changes

Add an env variable with reputation oracle url
Query reputation values from reputation oracle
Match each reputation value with their corresponding leader

## Related issues
#2415


portuu3 (changes requested):
Might be worth to update reputation endpoint from Reputation Oracle to filter by role (worker and operator). Also use chain id filter

portuu3 (approved):

portuu3 (approved):

PR #2428, opened 2024-08-22
[CVAT][Exchange Oracle] Data cleanup

Bobronium:
## Description
- Add support for `escrow_finished` event from `reputation_oracle`
- Introduce common abstractions for dealing with crons and webhooks
- Add data cleanup

#### **New**: handling webhooks from `reputation_oracle`
- `escrow_completed`
    - delete projects from CVAT
    - cleanup data in Exchange Oracle cloud storage
    - mark projects, tasks and jobs as `deleted`
    - send webhook to Recording Oracle with event `escrow_cleaned`

#### **Changed**: handling webhooks from `job_launcher`
- `escrow_created` handling failure:
  - before: 
      - delete project from CVAT
      - delete project from Exchange Oracle
  - now:
      - cleanup data in cloud storage
      - delete projects from CVAT
      - delete projects from Exchange Oracle
- `escrow_cancelled`:
  - before:
    - mark projects as cancelled
  - now:
    - delete projects from CVAT
    - cleanup data in Exchange Oracle cloud storage
    - mark projects as cancelled
    - send webhook to Recording Oracle with event `escrow_cleaned`

## How test the changes
New test module has been added as well as necessary changes to existing ones.

## Related issues


packages/examples/cvat/exchange-oracle/src/services/cloud/s3.py
```diff
@@ -8,6 +9,8 @@
 from src.services.cloud.client import StorageClient
 
 DEFAULT_S3_HOST = "s3.amazonaws.com"
+if TYPE_CHECKING:
+    from mypy_boto3_s3.client import S3Client
```
zhiltsov-max, line 13:
```suggestion
    from mypy_boto3_s3.client import _S3Client
```
Probably, it's not a good idea to have 2 different symbols with the same name in the same file.

packages/examples/cvat/exchange-oracle/src/services/cvat.py
```diff
@@ -207,23 +207,50 @@ def update_project_statuses_by_escrow_address(
...
+        )
+        session.execute(task_update_statement)
+
+    if jobs_status is not None:
+        job_update_statement = (
+            update(Job)
+            .where(Job.cvat_project_id.in_(updated_project_ids))
+            .values(status=jobs_status.value)
+        )
+        session.execute(job_update_statement)
```
zhiltsov-max, line 242: I'm not sure it should be a part of the same function with such a name. Consider splitting it into 3 functions or adding one which deletes an escrow.

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+        storage_client.remove_files(
+            [
+                *storage_client.list_files(
+                    prefix=compose_data_bucket_prefix(self.escrow_address, self.chain_id),
+                ),
+                *storage_client.list_files(
+                    prefix=compose_results_bucket_prefix(self.escrow_address, self.chain_id),
+                ),
+            ]
```
zhiltsov-max, line 62: I believe we can use a filter instead of passing thousands of file paths.

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
```diff
@@ -0,0 +1,63 @@
...
+
+@dataclass
+class EscrowCleaner:
+    escrow_address: str
+    chain_id: int
+    projects: list[Project]
+
+    def cleanup(self) -> None:
+        self._cleanup_cvat()
+        self._cleanup_storage()
```
zhiltsov-max, line 31: I believe we should perform all the steps regardless of errors on any of the steps. Maybe, re-raise the error later, if anything happened. Otherwise we can end up with a half-cleaned escrow.

packages/examples/cvat/exchange-oracle/src/crons/_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        logger = logging.getLogger(logger_name).getChild(fn.__name__)
+        cron_repr = repr(fn.__name__)
+
+        # validate signature
+        parameters = dict(inspect.signature(fn).parameters)
+        session_param = parameters.pop("session", None)
+        if session_param is not None:
+            assert (
+                session_param.annotation == Session
+            ), f"{cron_repr} session argument type muse be of type {Session.__qualname__}"
```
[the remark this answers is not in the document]
Bobronium: I currently use PyCharm and it has spellchecker built-in, but
muse is a correct English word, so spellchecker was cool with it :D 

Good catch, though!

packages/examples/cvat/exchange-oracle/src/services/cvat.py
Bobronium: After some consideration I've decided that there's indeed no need to change Job/Task status and reverted these changes.

packages/examples/cvat/exchange-oracle/src/services/cloud/s3.py
Bobronium: Ah, good catch!

packages/examples/cvat/exchange-oracle/src/crons/process_job_launcher_webhooks.py
```diff
@@ -91,20 +70,28 @@ def handle_job_launcher_event(webhook: Webhook, *, db_session: Session, logger:
...
-                    oracle_db_service.outbox.create_webhook(
-                        session=db_session,
-                        escrow_address=webhook.escrow_address,
-                        chain_id=webhook.chain_id,
-                        type=OracleWebhookTypes.job_launcher,
-                        event=ExchangeOracleEvent_TaskCreationFailed(reason=str(ex)),
-                    )
+                    # new session is required since exception might roll back the current session
+                    # creation of this webhook should not depend on this
+                    with SessionLocal.begin() as session:
```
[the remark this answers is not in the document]
Bobronium: I've added `on_fail` parameter to `handle_webhook` to pass `src.crons.webhooks.job_launcher.handle_failure` which will be called in case of exception. This way `handle_webhook` also is able to abstract session handling away from callee.

packages/examples/cvat/exchange-oracle/src/crons/_utils.py
```diff
@@ -0,0 +1,131 @@
...
+
+@contextmanager
+def handle_webhook(logger: logging.Logger, session: Session, webhook: Webhook):
+    savepoint = session.begin_nested()
+    logger.debug(
+        "Processing webhook "
+        f"{webhook.type}.{webhook.event_type}~{webhook.signature} "
+        f"in escrow_address={webhook.escrow_address} "
+        f"(attempt {webhook.attempts + 1})"
+    )
```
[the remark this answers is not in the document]
Bobronium: I'm not entirely sure, what you mean when you say that logging can fail in this case. Could you elaborate on that?

In any case, I've rearranged these statements since it seemed logical to me after you pointed that out.

packages/examples/cvat/exchange-oracle/src/crons/__init__.py
```diff
@@ -74,4 +75,9 @@ def cron_record():
             "interval",
             seconds=Config.cron_config.track_assignments_int,
         )
+        scheduler.add_job(
+            process_incoming_reputation_oracle_webhooks,
+            "interval",
+            seconds=Config.cron_config.track_finished_projects_int,
```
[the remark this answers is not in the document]
Bobronium: Dough! Yes, the name is left dangling from early implementation. Fixed!

packages/examples/cvat/exchange-oracle/src/core/oracle_events.py
```diff
@@ -44,13 +43,23 @@ class ExchangeOracleEvent_TaskFinished(OracleEvent):
     pass  # escrow is enough for now
 
 
+class ExchangeOracleEvent_EscrowCleaned(OracleEvent):
+    pass  # escrow is enough for now
+
+
+class ReputationOracleEvent_EscrowCompleted(OracleEvent):  # (will refactor later)
```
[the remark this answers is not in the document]
Bobronium: I guess I was pointing out `_` in the class name which is prohibited by `PEP-8` naming convention back when `ruff` was complaining about that.

I've removed the comment for now, I believe we can deal with this later.

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
Bobronium: The reasoning behind this way of handling things was to make sure that everything that might use these files (for instance CVAT), would be cleared first, so there would not be a condition when files are still referenced and used somewhere after they've been deleted from the bucket. 

If that's not a problem, of course we can clean them regardless. Any thoughts on this?

packages/examples/cvat/exchange-oracle/src/crons/_utils.py
```diff
@@ -0,0 +1,131 @@
...
+import src.services.webhook as oracle_db_service
+import src.services.webhook as webhook_service
+from src.core.types import OracleWebhookTypes
+from src.db import SessionLocal
+from src.db.utils import ForUpdateParams
+from src.models.webhook import Webhook
+from src.utils.webhooks import prepare_outgoing_webhook_body, prepare_signed_message
+
+
+def cron_job(logger_name: str) -> Callable[[Callable[..., None]], Callable[[], None]]:
```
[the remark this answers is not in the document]
Bobronium: As we discussed personally, leaving this a decorator.

I've extracted validation code from the decorator to make it more readable, and removed parameters in favour of automatic logging configuration.

packages/examples/cvat/exchange-oracle/src/crons/_utils.py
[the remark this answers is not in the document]
zhiltsov-max: The logging call, as any other call, can fail. In the implementation variant being commented, an exception in the logging call could lead to a session object not being closed properly.

packages/examples/cvat/exchange-oracle/src/core/oracle_events.py
```diff
@@ -44,13 +43,23 @@ class ExchangeOracleEvent_TaskFinished(OracleEvent):
     pass  # escrow is enough for now
 
 
+class ExchangeOracleEvent_EscrowCleaned(OracleEvent):
+    pass  # escrow is enough for now
```
zhiltsov-max, line 47:
```suggestion
    pass
```

With such an event name, I wouldn't expect any extra parameters needed.

packages/examples/cvat/exchange-oracle/src/core/types.py
```diff
@@ -72,6 +75,11 @@ class RecordingOracleEventTypes(str, Enum, metaclass=BetterEnumMeta):
     task_rejected = "task_rejected"
 
 
+class ReputationOracleEventTypes(str, Enum, metaclass=BetterEnumMeta):
+    # TODO: rename to ReputationOracleEventType
```
zhiltsov-max, line 79: What is this comment? Do you want to rename this specific enum?

packages/examples/cvat/exchange-oracle/src/crons/_cron_job.py
```diff
@@ -0,0 +1,71 @@
...
+def _validate_cron_function_signature(fn: Callable[..., None]) -> CronSpec:
+    cron_repr = repr(fn.__name__)
+    parameters = dict(inspect.signature(fn).parameters)
+
+    session_param = parameters.pop("session", None)
+    if session_param is not None and session_param.annotation is not Session:
+        raise TypeError(f"{cron_repr} session argument type must be of type {Session.__qualname__}")
+
+    logger_param = parameters.pop("logger", None)
+    if logger_param is None or logger_param.annotation is not logging.Logger:
```
zhiltsov-max, line 27: Can it be a logger subclass?

packages/examples/cvat/exchange-oracle/src/crons/_cron_job.py
```diff
@@ -0,0 +1,71 @@
...
+    manage_session: bool
+    repr: str
+
+
+def _validate_cron_function_signature(fn: Callable[..., None]) -> CronSpec:
+    cron_repr = repr(fn.__name__)
+    parameters = dict(inspect.signature(fn).parameters)
+
+    session_param = parameters.pop("session", None)
+    if session_param is not None and session_param.annotation is not Session:
```
zhiltsov-max, line 23: Can it be a session subclass?

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
zhiltsov-max: No, I don't think it's a problem. There should be no access to the project after it's marked deleted. It can be considered an error, if there is any access happening, it's not expected to work.

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
```diff
@@ -0,0 +1,52 @@
...
+    deleted_cloud_storage_ids = set()
+    for project in projects:
+        if (
+            project.cvat_cloudstorage_id is not None
+            and project.cvat_cloudstorage_id not in deleted_cloud_storage_ids
+        ):
+            with contextlib.suppress(NotFoundException):
+                # probably will allways call this just once
+                cvat_api.delete_cloudstorage(project.cvat_cloudstorage_id)
+            deleted_cloud_storage_ids.add(project.cvat_cloudstorage_id)
```
zhiltsov-max, line 31: Probably, it's better to reorder calls here, so that CS ids are collected first, then removed. Otherwise it's possible that a project depending on a CS in CVAT has no CS link. Probably, it should be controlled by CVAT and a 400 error should be raised on such a removal attempt. A CS should only be removed when there are no projects using it.

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
```diff
@@ -0,0 +1,52 @@
...
+    compose_data_bucket_prefix,
+    compose_results_bucket_prefix,
+)
+from src.services.cloud.utils import BucketAccessInfo
+
+if TYPE_CHECKING:
+    from src.models.cvat import Project
+
+
+def _cleanup_cvat(projects: list[Project]) -> None:
```
zhiltsov-max, line 21: It's possible to get some timeout or 5xx errors here, maybe it's better to try to remove all of the projects anyway.

packages/examples/cvat/exchange-oracle/src/crons/_cron_job.py
Bobronium: No. It wouldn't make sense to allow annotation of subclasses here since `cron_job` supplies `logging.Logger`, not its subclasses.

packages/examples/cvat/exchange-oracle/src/crons/_cron_job.py
Bobronium: No, see https://github.com/humanprotocol/human-protocol/pull/2428#discussion_r1740695273

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
Bobronium: You mean remove files from cloud storage?

packages/examples/cvat/exchange-oracle/src/core/types.py
Bobronium: All of them, really. It's unconventional for enums to have plural name, however to be consistent with current code base I went with it for now.

As for renaming it in this PR, I didn't want to introduce unnecessary noise which would be inevitable, if I was to rename all of the enums.

zhiltsov-max: Which naming convention do you mean? This project currently uses plurals for enum names.

Bobronium: I wasn't referencing any specific convention within our project. Rather, I was referring to the general standard in Python (and other languages for that matter), where classes, including enums, are typically named using singular nouns in `PascalCase`. 

You can see example of following this convention here: https://github.com/python/cpython/issues/104659.

I understand that our codebase currently uses plurals for enum names (and seen the change in history from once singular variant), so I've maintained that consistency for now. However, considering the broader community standards, we might want to think about transitioning back to singular names. This would make our code more intuitive for new contributors who are familiar conventions. Hence the `# TODO`.

As I said, I chose not to make this change in the current PR to avoid introducing unnecessary  changes.

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
Bobronium: Done. Errors will be ignored during deletion, but will be raised later in order to signal that something went wrong, so webhook processing can be retried later.

packages/examples/cvat/exchange-oracle/src/handlers/escrow_cleanup.py
```diff
@@ -0,0 +1,95 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from src.models.cvat import Project
+
+logger = logging.getLogger(get_logger_name(__name__))
+
+
+@contextlib.contextmanager
+def log_error(errors_container: list[Exception], message: str) -> Generator[None, None, None]:
```
zhiltsov-max, line 28:
```suggestion
def _log_error(errors_container: list[Exception], message: str) -> Generator[None, None, None]:
```

Probably, it was supposed for internal use in this component.

zhiltsov-max (dismissed):

zhiltsov-max (approved):

PR #2442, opened 2024-08-26
[CVAT][Recording Oracle] Handle escrow_cleaned event

Bobronium:
## Description
- Add support for `escrow_cleaned` event from `exchange_oracle`
- Introduce common abstractions for dealing with crons and webhooks
- Add data cleanup

#### Handling new event type in `exchange_oracle` webhooks
- `escrow_cleaned`
    - cleanup data in Exchange Oracle cloud storage
    - Mark escrows tasks with `cleaned_at` (new column)

## How test the changes
New test case has been added.

## Related issues
#2428


packages/examples/cvat/recording-oracle/src/crons/process_exchange_oracle_webhooks.py
```diff
@@ -1,135 +1,61 @@
...
-
             validate_results(
                 escrow_address=webhook.escrow_address,
                 chain_id=webhook.chain_id,
                 db_session=db_session,
             )
-
+        case ExchangeOracleEventTypes.escrow_cleaned:
+            clean_escrow(escrow_address=webhook.escrow_address, chain_id=webhook.chain_id)
+            mark_escrow_as_cleared(
```
zhiltsov-max, line 46: 1. Probably, it should be a single function
2. Please be consistent about using clean and clear

packages/examples/cvat/recording-oracle/src/services/cloud/gcs.py
```diff
@@ -32,6 +32,11 @@ def remove_file(self, key: str, *, bucket: str | None = None) -> None:
         bucket_client = self.client.get_bucket(bucket)
         bucket_client.delete_blob(unquote(key))
 
+    def remove_files(self, keys: list[str], *, bucket: str | None = None):
+        bucket = unquote(bucket) if bucket else self._bucket
+        bucket_client = self.client.get_bucket(bucket)
+        bucket_client.delete_blobs([unquote(key) for key in keys])
```
zhiltsov-max, line 38: The same comment about using a filter instead. I'm pretty sure there is such an option.

Bobronium: Not to my knowledge. 
See https://cloud.google.com/storage/docs/deleting-objects#delete-objects-in-bulk

Since we're not using GCS for uploads right now, it's not really an issue right now, but I've added a warning in case it's going to be used some time.

zhiltsov-max: Discussed, GCS wrapper can just provide an unoptimized version.

zhiltsov-max (approved):

Dzeranov (approved):

Issue #2469, opened 2024-08-30
[Human App] User Qualification Filter


flopez7:
**Description**
Update Human App to filter jobs based on the user’s qualifications. The app receives all jobs from the Exchange Oracle without filtering, so it needs to filter out jobs that the user is not qualified to handle based on their qualifications.

**Motivation**
Filtering jobs based on user qualifications directly within the Human App ensures that users only see jobs they are able to handle. This enhances the user experience by preventing them from viewing jobs they cannot resolve, improving efficiency and satisfaction.

**Implementation Details**
- Retrieve user’s qualifications from the JWT.
- Implement a filtering mechanism that uses the user’s qualifications to filter the list of jobs received from the Exchange Oracle.
- Only display jobs to the user that match their qualifications.
- Test the filter functionality to confirm that jobs are correctly filtered based on user qualifications and that no jobs are incorrectly excluded or included.

PR #2481, opened 2024-09-02
[Human App] Jobs qualifications filter

flopez7:
## Description
Update Human App to filter jobs based on the user’s qualifications. The app receives all jobs from the Exchange Oracle without filtering, so it needs to filter out jobs that the user is not qualified to handle based on their qualifications

## Summary of changes

Add qualifications filter to Human App getAvailableJobs
Fix extra fields in getAvailableJobs endpoint

## How test the changes
`yarn test`

## Related issues
#2469

### DIFF
--- a/packages/apps/human-app/server/src/modules/jobs-discovery/jobs-discovery.controller.ts
+++ b/packages/apps/human-app/server/src/modules/jobs-discovery/jobs-discovery.controller.ts
@@ -38,7 +38,7 @@ export class JobsDiscoveryController {
     @JwtPayload() jwtPayload: JwtUserData,
     @Authorization() token: string,
   ): Promise<JobsDiscoveryResponse> {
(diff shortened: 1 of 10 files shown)


packages/apps/human-app/server/src/modules/jobs-discovery/jobs-discovery.controller.ts
```diff
@@ -41,6 +46,9 @@ export class JobsDiscoveryController {
         JobsDiscoveryParamsCommand,
       );
     jobsDiscoveryParamsCommand.token = token;
-    return await this.service.processJobsDiscovery(jobsDiscoveryParamsCommand);
+    return await this.service.processJobsDiscovery(
```
portuu3, line 49: Use JobsDiscoveryParamsCommand to pass the qualifications and follow the rules implemented in the whole application

portuu3 (changes requested):

portuu3 (approved):

PR #2483, opened 2024-09-03
[CVAT][Exchange Oracle] Avoid reassigning a job to the same wallet address

Bobronium:
## Description

> In the current implementation a job can be annotated by the same worker multiple times. It is especially bad, because it could be siently rejected previously and now the worker has to annotate the job one more time.

> Delivery: A job can be annotated by an annotator only once. If it is rejected, it should not be visible again by the same annotator.

## Summary of changes

- Updated the `create_assignment` endpoint to check for existing assignments for a given `wallet_address` and return a relevant message if none are available.
- Modified `get_free_job` to exclude jobs already assigned to the specified `wallet_address`.

## How test the changes
- Run integration tests in `test_cvat.py` and `test_exchange.py` to validate new behavior.
- Manually test the `create_assignment` endpoint by attempting to assign jobs to the same wallet address multiple times; ensure it blocks subsequent assignments correctly.

### DIFF
--- a/packages/examples/cvat/exchange-oracle/src/services/cvat.py
+++ b/packages/examples/cvat/exchange-oracle/src/services/cvat.py
@@ -515,17 +515,24 @@ def get_free_job(
     session: Session,
     cvat_projects: list[int],
     *,
+    user_wallet_address: str,
     for_update: bool | ForUpdateParams = False,
 ) -> Job | None:
+    """
+    Returns the first available job that wasn't previously assigned to that user_walled_address.
+    """
     return (
         _maybe_for_update(session.query(Job), enable=for_update)
         .where(
             Job.cvat_project_id.in_(cvat_projects),
             Job.status == JobStatuses.new,
             ~Job.assignments.any(
-                (Assignment.status == AssignmentStatuses.created.value)
-                & (Assignment.completed_at == None)
-                & (utcnow() < Assignment.expires_at)
+                (
+                    (Assignment.status == AssignmentStatuses.created.value)
+                    & (Assignment.completed_at == None)
+                    & (utcnow() < Assignment.expires_at)
+                )
+                | (Assignment.user_wallet_address == user_wallet_address),
             ),
         )
         .first()
--- a/packages/examples/cvat/exchange-oracle/tests/integration/services/test_exchange.py
+++ b/packages/examples/cvat/exchange-oracle/tests/integration/services/test_exchange.py
@@ -422,7 +422,45 @@ class ServiceIntegrationTest(unittest.TestCase):
 
         assert assignment_id == None
 
-    def test_create_assignment_in_validated_and_rejected_job(self):
+    def test_create_assignment_wont_reassign_job_to_previous_user(self):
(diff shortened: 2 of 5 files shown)


packages/examples/cvat/exchange-oracle/src/services/cvat.py
```diff
@@ -515,17 +515,24 @@ def get_free_job(
...
             ~Job.assignments.any(
-                (Assignment.status == AssignmentStatuses.created.value)
-                & (Assignment.completed_at == None)
-                & (utcnow() < Assignment.expires_at)
+                (
+                    (Assignment.status == AssignmentStatuses.created.value)
+                    & (Assignment.completed_at == None)
+                    & (utcnow() < Assignment.expires_at)
+                )
+                | (Assignment.user_wallet_address == user_wallet_address),
```
nmanovic, line 535: What if it is assigned to user1, after that to user2, and again to user1. Will user1 get it second time?

Bobronium: This part of expression roughly translates to: look for job that has no previous assignments with this wallet id. 

So, answering your question, no, once job has been assigned to a user, it will never be assigned to that user again.

packages/examples/cvat/exchange-oracle/tests/integration/services/test_exchange.py
```diff
@@ -457,10 +462,12 @@ def test_create_assignment_in_validated_and_rejected_job(self):
         ):
             manifest = json.load(data)
             mock_get_manifest.return_value = manifest
-            assignment_id = create_assignment(cvat_project_1.id, user_address)
+            assignment_id = create_assignment(cvat_project_1.id, previous_user.wallet_address)
+            assert assignment_id is None  # should not assign job to the same user more than once
```
zhiltsov-max, line 466: I suggest to add another test for this scenario.

zhiltsov-max (dismissed):

packages/examples/cvat/exchange-oracle/tests/integration/services/test_exchange.py
Bobronium: Can't imagine how separate test will be beneficial. It would've make sense if this scenario tested something different, but it in fact tests the opposite behaviour to the new one. 

So making changes the only thing that made sense.

I'm happy to add a new test case, but just want to get the clear picture of how it should look like and how should it be different from this one.

zhiltsov-max: Currently, from what I can see, this test checks that:
- a previous assignee cannot be assigned
- a new assignee can be assigned

In this PR we're covering a new requirement, so I suggest that there is a separate test, focused on this. The benefit is that we'll have a test checking this requirement specifically, it will be clear that this is a requirement from the test name and not from a comment.

Bobronium: A bit verbose to my taste, but I guess it makes sense indeed. Split these test-cases in two.

zhiltsov-max (approved):

Dzeranov (approved):

PR #2486, opened 2024-09-04
fix(human-app/frontend): layout improvments

KacperKoza343:
Contains fixes for all reported issues


packages/apps/human-app/frontend/src/i18n/en.json
```diff
@@ -199,7 +199,7 @@
       "emailNotifications": "Email Notifications",
       "notificationsConsent": "I would like to be notified about News and Offers",
       "topNotifications": {
-        "noKYC": "Please complete your profile! You still need to complete KYC, connect your wallet and add KYC info on-chain.",
+        "noKYC": "Please set up your profile by completing the steps below.",
         "noWalletConnected": "Please complete your profile! You still need to connect your wallet and add KYC info on-chain.",
```
portuu3, line 203: Also for the case of a no wallet connected, we should show "Please set up your profile by completing the steps below.". It is a general error message for any missing step.

portuu3 (changes requested):
73 not applied

Issue #2466, opened 2024-08-30
[Job Launcher Frontend] Include Qualifications Dropdown on Job Creation


flopez7:
Description
Add a dropdown menu to the job creation page. This dropdown will allow users to select from a list of available qualifications when creating a job.

Motivation
Providing a section for selecting qualifications simplifies the process for users to associate qualifications with a job. It ensures that job creators can easily assign the appropriate qualifications, leading to better job-user matching.

Implementation Details

- Add a dropdown menu to the job creation page under the section for qualifications.
- Fetch the available qualifications from the get qualifications endpoint.
- Display the fetched qualifications in a multiselect input, allowing users to select zero, one or more.
- Ensure that the selected qualifications are included in the job creation request sent to the server.
- Test the dropdown functionality to ensure it correctly retrieves and displays qualifications and integrates smoothly with job creation.

PR #2488, opened 2024-09-06
Add qualifications in Job Launcher Frontend

flopez7:
## Description

Include Qualifications in Job Creation

## Summary of changes

- Fetch the available qualifications from the get qualifications endpoint.
- Add a dropdown menu to the job creation page under the section for qualifications.
- Display the fetched qualifications in a multiselect input, allowing users to select zero, one or more.
- Add the selected qualifications to the job creation request sent to the server.

## Related issues
#2466


portuu3 (changes requested):
The endpoint is called twice
image

In the dropdown show the title instead of the reference

flopez7:
> The endpoint is called twice image
> 

The endpoint is being called twice due to React's StrictMode in development, which intentionally invokes certain lifecycle methods (like useEffect) twice to help identify side effects. This behavior only occurs in development and will not happen in production

Issue #2532, opened 2024-09-19
[Human APP][Server / Client] Implement a feature flag for "Jobs Discovery" mechanism


Dzeranov:
**Description**
Implement a feature flag for "Jobs Discovery" mechanism

**Motivation**
To be able to enable/disable this feature at any moment + being able to have the same codebase on staging and production while this feature is not fully tested and deployed.

**Implementation details** 
Frontend:
1. Add `FEATURE_FLAG_JOBS_DISCOVERY` env variable.
2. If this env is set to `false`, don't show Jobs Discovery tab and it's content (just in case someone will try to directly go to `/worker/jobs-discovery`

Backend:
1.  Add `FEATURE_FLAG_JOBS_DISCOVERY` env variable.
2. In case this variable is set to false, all mechanisms related to this feature should be disabled (cron-jobs, endpoints for calling exchnage oracles etc.)

PR #2533, opened 2024-09-19
[Human APP][Server / Client] Implement a feature flag for "Jobs Discovery" mechanism

flopez7:
## Description

Implement a feature flag for "Jobs Discovery" mechanism

## Summary of changes

Frontend:

Add FEATURE_FLAG_JOBS_DISCOVERY env variable.
If this env is set to false, don't show Jobs Discovery tab and it's content (just in case someone will try to directly go to /worker/jobs-discovery
Backend:

Add FEATURE_FLAG_JOBS_DISCOVERY env variable.
In case this variable is set to false, all mechanisms related to this feature should be disabled (cron-jobs, endpoints for calling exchnage oracles etc.)

## Related issues
#2532

### DIFF
--- a/packages/apps/human-app/server/src/modules/cron-job/cron-job.service.ts
+++ b/packages/apps/human-app/server/src/modules/cron-job/cron-job.service.ts
@@ -36,7 +36,7 @@ export class CronJobService {
   }
 
   initializeCronJob() {
-    const job = new CronJob('*/30 * * * * *', () => {
+    const job = new CronJob('*/3 * * * *', () => {
       this.updateJobsListCron();
     });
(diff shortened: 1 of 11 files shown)


portuu3 (dismissed):

packages/apps/human-app/server/src/modules/cron-job/cron-job.service.ts
portuu3, line 1: can we disable the cron expression somehow?

portuu3 (changes requested):

portuu3 (approved):

PR #2554, opened 2024-09-25
[CVAT][Exchange Oracle] Add EscrowValidation and optimise state trackers

Bobronium:
### `src.cron.cvat.state_trackers`
- **Optimized handling of `track_completed_tasks` and `track_completed_projects`:** Both now use a single query to update statuses efficiently.
- **New `track_escrow_validations` function:** This works similarly to `track_completed_escrows`, but specifically handles the validation phase.
- **Updates to `track_completed_escrows`:** Now creates or updates `EscrowValidation` records, which are used to manage the validation state of each escrow.

### `src.handlers.completed_escrows`
- **Unified handling of escrows with single and multiple projects:** The logic for handling escrows has been streamlined, removing unnecessary complexity between single and multiple project escrows.
- **Use of `EscrowValidation`:** The system now relies on `EscrowValidation` records to track and query escrows for validation, improving transparency and control.

### Other
- Alembic migration filename template has been changed to `{timestamp}_{name}_{revision}.py`, so they would be sorted in a chronological order. 

## New Logic for Handling Completed Escrows
Instead of selecting escrows for validation based on project type, we now first create `EscrowValidation` records corresponding to individual escrows.

- **Escrow selection:** `EscrowValidation` records are selected based on their number of validation attempts and status (`awaiting`). This ensures that escrows that have failed multiple validation attempts are deprioritized in favor of those with fewer attempts, preventing the queue from being filled with problematic escrows.
- **Status transitions:**
  - Once annotations are exported, the `EscrowValidation` status is set to `in_progress`, and the associated projects are marked as `validation`, as before.
  - Upon receiving a webhook from the recording oracle, the `EscrowValidation` status is updated to `completed` if validation is successful. If not, it eventually reverts to `awaiting` for further attempts.


packages/examples/cvat/exchange-oracle/alembic/versions/1696514176_init_16ecc586d685.py
zhiltsov-max, line 1: Why the existing migrations were renamed? I guess this will require extra actions to migrate the existing DB.

packages/examples/cvat/exchange-oracle/src/services/cvat.py
```diff
@@ -138,23 +156,170 @@ def get_projects_by_status(
...
+            },
+        )
+        .returning(EscrowValidation.id)
+    )
 
-    return escrows.limit(limit).all()
+    return session.execute(insert_stmt).fetchall()
+
+
+def get_escrows_for_validation(session: Session, limit=100) -> Sequence[tuple[str, str, int]]:
```
zhiltsov-max, line 283:
```suggestion
def get_escrows_for_validation(session: Session, *, limit: int = 100) -> Sequence[tuple[str, str, int]]:
```

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
```diff
@@ -30,382 +27,259 @@
...
+            limit=CronConfig.track_completed_escrows_chunk_size,
+        )
+
+    unhandled_validations = set(validation_id for validation_id, _, __ in escrow_validations)
+    handled_validations = set()
+    try:
+        for validation_id, escrow_address, chain_id in escrow_validations:
+            if _handle_escrow_validation(logger, escrow_address, chain_id):
+                handled_validations.add(validation_id)
+    finally:
```
zhiltsov-max, line 279: Shouldn't all of this be within a single transaction with select for update object locking?

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
```diff
@@ -30,382 +27,259 @@
     postprocess_annotations,
     prepare_annotation_metafile,
 )
+from src.models.cvat import Job, Project
 from src.services.cloud.types import BucketAccessInfo
 from src.utils.assignments import parse_manifest
-from src.utils.logging import NullLogger
 
 
-class _CompletedEscrowsHandler:
```
zhiltsov-max, line 38: What was the point of removing the class here? Now you need to pass some shared variables into every function (e.g. logger).

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
```diff
@@ -30,382 +27,259 @@
...
+    annotation_format = CVAT_EXPORT_FORMAT_MAPPING[manifest.annotation.type]
+    job_annotations = _download_job_annotations(logger, annotation_format, jobs)
+
+    if manifest.annotation.type == TaskTypes.image_skeletons_from_boxes.value:
+        # we'll have to merge annotations ourselves for skeletons
+        # might want to make this the only behaviour in the future
+        project_annotations_file = None
+        project_images = None
+    else:
+        # escrows with simple task types are ought to have only one project
```
zhiltsov-max, line 86:
```suggestion
        # escrows with simple task types must have only one project
```
I'd say it should be an assertion error instead. It's not like it's something ideal that never happens, which is supposed by "ought to".

packages/examples/cvat/exchange-oracle/src/crons/cvat/state_trackers.py
```diff
@@ -150,8 +127,17 @@ def track_assignments(logger: logging.Logger) -> None:
 
 
 @cron_job
-def track_completed_escrows(logger: logging.Logger) -> None:
-    handle_completed_escrows(logger)
+def track_completed_escrows(logger: logging.Logger, session: Session) -> None:
+    awaiting_validations = cvat_service.create_escrow_validations(session)
+    logger.info(
+        f"{len(awaiting_validations)} escrow validations "
+        f"are in awaiting state: {awaiting_validations}"
```
zhiltsov-max, line 134:
```suggestion
        f"Got {len(awaiting_validations)} new escrows"
        f"awaiting validation: {format_list(awaiting_validations)}"
```

Also, please change this (and other updated) loggings to full list in the debug mode and to `format_list()`-formatted for the `info` loglevel.

packages/examples/cvat/exchange-oracle/src/services/cvat.py
```diff
@@ -138,23 +156,170 @@ def get_projects_by_status(
...
+    result = session.execute(stmt)
+    updated_tasks = result.fetchall()
+    session.commit()
+
+    return [row.cvat_id for row in updated_tasks]
+
+
+def create_escrow_validations(session: Session):
+    # Escrow projects can be in either 'validation' or 'completed' status:
+    # - 'completed': Some jobs were rejected and reannotated.
```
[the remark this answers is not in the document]
Bobronium: I've rephrased the comment completely

packages/examples/cvat/exchange-oracle/src/services/cvat.py
```diff
@@ -138,23 +156,170 @@ def get_projects_by_status(
...
+    # - 'validation': Other jobs were successfully validated and remain in this state.
+    #
+    # Ensure all projects are in one of these two states
+    # and that at least one project is 'completed'.
+    # This prevents multiple threads from processing the same escrow concurrently
+    # before validation finishes.
```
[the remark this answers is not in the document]
Bobronium: AFAIK, whatever worker will be able to acquire the lock first will perform the insert (or update), and then other worker will do the same. I've also tested running this query concurrently in two workers and it worked as expected.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
Bobronium: I don't think having long connections to the database is a good idea, let alone holding a lock.

Besides, we don't really need it here — `get_escrows_for_validation` updates the status of validations to `in-progress`. Though it was not obvious from its name, so I've changed the function name to `prepare_escrows_for_validation`.

packages/examples/cvat/exchange-oracle/src/services/cvat.py
```diff
@@ -138,23 +156,170 @@ def get_projects_by_status(
...
+        .values(status=new_status.value)
+        .returning(Project.cvat_id)
+    )
+
+    if included_types is not None:
+        stmt = stmt.where(Project.job_type.in_([t.value for t in included_types]))
+
+    result = session.execute(stmt)
+    updated_projects = result.fetchall()
+    session.commit()
```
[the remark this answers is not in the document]
Bobronium: Good point. I moved `commit()` invocations outside of the service.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
```diff
@@ -30,382 +27,259 @@
...
         with SessionLocal.begin() as session:
-            escrows_with_completed_projects = cvat_service.get_escrows_by_project_status(
+            cvat_service.update_escrow_validation_statuses_by_ids(
                 session,
-                ProjectStatuses.completed,
-                included_types=[TaskTypes.image_skeletons_from_boxes],
-                limit=CronConfig.track_completed_escrows_chunk_size,
+                unhandled_validations - handled_validations,
+                EscrowValidationStatuses.awaiting,
             )
```
zhiltsov-max, line 285: I don't think this request optimization makes sense here. Probably, we need to do something similar to webhook processing here, so 1 by 1. Escrow export is a lengthy operation that takes about 5 minutes, so we're not going to launch many of them in one status update (cron) call.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
zhiltsov-max: So if I understand correctly, if we get some escrows for handling and then the function fails for some reason (e.g. the exchange oracle is shut down), these escrows will never be processed? Probably, we need something different. 

> I don't think having long connections to the database is a good idea, let alone holding a lock.

Do you know about any problems with doing so?

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
Bobronium: Makes sense, thank you. I've changed it to per-validation update and decreased the default limit.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
Bobronium: As a side note.

I think we would benefit here from adopting queues and workers approach and handle. This would abstract away retries logic and will allow us to focus just on task implementation at hand. 

State trackers then won't be doing any work and will just schedule jobs.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
Bobronium: As a side note.

I think we would benefit here from adopting jobs, queues and workers approach. This would abstract away retries logic and will allow us to focus just on task implementation at hand.

State trackers then won't be doing any work and will just schedule jobs.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
zhiltsov-max: The problem with a bunch of functions instead of a class is that at some point these functions become [this](https://github.com/cvat-ai/cvat/blob/12858584c8b33acb021becd239201de8440ed669/cvat/apps/engine/task.py#L556-L1516) - a single big function with lots of branching and context. And it can't be easily refactored or extended for another new use case.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
zhiltsov-max: Overall looks good.

> I think we would benefit here from adopting jobs, queues and workers approach.

Do you have some specific solution in mind? Anyway, it doesn't seem to be a problem now.

zhiltsov-max (approved):

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
Bobronium: I’m afraid the example you provided doesn’t really connect with the class vs. functions discussion, nor with this particular case.

It shows instead how function (or method) can grow in complexity when lacking proper decomposition. Which is a completely valid, but separate issue.

Your argument instead implies that code written in functional languages must all suffer from having large, complex functions, which is simply not the case.

Another contradiction to your example is the fact that current implementation is actually less complex and more concise than previous class-based one.

So it's clearly not a question of whether to use class to reduce complexity, since the complexity itself arises due to different reasons.

Issue #2524, opened 2024-09-18
[Subgraph] Add escrow to transactions


portuu3:
**Description**
For the cases where the transaction method is `bulkTransfer`, add the escrow address to the transaction entity

**Motivation**
Allow to link transactions with escrows

**Implementation details** 
- Modify subgraph.
- Update SDK,
- Update `/details/transactions` endpoint in dashboard backend.

PR #2572, opened 2024-10-01
[Subgraph][SDK] Refactor transactions

flopez7:
## Description

Refactor transactions to improve subgraph storage

## Summary of changes

Group all transactions with same txhash only in one 
Add token address to transactions
Add withdraw event to Escrow contract and subgraph
Update python and typescript SDK

## How test the changes

`yarn test`
`make run-test`

## Related issues
#2524


packages/core/contracts/Escrow.sol
```diff
@@ -212,6 +213,8 @@ contract Escrow is IEscrow, ReentrancyGuard {
         }
 
         _safeTransfer(_token, canceler, _amount);
+
+        emit Withdraw(_token, _amount);
```
portuu3, line 217: I think the tests should also be modified to check that event has been triggered

portuu3 (changes requested):

portuu3 (approved):

PR #2617, opened 2024-10-08
[CVAT][Exchange Oracle] Changed JWT algo from HS256 to ES256

Dzeranov:
## Description

Changed JWT decoding algorithm from `HS256` to `ES256`.

### DIFF
--- a/packages/examples/cvat/exchange-oracle/src/core/config.py
+++ b/packages/examples/cvat/exchange-oracle/src/core/config.py
@@ -244,6 +244,8 @@ class CoreConfig:
 
 
 class HumanAppConfig:
+    # jwt_public_key is obtained from the Human App.
+    # To generate a key pair for testing purposes:
     # openssl ecparam -name prime256v1 -genkey -noout -out ec_private.pem
     # openssl ec -in ec_private.pem -pubout -out ec_public.pem
     # HUMAN_APP_JWT_KEY=$(cat ec_public.pem)
--- a/packages/examples/cvat/exchange-oracle/src/core/config.py
+++ b/packages/examples/cvat/exchange-oracle/src/core/config.py
@@ -244,7 +244,10 @@ class CoreConfig:
 
 
 class HumanAppConfig:
-    jwt_key = os.environ.get("HUMAN_APP_JWT_KEY", "sample")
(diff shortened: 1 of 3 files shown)


Bobronium (approved):

packages/examples/cvat/exchange-oracle/src/core/config.py
```diff
@@ -244,7 +244,10 @@ class CoreConfig:
 
 
 class HumanAppConfig:
-    jwt_key = os.environ.get("HUMAN_APP_JWT_KEY", "sample")
+    # openssl ecparam -name prime256v1 -genkey -noout -out ec_private.pem
+    # openssl ec -in ec_private.pem -pubout -out ec_public.pem
+    # HUMAN_APP_JWT_KEY=$(cat ec_public.pem)
```
Dzeranov, line 249: Not sure if you need this comment here since JWT is being generated on Human App.

Bobronium: We certainly don't *need* it. 

Yet it might be helpful in case somebody will have to mock the Human App calls during development.

Do you see any harm having this comment or just haven't seen any reason to have it?

Dzeranov: IMO, this comment might be a bit misleading.
For instance, if I didn't know about this schema, I would think that this key should be generated on ExO side.

Bobronium: Hmm. The snippet clearly shows that only public key is known by ExO, while private key is discarded, but I see your point. 

The main reason I included it is that the config doesn’t make it clear what the key should look like (the previous default "sample" won't work anymore). Rather than describing what key is used, I chose to show how to generate it, since it makes both parts obvious.

I think, we should make clear distinction between generating the key pair for testing purposes, and obtaining Human App public key. Even making it default maybe?

Dzeranov: I think that we can just add a line to your comment saying that this is an example of how to generate a key for testing purposes.

Bobronium: Done!

Bobronium (approved):

PR #2620, opened 2024-10-09
[CVAT][Exchange Oracle] Use Alembic in tests

Bobronium:
## Summary of changes
Previously, to initialise test database, we've dropped all metadata then created it, based on current models. This lacked any awareness of our migrations and schema that was generated by it. Nothing really was testing how schema was applied from migrations, so error in alembic revision file wouldn't affect execution of tests and would be caught only on deployment attempt.

With these changes full `alembic upgrade head` and `alembic downgrade base` is ran before each test. 

Since running plain `alembic upgrade/downgrade` commands is too slow due to isolated revision executions, before each test session, full SQL of `alembic upgrade head` and `alembic downgrade head:base` is generated, and then reused for each test case which speeds up things significantly.

After tests are executed, the database remains in a clean state with all migrations applied.

Existing migrations had to be modified a bit to ensure they could be rendered as raw SQL in offline mode.

## How test the changes
Self-tested in CI.

## Related issues
- #2615


packages/examples/cvat/exchange-oracle/alembic/versions/1707173682_non_unique_escrows_c1e74c227cfe.py
```diff
@@ -22,6 +35,22 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
+    offline_mode = op.get_context().environment_context.is_offline_mode()
+    assert offline_mode or "TESTING" in os.environ or "test" in op.get_bind().engine.url, (
```
zhiltsov-max, line 39: This is a configuration error, so it's better to throw a regular exception.

packages/examples/cvat/exchange-oracle/alembic/versions/1707173682_non_unique_escrows_c1e74c227cfe.py
```diff
@@ -22,6 +35,22 @@ def upgrade() -> None:
...
+        "This downgrade deletes data and should only run in a test environment."
+        "If you are sure you want to run it, set the TESTING environment variable."
+    )
```
Bobronium, line 42:
```suggestion
    if not (offline_mode or "TESTING" in os.environ or "test" in op.get_bind().engine.url):
        raise RuntimeError(
            "This downgrade deletes data and should only run in a test environment."
            "If you are sure you want to run it, set the TESTING environment variable."
        )
```

zhiltsov-max (approved):

PR #2644, opened 2024-10-15
[CVAT] Fix oracles lint workflow

Bobronium:
## Description
Installing poetry without virtualenv was always discouraged, but now it's prohibited by pip without explicit `--break-system-packages` flag. Instead of supplying it, I updated installation to satisfy pip and poetry recommendations. 

## Related issues
- https://github.com/humanprotocol/human-protocol/pull/2639


Bobronium:
Reverting changes to oracles Dockerfiles for now. I was hoping it would be a harmless update, but faced issues that I don't have time to debug right now.

Dzeranov (approved):

PR #2650, opened 2024-10-17
Feat/leaderboard datagrid

KacperKoza343:
task HDM-20


dnechay:
Do we need this PR? It seems to be part of https://github.com/humanprotocol/human-protocol/pull/2661

packages/apps/dashboard/ui-2024/src/features/Leaderboard/components/DataGrid.tsx
```diff
@@ -0,0 +1,148 @@
...
+								display="flex"
+								justifyContent="center"
+								alignItems="center"
+							>
+								<Typography>No data</Typography>
+							</Box>
+						);
+					},
+					loadingOverlay() {
+						return <Loader height="30vh" />;
```
dnechay, line 75: Why `vh` here?

dnechay (commented):
"View All" is not available anymore, is this expected?

mgacek-blockydevs:
> "View All" is not available anymore, is this expected?
> 
> Also font size is changed from 16 to 14, not sure if it's an intended change

image
should be ok :)

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+		return formattedData;
+	}, [formattedData, chainId]);
+
+	const columns: GridColDef<(typeof visibleRows)[number]>[] = [
+		...(isMobile
+			? []
+			: [
+					{
+						field: 'id',
```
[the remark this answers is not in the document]
mgacek-blockydevs: fixed

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+	const visibleRows = useMemo(() => {
+		if (chainId !== -1) {
+			return formattedData.filter((elem) => elem.chainId === chainId);
+		}
+
+		return formattedData;
+	}, [formattedData, chainId]);
+
+	const columns: GridColDef<(typeof visibleRows)[number]>[] = [
```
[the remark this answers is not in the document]
mgacek-blockydevs: I'm ok with moving it into useMemo hook, good catch, I memoized visibleRows and forgot about it. I'm not sure about to move somewhere else, because this is hook with logic for this datagrid so I think filters, sorting functions and columns definitions should be together. What do you think?

packages/apps/dashboard/ui-2024/src/features/Leaderboard/index.tsx
```diff
@@ -5,7 +5,7 @@ import { LeaderBoardData } from '@services/api/use-leaderboard-details';
 import { useNavigate } from 'react-router-dom';
 import SimpleBar from 'simplebar-react';
 import { SelectNetwork } from './components/SelectNetwork';
-import { Table } from './components/Table';
+import { DataGrid } from './components/DataGrid';
```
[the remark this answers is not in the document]
mgacek-blockydevs: fixed

packages/apps/dashboard/ui-2024/src/features/Leaderboard/components/DataGrid.tsx
```diff
@@ -0,0 +1,148 @@
...
+	const {
+		mobile: { isMobile },
+	} = useBreakPoints();
+
+	const tableIsEmpty = status === 'success' && visibleRows.length === 0;
+	const tableMinHeight = status === 'success' && !tableIsEmpty ? 'unset' : 300;
+
+	return (
+		<Box width="100%" height={tableMinHeight} sx={{ overflow: 'hidden' }}>
+			<MuiDataGrid
```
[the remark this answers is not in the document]
mgacek-blockydevs: fixed

packages/apps/dashboard/ui-2024/src/features/Leaderboard/components/DataGrid.tsx
```diff
@@ -0,0 +1,148 @@
...
+				hideFooter
+				disableVirtualization={isMobile}
+				apiRef={apiRef}
+				loading={status === 'pending'}
+				rows={visibleRows}
+				columns={columns}
+				autosizeOptions={{
+					expand: true,
+				}}
+				onSortModelChange={() => {
```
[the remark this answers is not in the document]
mgacek-blockydevs: yes, I left comment why we need do it

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
```diff
@@ -0,0 +1,211 @@
...
+							headerClassName: 'home-page-table-header',
+							sortable: false,
+							renderCell: (params: GridRenderCellParams) => (
+								<Typography
+									variant="body1"
+									height="100%"
+									display="flex"
+									alignItems="center"
+								>
+									{params.api.state.sorting.sortedRows.indexOf(params.id) + 1}
```
dnechay, line 61: This logic is implicit and easy to break, esp. with the fact we have a hack with `onSortModelChange` for it. The right way of doing this would be to implement something similar ["server-side" sorting](https://mui.com/x/react-data-grid/sorting/#server-side-sorting), where we sort on our end and change raw index accordingly.

IMU the idea of this leaderboard is to show items depending on "amountStaked" by default and "index" column should show the "rank" based on this criteria. In case user decides to sort the table based on some other column - rank won't change.

So I see three possible ways of making it right:
1) Implement it like "server-side" to avoid hacks with re-render triggering
2) Remove this column at all
3) Display it as is, assuming it is a `rank` value (so it will work [like this table](https://www.coingecko.com/en/chains))

IMO #1 is overhead here, so I would go #3 since it corresponds to the idea of leaderboard

dnechay (commented):
1. Sorting arrows are cut
Screenshot 2024-11-11 at 12 39 43

2. Please fix the git history, maybe by re-applying changes on top of current HEAD in develop. Right now we see excessive commits (like [this one](https://github.com/humanprotocol/human-protocol/pull/2650/commits/579e03c9a402f7ad1086ab22995df95d2d52d068)) when we should not
3. Some other minor comments

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
dnechay: Discussed with the team, let's do the following now:
- keep sorting only for `stake`, `reputation` and `fee` columns
- always keep 1-2-3 raw index, no matter what is the sort criteria; either go with a suggested option #1 or find some other ways to achieve that w/o implicit trigger of re-rendering

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDatagridOptions.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+      return colsWithOptions;
+    },
+    [handlePinnedColOption, handleRowOrderNum, isRowIdx, pinnedColumnName]
+  );
+
+  const handleSortModelChange = useCallback(
+    (_model: GridSortModel, details: GridCallbackDetails) => {
+      details.api.getSortedRows().forEach((row, idx) => {
+        apiRef.current.updateRows([{ ...row, id: idx }]);
```
dnechay, line 95: That's definitely better and cleaner than implicit re-render! Love it!

I assume that calling `updateRows` for each row will trigger re-render multiple times, so maybe we
```
const sortedRows = details.api.getSortedRows();
apiRef.current.updateRows(sortedRows.map((row, idx) => ({ ...row, id: idx })))
```
wdyt?

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
```diff
@@ -0,0 +1,190 @@
...
+  } = useLeaderboardSearch();
+
+  const {
+    mobile: { isMobile },
+  } = useBreakPoints();
+  const formattedData = useMemo(() => {
+    return data.map((row, index) => {
+      return {
+        ...row,
+        id: index,
```
dnechay, line 30: AFAIK data-grid relies on `id` for different internal things, so I would not use it for row index to avoid implicit/unexpected issues.

Would be better to use `address` as `id` and have `rowIndex: index`

dnechay (commented):
Looks better now!
Some minor comments before we move on

yarn.lock
```diff
@@ -1972,6 +1972,11 @@
     "@ethersproject/properties" "^5.7.0"
     "@ethersproject/strings" "^5.7.0"
```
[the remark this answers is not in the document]
adrian-oleskiewicz: Nice catch! I probably messed something up during the rebase, thanks :)

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
```diff
@@ -0,0 +1,190 @@
...
+          <Typography component="span" variant="body3">
+            Role
+          </Typography>
+        ),
+        renderCell: (params: GridRenderCellParams) => (
+          <RoleCell role={params.value} />
+        ),
+      },
+      {
+        field: 'address',
```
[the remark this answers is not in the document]
adrian-oleskiewicz: done

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
```diff
@@ -0,0 +1,190 @@
...
+        valueFormatter: (value: string) => {
+          return `${value} HMT`;
+        },
+        renderCell: (params: GridRenderCellParams) => (
+          <TextCell value={params.formattedValue} />
+        ),
+      },
+      {
+        field: 'chainId',
+        headerName: 'Network',
```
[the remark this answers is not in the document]
adrian-oleskiewicz: done

yarn.lock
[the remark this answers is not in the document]
dnechay: It's not because of `rebase`, the necessity of these steps is due to "yarn monorepo" setup 🤷🏻

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDatagridOptions.tsx
adrian-oleskiewicz: Yes that would be definitely better approach but unfortunately we are limited here because of a plan.
Zrzut ekranu 2024-11-21 o 09 37 33

But still I improved it a bit thanks to your comment, there is no need to update id

adrian-oleskiewicz: https://mui.com/x/react-data-grid/row-updates/#the-updaterows-method

dnechay: What a tricky surprise from them 😒

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
adrian-oleskiewicz: good point, thanks! I just realized that  we don't need an index for a row because we are checking indexOf a row by id so I changed this id from index to address.
Zrzut ekranu 2024-11-21 o 09 54 15

adrian-oleskiewicz: So here we can update just a row
Zrzut ekranu 2024-11-21 o 09 58 51

dnechay: > I just realized that we don't need an index for a row because we are checking indexOf a row

And that was exactly what we were trying to avoid :)

This `sortedRows.indexOf` won't return the proper value until the whole grid is re-rendered (and you achieve it implicitly by calling `updateRow([row])`. The latter in `handleSortModelChange` looks strange: why do we have to update the row with same data on sort model change if `getSortedRows` already has it?! That thing with row index is still "implicit magic" and hard to catch for someone how is not deep into data-grid behavior. Instead - code should be explicit so it's easy to maintain.

This is why I'm suggesting to have `rowIndex` property for data, update it in `handleSortModelChange` and use it as separate column

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
```diff
@@ -0,0 +1,194 @@
...
+          );
+        },
+        renderCell: (params: GridRenderCellParams) => (
+          <ChainCell chainId={params.value} />
+        ),
+      },
+      {
+        field: 'reputation',
+        headerName: 'Reputation Score',
+        sortable: false,
```
dnechay, line 140: Reputation Score should remain sortable

packages/apps/dashboard/ui-2024/src/features/Leaderboard/components/DataGridWrapper.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const {
+    mobile: { isMobile },
+  } = useBreakPoints();
+
+  const tableIsEmpty = status === 'success' && visibleRows.length === 0;
+  const tableMinHeight = status === 'success' && !tableIsEmpty ? 'unset' : 300;
+
+  return (
+    <Box width="100%" height={tableMinHeight} sx={{ overflow: 'hidden' }}>
+      <DataGrid
```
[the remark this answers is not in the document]
adrian-oleskiewicz: Should be ok :)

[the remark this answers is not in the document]
dnechay: Yup, already checked via vercel ✅

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
adrian-oleskiewicz: I disabled sorting because this feature is "coming soon"

Zrzut ekranu 2024-11-21 o 10 48 06

dnechay: That makes sense, we can enable it when we start displaying some values 👍🏻

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDataGrid.tsx
adrian-oleskiewicz: I investigated it a bit and this update from `handleSortModelChange` does not cause the component to be rerendered. So I can't make it working with rowIndex field. I fully agree this solution is more readable and easier to maintain but at the moment I have no idea how to implement it differently.

I can get correct order by refering this state `params.api.state.sorting.sortedRows`

I have 2 console.logs:
1. refers to `params.api.state.sorting.sortedRows`
2. refers to updated `rowIndex` which remains same after sorting what you can check in second screenshot

Zrzut ekranu 2024-11-21 o 11 25 17

Zrzut ekranu 2024-11-21 o 11 23 19

It turns out that this `apiRef.current.updateRows([{ ...row, rowIndex: idx }])` will not rerender component

dnechay: @adrian-oleskiewicz 

Tricky part is that `details.api.getSortedRows();` doesn't give you rows sorted by new sort model, so the order of operations is next:
1. you click on column header
2. sort model is updated in data grid
3. `handleSortModelChange` callback invoked
4. data sorted in data grid

You can easily verify this by adding logs to `handleSortModelChange` right 
```
const sortedRows = details.api.getSortedRows();
console.log(
  'sorted',
  sortedRows.map((r, index) => ({
    id: r.id,
    old: r.rowIndex,
    new: index,
  }))
);
setTimeout(() => {
  const sortedRows = details.api.getSortedRows();
  console.log(
    'sorted 2',
    sortedRows.map((r, index) => ({
      id: r.id,
      old: r.rowIndex,
      new: index,
    }))
  );
}, 1000);
```
I don't know what is their internal implementation (probably they rely on references and memo some stuff), but when the data is sorted - the do not call `renderCell`, even though the DOM is updated. In case you call `updateRows` - it causes `renderCell` call.

So, taking the above details into account, we can actually make it working using `rowIndex` because luckily there is ["applySorting"](https://mui.com/x/api/data-grid/grid-api/#grid-api-prop-applySorting) util in its API, i.e. `handleSortModelChange` should be
```
apiRef.current.applySorting();
// now rows are actually sorted
const sortedRows = details.api.getSortedRows();
sortedRows.forEach((row, index) => {
  apiRef.current.updateRows([{ ...row, rowIndex: index }]);
});
```

Seems like together we finally got how to work with it properly 😅

adrian-oleskiewicz: @dnechay Great job! You found the missing puzzle. Thanks for such a good code review!

packages/apps/dashboard/ui-2024/src/features/Leaderboard/hooks/useDatagridOptions.tsx
```diff
@@ -0,0 +1,132 @@
...
+    [isMobile, pinnedColumnName]
+  );
+
+  const handleRowOrderNum = useCallback(
+    (cols: GridColDef<T>[]) => {
+      return isMobile
+        ? cols
+        : [
+            {
+              field: 'id',
```
dnechay, line 55:
```suggestion
              field: 'rowIndex',
```

and `{params.value + 1}` below

adrian-oleskiewicz: Good catch again, sorry for oversight :) fixed

dnechay (approved):

Issue #2364, opened 2024-08-09
[Smart contracts] Remove allocation


portuu3:
**Description**
Remove allocation from staking contract.
Unstake process should be a 2 steps process:
1. Call unstake function: amount will be locked for the locking period.
2. Call withdraw function: can be called after the lock period.

**Motivation**
Avoid too many transactions to allocate to an escrow.

**Implementation details** 
Remove allocation
Add a lock period duration variable.
Add a function to update the lock period duration variable.

PR #2651, opened 2024-10-17
[CORE][SDK] Remove allocation from Staking contract

flopez7:
## Description

Remove allocation from Staking contract and update SDK

## Summary of changes

Removed allocation and all it's related changes from staking.
Remove RewardPool contract
Updated Python and Typescript SDK

## Related issues
#2364


packages/apps/reputation-oracle/server/src/modules/payout/payout.service.ts
```diff
@@ -69,6 +69,7 @@ export class PayoutService {
       results.amounts,
       results.url,
       results.hash,
+      true,
```
portuu3, line 72: I think this should not be hardcoded to true.

portuu3 (approved):

PR #2652, opened 2024-10-17
Hum 35 check paddings in navbar

KacperKoza343:
task HUM-35


dnechay (approved):
Changes in this PR lgtm in general

Side question noticed from this one.

`Main` container defined with `rem`, but logo and many other things with `px`.
In case we change base `font-size` - many paddings get crazy. Does it make sense to align paddings/margins to use `px` instead?

dnechay:
@KacperKoza343 please rebase this branch before merging because commits history looks a bit wild

KacperKoza343:
> @KacperKoza343 please rebase this branch before merging because commits history looks a bit wild

done

PR #2653, opened 2024-10-17
Hum 38 invalid status chip color on mobile

KacperKoza343:
task HUM-38

### DIFF
--- a/packages/apps/human-app/frontend/src/pages/worker/jobs/components/my-jobs/mobile/my-jobs-table-mobile.tsx
+++ b/packages/apps/human-app/frontend/src/pages/worker/jobs/components/my-jobs/mobile/my-jobs-table-mobile.tsx
@@ -24,6 +24,7 @@ import { useColorMode } from '@/hooks/use-color-mode';
 import { Chip } from '@/components/ui/chip';
 import type { JobType } from '@/smart-contracts/EthKVStore/config';
 import { EscrowAddressSearchForm } from '@/pages/worker/jobs/components/escrow-address-search-form';
+import { colorPalette as lightModeColorPalette } from '@/styles/color-palette';
 
 interface MyJobsTableMobileProps {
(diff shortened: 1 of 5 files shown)


dnechay (commented):
Good catch with "reject" button! I though it was intentionally disabled, but now it looks better

Please resolve merge conflict

One question: is there a plan to support dynamic colors per status (there are some in figma) or only main one?

KacperKoza343:
> Good catch with "reject" button! I though it was intentionally disabled, but now it looks better
> 
> Please resolve merge conflict
> 
> One question: is there a plan to support dynamic colors per status (there are some in figma) or only main one?

I need to find a message where we discussed status of these colors, because I believe that one color for all states was a temporary solution that we choose. If we want to handle all this colors I need to have a union of available colors in response

packages/apps/human-app/frontend/src/pages/worker/jobs/components/my-jobs/mobile/my-jobs-table-mobile.tsx
```diff
@@ -162,15 +172,27 @@ export function MyJobsTableMobile({
...
+                        <TableButton
+                          component={Link}
+                          disabled={buttonDisabled}
+                          fullWidth
+                          target="_blank"
+                          to={d.url}
+                        >
+                          {t('worker.jobs.solve')}
+                        </TableButton>
+                        <RejectButton
```
dnechay, line 185: @KacperKoza343 probably something went wrong while resolving conflicts and now we have 2 "reject" buttons

dnechay (changes requested):

packages/apps/human-app/frontend/src/pages/worker/jobs/components/my-jobs/mobile/my-jobs-table-mobile.tsx
KacperKoza343: done

dnechay (approved):

PR #2654, opened 2024-10-17
Hum 42 job description and job type chip are out of job card

KacperKoza343:
task HUM-42

### DIFF
--- a/packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/mobile/available-jobs-table-mobile.tsx
+++ b/packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/mobile/available-jobs-table-mobile.tsx
@@ -119,14 +119,7 @@ export function AvailableJobsTableMobile({
               <Grid columnSpacing="2rem" container>
                 <Grid item xs={12}>
                   <ListItem label={t('worker.jobs.jobDescription')}>
-                    <Typography
-                      sx={{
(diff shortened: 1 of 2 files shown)


packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/mobile/available-jobs-table-mobile.tsx
```diff
@@ -119,14 +119,7 @@ export function AvailableJobsTableMobile({
...
                   <ListItem label={t('worker.jobs.jobDescription')}>
-                    <Typography
-                      sx={{
-                        textOverflow: 'ellipsis',
-                        overflow: 'hidden',
-                        whiteSpace: 'nowrap',
-                      }}
-                      variant="subtitle1"
-                    >
+                    <Typography variant="subtitle1">
```
dnechay, line 122: When for `job type` it should be fine to break words since it can't be very big, I suppose it's not the case for job description.
Would it be better to still limit it to some length/size and show the full version on hover?

KacperKoza343: @dnechay  At this moment we fix issues so we will postpone idea with hover. HUM-42 was duplicated by HUM-64. That's related pr: https://github.com/humanprotocol/human-protocol/pull/2689

PR #2659, opened 2024-10-21
User registration step in annotation tools

portuu3:
## Description
If the annotation tool requires a user registration before annotate data, users should see an extra step in HUMAN App that drives them towards registration process.

## Summary of changes
- Add registration step to HUMAN App frontend.
- Add registartionNeeded and registrationInstructions to leader entity in subgraph.
- Update SDK based on subgraph changes.
- Post a request from HUMAN App to reputation oracle to save the registration completion.

## How test the changes

1. Set registration needed for some exchange oracle
2. Go to see the jobs of oracle. You should be required to go through the registration step.


packages/apps/human-app/frontend/src/api/services/worker/oracles.ts
```diff
@@ -24,6 +26,7 @@ const H_CAPTCHA_ORACLE: OracleSuccessResponse = {
   jobTypes: env.VITE_H_CAPTCHA_ORACLE_TASK_TYPES,
   role: env.VITE_H_CAPTCHA_ORACLE_ROLE,
   url: env.VITE_H_CAPTCHA_ORACLE_ANNOTATION_TOOL,
+  registrationNeeded: false,
```
Dzeranov, line 29: Technically, registration is needed for hcaptcha labeling but I think we can skip that for now.

packages/sdk/typescript/subgraph/src/mapping/KVStore.ts
```diff
@@ -108,6 +108,10 @@ export function handleDataSaved(event: DataSaved): void {
       operator.reputationNetwork = null;
     }
     operator.save();
+  } else if (key == 'registration_needed') {
+    leader.registrationNeeded = event.params.value == 'true';
```
Dzeranov, line 112: Maybe we need to bring `value` to lowercase before comparing it with 'true'?

packages/apps/human-app/frontend/src/api/services/worker/oracles.ts
portuu3: the process is just different that's why I set false.

Dzeranov (approved):

PR #2661, opened 2024-10-21
Hum 40 address validation and debounce for Search escrow address

KacperKoza343:
tasks HUM-40

### DIFF
--- a/packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
+++ b/packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
@@ -50,9 +50,8 @@ const getMyJobsTableData = async (dto: GetMyJobTableDataDto) => {
 
 export function useGetMyJobsData() {
   const { filterParams } = useMyJobsFilterStore();
-  const { address: oracle_address } = useParams<{ address: string }>();
-  const dto = { ...filterParams, oracle_address: oracle_address ?? '' };
-
+  const { address } = useParams<{ address: string }>();
+  const dto = { ...filterParams, oracle_address: address ?? '' };
   return useQuery({
     queryKey: ['myJobs', dto],
     queryFn: () => getMyJobsTableData(dto),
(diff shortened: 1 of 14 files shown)


packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
```diff
@@ -48,20 +50,49 @@ const getMyJobsTableData = async (dto: GetMyJobTableDataDto) => {
...
+
 export function useGetMyJobsData() {
-  const { filterParams } = useMyJobsFilterStore();
-  const { address } = useParams<{ address: string }>();
-  const dto = { ...filterParams, oracle_address: address ?? '' };
+  const {
+    filterParams: { escrow_address, ...filterParams },
+  } = useMyJobsFilterStore();
+  const { address: oracle_address } = useParams<{ address: string }>();
+  const [debouncedEscrowAddress] = useDebounce(
```
dnechay, line 60: Doing debounce in query is implicit and violates its responsibility: you might want to use that w/o debouncing, but having it here will cause side effects. So it shouldn't be done here, it should be done in place where we handle the user input.

No need to have "form" component" it can be just regular styled input where you pass `onChange={handleChange}`, and this `handleChange` is something like

```
const handleChange = debounce((event) => {
  queryData(event.target.value);
}, ms);
```

dnechay (changes requested):

packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
KacperKoza343: I moved debounce to `EscrowAddressSearchForm`. I would preferer to have this input wrapped in form, that gives me convenient validation and error states handling like in all other inputs in app. It that's important for you I can remove that completely, let me know.

KacperKoza343: I also removed changes that were a part of HUM-50

packages/apps/human-app/frontend/src/pages/worker/jobs/components/escrow-address-search-form.tsx
```diff
@@ -0,0 +1,89 @@
...
+    defaultValues: {
+      searchValue: '',
+    },
+    resolver: zodResolver(
+      z.object({ searchValue: addressSchemaOrEmptyString })
+    ),
+  });
+
+  const debouncedUpdater = debounce((value: string) => {
+    if (updater) updater(value);
```
dnechay, line 40: Maybe instead of doing this check we make `updated` required parameter for this, wdyt?

dnechay (commented):
lgtm

One non-blocking question

packages/apps/human-app/frontend/src/pages/worker/jobs/components/escrow-address-search-form.tsx
KacperKoza343: done

dnechay (approved):

PR #2667, opened 2024-10-22
Hum 50

KacperKoza343:
task HUM-50

### DIFF
--- a/packages/apps/human-app/frontend/src/api/fetcher.ts
+++ b/packages/apps/human-app/frontend/src/api/fetcher.ts
@@ -73,17 +73,20 @@ export function createFetcher(defaultFetcherConfig?: {
 }) {
   async function fetcher<SuccessInput, SuccessOutput>(
     url: string | URL,
-    fetcherOptions: FetcherOptionsWithValidation<SuccessInput, SuccessOutput>
+    fetcherOptions: FetcherOptionsWithValidation<SuccessInput, SuccessOutput>,
+    abortSignal?: AbortSignal
   ): Promise<SuccessOutput>;
 
   async function fetcher(
     url: FetcherUrl,
-    fetcherOptions: FetcherOptionsWithoutValidation
+    fetcherOptions: FetcherOptionsWithoutValidation,
+    abortSignal?: AbortSignal
   ): Promise<unknown>;
 
   async function fetcher<SuccessInput, SuccessOutput>(
     url: FetcherUrl,
-    fetcherOptions: FetcherOptions<SuccessInput, SuccessOutput>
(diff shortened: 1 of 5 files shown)


packages/apps/human-app/frontend/src/api/fetcher.ts
```diff
@@ -99,10 +102,18 @@ export function createFetcher(defaultFetcherConfig?: {
       : fetcherOptions.options;
     if (fetcherOptions.authenticated) {
       fetcherOptionsWithDefaults = appendHeader(fetcherOptionsWithDefaults, {
-        Authorization: `Bearer ${browserAuthProvider.getAccessToken() ?? ''}`,
+        // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
dnechay, line 105: Wouldn't it be better if we change the rule globally in order to allow not only strings, but also booleans, numbers and null/undefined?

dnechay (commented):
Approach lgtm, 2 points

1) We should apply the same logic to oracles table queries (now you can spawn multiple requests by selecting in job type filter)
2) Make sure your git history is clean before we merge
3) Do we need this PR? It seems to be part of https://github.com/humanprotocol/human-protocol/pull/2661

KacperKoza343:
> Approach lgtm, 2 points
> 
> 1. We should apply the same logic to oracles table queries (now you can spawn multiple requests by selecting in job type filter)
> 2. Make sure your git history is clean before we merge
> 3. Do we need this PR? It seems to be part of [Hum 40 hum 50 #2661](https://github.com/humanprotocol/human-protocol/pull/2661)

1. I will add it there
2. Done
3. Yes, that's right, but HUM-40 also have debounce so I added this one separately to have a better ability to test it, with debounce it's hard to see cancelled requests

dnechay:
@KacperKoza343 
> so I added this one separately to have a better ability to test it

That's true, but you always can comment some code in your working branch or alter git history for testing purpose.

If we keep two separate PRs and merge this one first - it will create unnecessary difficulties for you in the second one with resolving conflicts.

Up to you in the end

packages/apps/human-app/frontend/src/api/fetcher.ts
dnechay: @KacperKoza343 could you please do it in this PR?

dnechay (commented):
Looks good, one comment about linter comment

packages/apps/human-app/frontend/src/api/fetcher.ts
KacperKoza343: done

dnechay (approved):

PR #2668, opened 2024-10-22
Hum 44 overlap with some layout element blocks UI interaction

KacperKoza343:
task HUM-44


dnechay (commented):
Fix lgtm

Before approval, please align the git history so we don't have these extra commits displayed and can keep the history clean by adding your commit with changes just on top of HEAD

KacperKoza343:
> Fix lgtm
> 
> Before approval, please align the git history so we don't have these extra commits displayed and can keep the history clean by adding your commit with changes just on top of HEAD

Done

dnechay (approved):

PR #2677, opened 2024-10-22
Hum 47 bad validation when setting up operator

KacperKoza343:
## Description

HUM-47

### DIFF
--- a/packages/apps/human-app/frontend/package.json
+++ b/packages/apps/human-app/frontend/package.json
@@ -44,7 +44,6 @@
     "react-imask": "^7.4.0",
     "react-number-format": "^5.3.4",
     "react-router-dom": "^6.22.0",
-    "tldts": "^6.1.53",
     "vite-plugin-svgr": "^4.2.0",
     "zod": "^3.22.4",
     "zustand": "^4.5.0"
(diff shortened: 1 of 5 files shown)


portuu3 (dismissed):

portuu3 (dismissed):

KacperKoza343:
> @KacperKoza343 could you please update this branch with latest `develop`? It's necessary to have it in order to test after some recent changes

done

dnechay (commented):
Seems like something went wrong during url validation discussion 😅 

IMU we don't need to enforce http(s) (i.e. if entered - considered valid, but need to make sure that the whole string is valid URL.
E.g., webhook url can be something like `https://webhool-consumer.io/webhooks/hp?someParam=someValue`

You can achieve that either by custom function around URL or by using some smart regex

mpblocky:
@dnechay changes were added, ready for CR

dnechay (commented):
LGTM

One question before approval: is it high effort to change the behavior of operator form to validate values in live, i.e. once you start input, not waiting for "Add new Keys ..." click?
It is already reactive after you click this button, but would be nice to inform the user in live mode

mpblocky:
@dnechay it seems for me that sadly it's not a simple switch for that, as I see the form validation is pretty complicated. We surely can do that, however I warn that it might take some time. Should we proceed with the PR as is or should we change to live validation?

dnechay:
@mpblocky it's "nice to have" not "must have", so can be done later if needed 👌🏼

PR #2680, opened 2024-10-23
Worker registration

flopez7:
## Description

Fix registration page and rename registration-related stuff from Register to Registration

## Summary of changes

Fix registration page and rename registration-related stuff from Register to Registration

## How test the changes

Try to see jobs from an exchange oracle that has registration enabled in Human App. You'll see the new page with the instructions to register


packages/apps/fortune/exchange-oracle/server/src/modules/user/user.controller.ts
```diff
@@ -12,7 +12,7 @@ import { JwtAuthGuard } from '../../common/guards/jwt.auth';
 @UseGuards(JwtAuthGuard)
 @ApiBearerAuth()
 export class UserController {
-  @Post('register')
+  @Post('Registration')
```
portuu3, line 15: should be lower case

portuu3 (changes requested):

portuu3 (approved):

PR #2684, opened 2024-10-23
HUM-49

KacperKoza343:
## Description

HUM-49

### DIFF
--- a/packages/apps/human-app/frontend/src/api/services/common/get-access-token.ts
+++ b/packages/apps/human-app/frontend/src/api/services/common/get-access-token.ts
@@ -23,7 +23,7 @@ export function useGetAccessTokenMutation() {
     user: web3User,
   } = useWeb3Auth();
 
-  return useMutation({
+  const mutation = useMutation({
     mutationFn: async ({
       authType,
       throwExpirationModalOnSignOut = true,
@@ -73,4 +73,26 @@ export function useGetAccessTokenMutation() {
       await queryClient.invalidateQueries();
     },
   });
+
+  // Block new mutations if one is already loading
+  const mutateIfNotLoading = (args: Parameters<typeof mutation.mutate>[0]) => {
+    if (!mutation.isPending) {
+      mutation.mutate(args);
+    }
+  };
+
+  // Block new async mutations if one is already loading
+  const mutateAsyncIfNotLoading = async (
+    args: Parameters<typeof mutation.mutateAsync>[0]
+  ) => {
+    if (!mutation.isPending) {
+      return mutation.mutateAsync(args);
+    }
+  };
+
+  return {
+    ...mutation,
+    mutateIfNotLoading,
+    mutateAsyncIfNotLoading,
+  };
 }
--- a/packages/apps/human-app/frontend/src/api/services/common/get-access-token.ts
+++ b/packages/apps/human-app/frontend/src/api/services/common/get-access-token.ts
@@ -91,7 +91,6 @@ export function useGetAccessTokenMutation() {
   };
 
   return {
-    ...mutation,
     mutateIfNotLoading,
     mutateAsyncIfNotLoading,
   };
--- a/packages/apps/human-app/frontend/src/api/services/operator/disable-operator.ts
+++ b/packages/apps/human-app/frontend/src/api/services/operator/disable-operator.ts
@@ -7,8 +7,7 @@ import { useWeb3AuthenticatedUser } from '@/auth-web3/use-web3-authenticated-use
 
 export function useDisableWeb3Operator() {
   const { address, chainId } = useConnectedWallet();
-  const { refreshAccessTokenAsync: mutateAsyncIfNotLoading } =
-    useAccessTokenRefresh();
+  const { refreshAccessTokenAsync } = useAccessTokenRefresh();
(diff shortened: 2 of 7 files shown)


packages/apps/human-app/frontend/src/api/services/common/get-access-token.ts
```diff
@@ -73,4 +73,26 @@ export function useGetAccessTokenMutation() {
...
+      mutation.mutate(args);
+    }
+  };
+
+  // Block new async mutations if one is already loading
+  const mutateAsyncIfNotLoading = async (
+    args: Parameters<typeof mutation.mutateAsync>[0]
+  ) => {
+    if (!mutation.isPending) {
+      return mutation.mutateAsync(args);
```
dnechay, line 89:
```suggestion
      await mutation.mutateAsync(args);
```

packages/apps/human-app/frontend/src/api/services/common/get-access-token.ts
```diff
@@ -73,4 +73,26 @@ export function useGetAccessTokenMutation() {
...
+    }
+  };
+
+  return {
+    ...mutation,
```
dnechay, line 94:
```suggestion
```

dnechay (commented):
@KacperKoza343

Suggested solution seems working. It would be better if such race condition was handled in some centralized place (we also do refresh when data requests fail in `fetcher`), but it introduces significant refactoring to the application, that is not necessary now.

Wrt to this solution, I would like to suggest some changes here.

`useGetAccessTokenMutation` is not actually a "mutation" because we now "hijack" it's API and add some extra pieces for our needs. There are no consumers of `isPending`, `error` and other mutation values, it's being used as "fire & forget", so it would be cleaner if we change its name to something like `useAccessTokenRefresh` and inside of it create single `refreshAccessToken` async function with `isPending` logic, that uses `mutateAsync` and return only it.

As alternative solution we can use [mutation scopes](https://tanstack.com/query/v5/docs/framework/react/guides/mutations#mutation-scopes) with constant key. Even though it spawns new requests, it's being done in sequential way which is enough to address the bug and we don't have to build a workaround 

wdyt?

mpblocky:
@dnechay updated with the suggestions. I tried the scope id solution that You proposed but either I couldn't recreate the issue or it didn't work :confused:  

Ready for CR

dnechay:
> I tried the scope id solution that You proposed but either I couldn't recreate the issue or it didn't work

@mpblocky what was your setup and/or steps to reproduce this?
I tried `develop` branch, modified [jwt expiration check](https://github.com/humanprotocol/human-protocol/blob/caf7a33768ef1b7386f4b40adfeb71ffade40bd3/packages/apps/human-app/frontend/src/contexts/jwt-expiration-check.tsx#L37) to "simulate" expiration in next 10 seconds and was clicking between "my jobs" and "tasks" tabs to reproduce this (while being on `3G` network).

packages/apps/human-app/frontend/src/api/services/operator/disable-operator.ts
```diff
@@ -21,8 +22,9 @@ export function useDisableWeb3Operator() {
         },
       });
 
-      await mutateAsync({ authType: 'web3' });
+      await mutateAsyncIfNotLoading({ authType: 'web3' });
```
dnechay, line 25: Here and in other places, why not imported name `refreshAccessTokenAsync`/`refreshAccessToken`?
Using it as `mutate` is confusing

mpblocky: Sure, will fix that, I used auto rename in IDE

mpblocky:
@dnechay I added invalid bearer in `fetcher.ts`, added some queries to tasks and jobs discovery, clicked between them. Then I modified `jwt-expiration-check.tsx` to call refresh sometimes/all the time. Will try with Your setup and let You know about the results! Thanks for sharing

dnechay:
@mpblocky if option with `scope` also works for you - then I would prefer it since it's simpler and cleaner. Otherwise we can go with solution from this PR

mpblocky:
@dnechay I couldn't get Your way to achieve the bug, but I managed to do it by simply replacing `web2TokenExpired` with `true` in `jwt-expiration-check.tsx` (note that this heavily impacts the perfomance while testing the app). 

During testing like that:
- bug showed on develop
- the 1st solution was working
- the 2nd solution with scoped request worked as well

Please test it on Your side as well. I also reduced the amount of things returned from `useAccessTokenRefresh.ts` and renamed returned mutation functions.

dnechay:
@mpblocky 
Expiration time is a point in the future (in seconds; token TTL is 10 minutes), so in order to reproduce the expiration you need either to "rewind" it to the past, e.g.
```
const web2TokenExpired = Boolean(
  web2Auth.user?.exp && web2Auth.user.exp - 595 < Date.now() / 1000
);
```
or "rewind" the expiration point to compare 
```
const web2TokenExpired = Boolean(
  web2Auth.user?.exp && web2Auth.user.exp < (Date.now() + 1000 * 595) / 1000
);
```
where 595 - is 10 minutes w/o 5 seconds

dnechay (approved):

PR #2689, opened 2024-10-23
(HUM-64): change copy, wrap whitespace, move chip to next line

mpblocky:
## Description

HUM-64


dnechay (approved):
With new Grid it looks better 👍🏼

PR #2690, opened 2024-10-23
Hum 45 possibility to click on select job while one is in progress

KacperKoza343:
## Description

HUM-45

### DIFF
--- a/packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/desktop/available-jobs-table.tsx
+++ b/packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/desktop/available-jobs-table.tsx
@@ -36,12 +36,13 @@ export type AvailableJobsTableData = AvailableJob & {
   };
 };
 
-const getColumns = (
-  callbacks: {
-    assignJob: (data: AssignJobBody) => undefined;
-  },
-  disableButtons: boolean
-): MRT_ColumnDef<AvailableJob>[] => {
+const getColumns = ({
+  assignJob,
+  loadingButtons,
+}: {
+  assignJob: (data: AssignJobBody) => undefined;
+  loadingButtons: boolean;
+}): MRT_ColumnDef<AvailableJob>[] => {
   return [
     {
       accessorKey: 'job_description',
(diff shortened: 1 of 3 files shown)


packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/desktop/available-jobs-table.tsx
```diff
@@ -36,9 +36,12 @@ export type AvailableJobsTableData = AvailableJob & {
...
 };
 
-const getColumns = (callbacks: {
-  assignJob: (data: AssignJobBody) => undefined;
-}): MRT_ColumnDef<AvailableJob>[] => {
+const getColumns = (
+  callbacks: {
+    assignJob: (data: AssignJobBody) => undefined;
+  },
+  disableButtons: boolean
```
dnechay, line 43: It would be cleaner if we pass just object to `getColumns`, something like
```
{
  handleAssignJobClick: (data: AssignJobBody) => undefined,
  disableButtons: boolean
}
```


This way in place where you call `getColumns` it's clear what to expect and you don't have to check the signature of that function

packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/desktop/available-jobs-table.tsx
```diff
@@ -134,6 +137,7 @@ const getColumns = (callbacks: {
         return (
           <Grid sx={{ display: 'flex', justifyContent: 'flex-end' }}>
             <TableButton
+              disabled={disableButtons}
```
dnechay, line 140: Let's maybe use `loading` here for better UX?

packages/apps/human-app/frontend/src/pages/worker/jobs/components/available-jobs/mobile/available-jobs-table-mobile.tsx
```diff
@@ -165,6 +165,7 @@ export function AvailableJobsTableMobile({
                   <TableButton
                     color="secondary"
                     fullWidth
+                    disabled={status === 'pending'}
```
dnechay, line 168: Let's use `loading` state instead? Didn't now it was already implemented

dnechay (commented):
Fix looks good. Left some suggestions

dnechay (approved):

PR #2698, opened 2024-10-25
(HUM-52): dm button colors and states

mpblocky:
## Description

HUM-52

### DIFF
--- a/packages/apps/human-app/frontend/src/styles/dark-color-palette.ts
+++ b/packages/apps/human-app/frontend/src/styles/dark-color-palette.ts
@@ -6,7 +6,7 @@ export const darkColorPalette = {
   backgroundColor: 'rgba(16, 7, 53, 1)',
   text: {
     primary: '#D4CFFF',
-    secondary: 'rgba(212, 207, 255, 0.7)',
+    secondary: '#6C6978',
     disabled: 'rgba(212, 207, 255, 0.5)',
     disabledSecondary: 'rgba(147, 135, 255, 1)',
   },
--- a/packages/apps/human-app/frontend/src/styles/dark-color-palette.ts
+++ b/packages/apps/human-app/frontend/src/styles/dark-color-palette.ts
@@ -45,7 +45,7 @@ export const darkColorPalette = {
     main: 'rgba(203, 207, 232, 0.28)',
   },
   button: {
-    disabled: 'rgba(218, 222, 240, 0.8)',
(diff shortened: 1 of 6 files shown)


packages/apps/human-app/frontend/src/styles/dark-color-palette.ts
```diff
@@ -45,7 +45,7 @@ export const darkColorPalette = {
     main: 'rgba(203, 207, 232, 0.28)',
   },
   button: {
-    disabled: 'rgba(218, 222, 240, 0.8)',
+    disabled: '#FFFFFF1F',
```
dnechay, line 48: It's still different from the color framework in both modes

What we should have:
Screenshot 2024-10-28 at 13 35 55
What we actually have
Screenshot 2024-10-28 at 13 36 37

Screenshot 2024-10-28 at 13 36 47

Applies to all button variants

mpblocky:
image
image

@dnechay I picked colors from figma with color picker, should be good now. I also adjusted the text color, as it was indeed slightly different. While checking the colors pay attention to the background, as it also impacts how the button looks (whether it is an illusion or a transparency)

dnechay:
@mpblocky looks better now, however there are still buttons that mismatch the colors, e.g. in tables
Screenshot 2024-10-29 at 11 32 52
Screenshot 2024-10-29 at 11 32 54

Could you please revisit all (at least obvious) buttons and make sure they match?

wrt to the colors, it might be easier to clarify with designer what exact hex to use

mpblocky:
@dnechay should be fixed now

dnechay:
@mpblocky we have the [same thing with two-dot diff](https://github.com/humanprotocol/human-protocol/compare/develop..KacperKoza343:HUM-52-dm-button-colors-and-states)

Could you please clean up git history?

Next time instead of doing `git merge` try doing `git pull --rebase origin develop` and force push to remote if everything is ok. We should use merge commits only in case of complicated conflicts

mpblocky:
@dnechay all cleaned up 🧹

dnechay (approved):
@mpblocky Nice job! Looks much better now!

As a side-note: maybe we could discuss with the designer some options on how to handle color "overlap" when hover over table line?

It's hard to distinguish elements when you hover over table line (applicable only for light mode)
Screenshot 2024-10-29 at 18 47 09

mpblocky:
@dnechay I'll pass it on, we noticed the same thing :)

PR #2709, opened 2024-10-28
[HUM-48] fix router issues

KacperKoza343:
## Description

Task HUM-48


dnechay (approved):
Fix looks good
Would be nice to think about an option of moving it to some higher level, because looks like:
- `Tasks` menu item should still be selected when you go to captcha labeling or specific available job
- maybe we should group labeling and jobs discovery under `/tasks` route and decide if we allow entering it on a single places

But it's not for this PR

PR #2710, opened 2024-10-28
Hum 61 dm filters on mobile

KacperKoza343:
## Description

Task HUM-61


dnechay:
@KacperKoza343 we have the [same thing with two-dot diff](https://github.com/humanprotocol/human-protocol/compare/develop..KacperKoza343:HUM-61-dm-filters-on-mobile)

Could you please clean up git history? (i.e. removing merge commit from it)

Next time instead of doing `git merge` try doing `git pull --rebase origin develop` and force push to remote if everything is ok. We should use merge commits only in case of complicated conflicts

KacperKoza343:
> @KacperKoza343 we have the [same thing with two-dot diff](https://github.com/humanprotocol/human-protocol/compare/develop..KacperKoza343:HUM-61-dm-filters-on-mobile)
> 
> Could you please clean up git history? (i.e. removing merge commit from it)
> 
> Next time instead of doing `git merge` try doing `git pull --rebase origin develop` and force push to remote if everything is ok. We should use merge commits only in case of complicated conflicts

done

dnechay (approved):
Looks better with refactoring 👍🏼

PR #2711, opened 2024-10-28
[HUM-58] dm align icon with the typo

KacperKoza343:
## Description

Task HUM-58


dnechay (commented):
The fix is working, but it doesn't seem to be a good path of fixing it.

From what I see we have 128x129 icons in light mode and margins set for them where necessary. So question is why we have a different set of icons for `DarkModeIcons`?
Cant we have the same set of icons of 128x129 for dark mode and follow the same pattern instead of having icons of different size?

I could understand that if icons were completely different, but IMU they are just different in colors. So let's alight sizes in light/dark mode (either 88x92 or 128x129) and have clear UI code to handle that

KacperKoza343:
> The fix is working, but it doesn't seem to be a good path of fixing it.
> 
> From what I see we have 128x129 icons in light mode and margins set for them where necessary. So question is why we have a different set of icons for `DarkModeIcons`? Cant we have the same set of icons of 128x129 for dark mode and follow the same pattern instead of having icons of different size?
> 
> I could understand that if icons were completely different, but IMU they are just different in colors. So let's alight sizes in light/dark mode (either 88x92 or 128x129) and have clear UI code to handle that

Yes, you right. I could not get corresponding icon from Figma (128x129), finally I created proper icons based on 88x92. Now there is no need to have two sets of icons for home page

dnechay (commented):
It still needs proper size for `user` icon, otherwise layout is different

Screenshot 2024-10-30 at 20 02 45
Screenshot 2024-10-30 at 20 02 50

Work icon is still 126x126

dnechay:
@KacperKoza343 is this ready for review?

dnechay:
Probably not, because some alignments missing

Screenshot 2024-11-11 at 18 38 29

Screenshot 2024-11-11 at 18 38 26

dnechay (commented):
Main page seems fine, some alignments missing inside of app

Screenshot 2024-11-11 at 18 38 29

Screenshot 2024-11-11 at 18 38 26

dnechay (approved):
Looks good

PR #2715, opened 2024-10-28
(HUM-60): fix dropdown icon color in jobs discovery

mpblocky:
## Description

HUM-60

### DIFF
--- a/packages/apps/human-app/frontend/src/components/data-entry/multi-select.tsx
+++ b/packages/apps/human-app/frontend/src/components/data-entry/multi-select.tsx
@@ -134,6 +134,11 @@ export function MultiSelect({
               onChange={(event) => {
                 handleChange(event as SelectChangeEvent<string[]>, field);
               }}
+              sx={{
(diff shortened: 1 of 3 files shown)


packages/apps/human-app/frontend/src/components/data-entry/multi-select.tsx
```diff
@@ -134,6 +134,11 @@ export function MultiSelect({
               onChange={(event) => {
                 handleChange(event as SelectChangeEvent<string[]>, field);
               }}
+              sx={{
+                '.MuiSvgIcon-root': {
```
dnechay, line 138: It works, but we shouldn't be passing internal element styles here.

If you check default theming on MUI - each component (including select) is able to change their colors (including things like this "arrow icon") when changing modes, hence if color theme is defined properly - you don't have to pass such "hacks".

Please attempt to fix it in the right place and use "hacks" only if it's last resort

dnechay (approved):
That's definitely better :)

PR #2720, opened 2024-10-29
[CVAT] Adapt exchange/recording oracles for honeypots

Marishka17:
## Description
Adapted CVAT Exchange and Recording oracles to use the recently introduced CVAT honeypots. 

### Exchange oracle changes:
  -  The logic to select validation frames for injection into frames to be annotated has been moved into CVAT. CVAT tasks are now created with a validation layout (`gt_pool`) and a specified number of validation frames per job, defined by `manifest.validation.val_size` or `manifest.validation.val_size` * `job_size_mult` for skeleton tasks. Random sorting is used on CVAT when creating a task.
  -  Waiting unitl task is created is moved partially into `process_incoming_job_launcher_webhooks` cron job to be able upload GT (`CVAT_TASK_CREATION_CHECK_INTERVAL` env is added (default 5s)).
  - Oracle prepares GT annotations based on the GT dataset shared by the requester and uploads them to the CVAT GT job from the task. The `DEFAULT_IMPORT_TIMEOUT` environment is introduced to set the upper limit of waiting when GT annotations are uploaded. The default is 1h.
  - Oracle configures quality control settings for each CVAT task (there are several new env variables):
      - `CVAT_MAX_VALIDATION_CHECKS` - defines the maximum number of validation checks to be performed when checking job immediate feedback on CVAT (default is 3). Keep this number small enough to avoid honeypot recognition by annotators
      - `CVAT_IOU_THRESHOLD` (default 0.8)
      - `CVAT_OKS_SIGMA` (default 0.1, for skeletons task)
  -  The annotation meta created by exchange oracle and shared with recording oracle now includes new fields for each job meta:
      - task_id (cvat task id)
      - start and stop frames (obtained from CVAT)
 - Introduced a new setting `CVAT_MAX_JOBS_PER_TASK` (default 1000) that defines the maximum number of jobs that can be created in a cvat task. This allows us to avoid potential problems when working with unpaginated endpoints.
 - Fixed an issue where GT frames might be missing from a job
 - The `Job` model now includes new **non-nullable** fields: `start_frame` and `stop_frame`. That means that db must be reset. @Dzeranov 
 - Validation approach for the `image_points` task changed to compare user annotations with center points of GT boxes for compatibility with CVAT. This is going to be less accurate, so may lead to increased false positives and negatives during validation. Annotations are not changed.
 
### Recording oracle updates:
- CVAT’s quality control functionality is now used to determine the quality of job annotations. The custom logic for comparing datasets across different task types has been removed.
- The oracle now interacts directly with CVAT, using the same environment variables as the exchange oracle (`CVAT_URL`, `CVAT_ADMIN`, `CVAT_ADMIN_PASS`, and `CVAT_ORG_SLUG`) (+ added `cvat_sdk` dependency).
- The oracle doesn't use job annotations from the internal Exchange Oracle bucket anymore
- Jobs accepted on the previous epochs are no longer assessed again
- It's not possible now to have only some jobs unverifiable. Now, it's an escrow property, so either the whole escrow is verifiable or not. This depends on the number of remaining valid honeypots.
 
## Related issues


packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -345,21 +369,42 @@ def put_task_data(
...
+            if missed_filenames := set(gt_filenames) - set(filenames):
+                filenames.extend(missed_filenames)
+
+            kwargs["validation_params"] = models.DataRequestValidationParams(
+                mode=models.ValidationMode("gt_pool"),
+                frames=gt_filenames,
+                frame_selection_method=models.FrameSelectionMethod("manual"),
+                frames_per_job_count=validation_params.get(
+                    "gt_frames_per_job_count",
+                    Config.cvat_config.cvat_val_frames_per_job_count,
```
zhiltsov-max, line 389: This value should be taken from the manifest: `validation.val_size`

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -563,36 +608,194 @@ def clear_job_annotations(job_id: int) -> None:
...
+def get_gt_job(task_id: int) -> models.JobRead:
     logger = logging.getLogger("app")
 
     with get_api_client() as api_client:
         try:
-            api_client.jobs_api.partial_update(
-                id=id,
-                patched_job_write_request=models.PatchedJobWriteRequest(assignee=assignee_id),
+            (paginated_jobs, _) = api_client.jobs_api.list(task_id=task_id, type="ground_truth")
+            assert (
```
zhiltsov-max, line 625: I'd change this assert to an explicit check and exception, as it's not a developer (the caller) error.

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -563,36 +608,194 @@ def clear_job_annotations(job_id: int) -> None:
             raise
 
 
-def update_job_assignee(id: str, assignee_id: int | None):
+def setup_gt_job(task_id: int, dataset_path: Path, format_name: str) -> None:
+    gt_job = get_gt_job(task_id)
+    upload_gt_annotations(gt_job.id, dataset_path, format_name=format_name)
+    finish_gt_job(gt_job.id)
+    settings = get_quality_control_settings(task_id)
+    update_quality_control_settings(settings.id)
```
zhiltsov-max, line 616: Consider moving the function into `job_creation.py`. This file is supposed to contain generic functions without use case logic.

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -563,36 +608,194 @@ def clear_job_annotations(job_id: int) -> None:
...
+            logger.exception(f"Exception when calling QualityApi.list_settings(): {e}\n")
+            raise
+
+
+def update_quality_control_settings(
+    settings_id: int,
+    *,
+    max_validations_per_job: int = Config.cvat_config.cvat_max_validation_checks,
+    target_metric: str = "accuracy",
+    target_metric_threshold: float = Config.cvat_config.cvat_target_metric_threshold,
```
zhiltsov-max, line 722: This value should be taken from the manifest, `validation.min_quality`.

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -563,36 +608,194 @@ def clear_job_annotations(job_id: int) -> None:
...
+    low_overlap_threshold: float = Config.cvat_config.cvat_low_overlap_threshold,
```
zhiltsov-max, line 723: This value can be set to the same value as iou threshold or to 1, it only affects warnings

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2327,18 +2433,25 @@ def _job_params_label_key(ts):
         oracle_bucket = self.oracle_data_bucket
 
         # Register cloud storage on CVAT to pass user dataset
-        cvat_cloud_storage = cvat_api.create_cloudstorage(
-            **_make_cvat_cloud_storage_params(oracle_bucket)
+        _params = _make_cvat_cloud_storage_params(oracle_bucket)
+        _params["bucket_host"] = "http://minio:9010"
```
zhiltsov-max, line 2437:
```suggestion
        _params["bucket_host"] = "http://minio:9010" # TODO: remove mock
```

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2402,26 +2569,40 @@ def _job_params_label_key(ts):
...
                         task_id = db_service.create_task(
                             session, cvat_task.id, cvat_project.id, TaskStatuses[cvat_task.status]
                         )
                         db_service.get_task_by_id(session, task_id, for_update=True)  # lock the row
 
+                        # FUTURE-FIXME: now we must wait for the task to be created to set up GT
                         # Actual task creation in CVAT takes some time,
                         # so it's done in an async process.
                         # The task is fully created once 'update:task' or 'update:job'
                         # webhook is received.
```
zhiltsov-max, line 2590: The comment is not valid anymore, I guess

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2347,33 +2460,87 @@ def _job_params_label_key(ts):
...
+                        updated_points = [
+                            gt_points.points[0] - roi_info.roi_x,
+                            gt_points.points[1] - roi_info.roi_y,
+                        ]
+                        gt_points_dataset.put(
+                            dm.DatasetItem(
+                                id=dataset_item_id,
+                                annotations=[dm.Points(points=updated_points, label=0)],
+                            )
+                        )
```
zhiltsov-max, line 2528: Consider extracting a method for this. Also, I don't think it's a bad idea to save prepared GT annotations into the oracle bucket prior to any CVAT calls. It should definitely help with debugging and traceability, in the case of task creation problems.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2327,18 +2433,25 @@ def _job_params_label_key(ts):
...
 
         # Register cloud storage on CVAT to pass user dataset
-        cvat_cloud_storage = cvat_api.create_cloudstorage(
-            **_make_cvat_cloud_storage_params(oracle_bucket)
+        _params = _make_cvat_cloud_storage_params(oracle_bucket)
+        _params["bucket_host"] = "http://minio:9010"
+        cvat_cloud_storage = cvat_api.create_cloudstorage(**_params)
+
+        segment_size = (
+            self.manifest.annotation.job_size or Config.cvat_config.cvat_task_segment_size
```
zhiltsov-max, line 2441: Why there is `or`? Also, for skeleton tasks, there is a size multiplier currently (`DEFAULT_ASSIGNMENT_SIZE_MULTIPLIER`).

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2130,29 +2216,49 @@ def _prepare_job_params(self):
...
+                for point in skeleton.elements:
+                    point_label_name = (
+                        self._gt_dataset.categories()[dm.AnnotationType.label]
+                        .items[point.label]
+                        .name
+                    )
+
+                    # TODO: exclude such images
+                    if dm.Points.Visibility.absent in point.visibility:
+                        continue
```
zhiltsov-max, line 2258: Why such images should be excluded?

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2102,18 +2192,14 @@ def _prepare_job_params(self):
...
-        # Make job layouts wrt. manifest params
-        # 1 job per task, 1 task for each point label
-        #
-        # Unlike other task types, here we use a grid of RoIs,
-        # so the absolute job size numbers from manifest are multiplied by the job size multiplier.
-        # Then, we add a percent of job tiles for validation, keeping the requested ratio.
-        gt_ratio = self.manifest.validation.val_size / (self.manifest.annotation.job_size or 1)
-        job_size_mult = self.job_size_mult
-
-        job_params: list[self._JobParams] = []
```
zhiltsov-max, line 2114: Why this logic was removed? A multiplier is still used, even though there is no grid logic.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2102,18 +2192,14 @@ def _prepare_job_params(self):
...
+        task_params: list[self._TaskParams] = []
 
         roi_info_by_id = {roi_info.bbox_id: roi_info for roi_info in self._roi_infos}
+        self._roi_info_by_id = roi_info_by_id
+        segment_size = (
+            self.manifest.annotation.job_size or Config.cvat_config.cvat_task_segment_size
+        )
```
zhiltsov-max, line 2201: Read the other comment, there is a multiplier.

packages/examples/cvat/exchange-oracle/src/core/config.py
```diff
@@ -167,14 +167,24 @@ class CronConfig:
...
     cvat_url = os.environ.get("CVAT_URL", "http://localhost:8080")
     cvat_admin = os.environ.get("CVAT_ADMIN", "admin")
     cvat_admin_pass = os.environ.get("CVAT_ADMIN_PASS", "admin")
     cvat_org_slug = os.environ.get("CVAT_ORG_SLUG", "")
 
     cvat_job_overlap = int(os.environ.get("CVAT_JOB_OVERLAP", 0))
-    cvat_job_segment_size = int(os.environ.get("CVAT_JOB_SEGMENT_SIZE", 150))
+    cvat_task_segment_size = int(os.environ.get("CVAT_TASK_SEGMENT_SIZE", 150))
     cvat_default_image_quality = int(os.environ.get("CVAT_DEFAULT_IMAGE_QUALITY", 70))
+    cvat_max_jobs_per_task = int(os.environ.get("CVAT_MAX_JOBS_PER_TASK", 10 * 1000))
```
zhiltsov-max, line 179: What is this restriction?

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -1419,15 +1512,11 @@ def __init__(self, manifest: TaskManifest, escrow_address: str, chain_id: int) -
...
         self._roi_filenames: _MaybeUnset[dict[int, str]] = _unset
-        self._job_params: _MaybeUnset[list[self._JobParams]] = _unset
+        self._task_params: _MaybeUnset[list[self._TaskParams]] = _unset
 
         self._excluded_gt_info: _MaybeUnset[_ExcludedAnnotationsInfo] = _unset
         self._excluded_boxes_info: _MaybeUnset[_ExcludedAnnotationsInfo] = _unset
 
-        # Configuration / constants
-        self.job_size_mult = skeletons_from_boxes_task.DEFAULT_ASSIGNMENT_SIZE_MULTIPLIER
-        "Job size multiplier"
```
zhiltsov-max, line 1429: Did you decide to remove the size multiplier?

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -1284,17 +1314,57 @@ def _roi_key(e):
                     roi_bytes,
                 )
 
+    def split_dataset_per_task(
```
zhiltsov-max, line 1317: There is only the `build()` public method in these classes.

packages/examples/cvat/exchange-oracle/src/core/config.py
Marishka17: https://docs.google.com/presentation/d/17Hppt-uC_zAT_Qyr1Wiw60nd8ZFo3ZQfCqbKaT6vO_k/edit?usp=sharing

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -135,6 +137,42 @@ def add_message(self, message: str, *, sample_id: str, sample_subset: str):
...
+    while task_status != cvat_api.UploadStatus.FINISHED:
+        task_status, _ = cvat_api.get_task_upload_status(task_id)
+        if not task_status or task_status == cvat_api.UploadStatus.FAILED:
+            return  # will be handled in state_trackers.py::track_task_creation
+
+    if task_status == cvat_api.UploadStatus.FINISHED:
+        with TemporaryDirectory() as tmp_dir:
+            gt_dataset.export(
+                save_dir=tmp_dir, save_images=False, format=dm_export_format
+            )  # reindex=True ?
```
zhiltsov-max, line 167: Reindex will change image and annotation ids. If you're not going to use this dataset for anything except uploading, and it's possible that annotation ids are repeated, you can use the option.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -360,8 +400,11 @@ def __init__(self, manifest: TaskManifest, escrow_address: str, chain_id: int) -
         self._input_points_data: _MaybeUnset[bytes] = _unset
 
         self._data_filenames: _MaybeUnset[Sequence[str]] = _unset
+        self._data_filenames_to_be_annotated: _MaybeUnset[Sequence[str]] = _unset
         self._input_gt_dataset: _MaybeUnset[dm.Dataset] = _unset
         self._gt_dataset: _MaybeUnset[dm.Dataset] = _unset
+        self._gt_roi_dataset: _MaybeUnset[dm.Dataset] = _unset
+        self._gt_filenames: _MaybeUnset[dm.Dataset] = _unset
```
zhiltsov-max, line 407: Probably, invalid type

packages/examples/cvat/exchange-oracle/src/core/config.py
zhiltsov-max: Ok, makes sense.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Marishka17: Oh, I missed that `AnnotationInfo.job_size` is not nullable. I used `Config.cvat_config.cvat_task_segment_size` as backup value because it was [used](https://github.dev/humanprotocol/human-protocol/blob/dfefc91b976abcd35fa8a7f84582e7acc118ab49/packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py#L307) previously when creating a task on CVAT. I'll fix.

Marishka17: Just to clarify, in that case, we do not need to have the ability to configure task segment size based on some value from the exchange oracle config, right?

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
Marishka17: @zhiltsov-max, Could you please mark comments that you've fixed yourself in your PR?  Otherwise, it's not obvious for me which comments I should fix.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Marishka17: reverted logic related to `job_size_mult`

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Marishka17: Is there an option to skip frames that "should not be annotated" in CVAT `single shape mode`?

Marishka17: I just thought if we know from the GT that some frames contain outside points, why should we include such images in the dataset to be annotated? Why should we mislead an annotator?

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
zhiltsov-max: I remember HP wanted the oracle to handle all of this, but there are arguments against it, so we haven't come to a final solution. Let's extract a function for this now.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
zhiltsov-max: Yes, there is a skip button:

image

The image should be skipped in such situations.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
zhiltsov-max: Updated in https://github.com/humanprotocol/human-protocol/pull/2730

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
```diff
@@ -400,20 +392,33 @@ def process_intermediate_results(  # noqa: PLR0912
...
+            )
 
-            jobs_with_honeypots_to_be_shuffled = set(jobs_with_honeypots_to_be_shuffled)
-            cvat_api.shuffle_honeypots_in_jobs(jobs_with_honeypots_to_be_shuffled)
+            rng = np.random.default_rng()
+            upd_honeypot_real_frames = [
+                frame
+                if frame not in updated_disable_frames
+                else int(rng.choice(not_disabled_frames))
+                for frame in task_validation_layout.honeypot_real_frames
```
zhiltsov-max, line 414: It's possible to get several identical honeypots in a job with such an algorithm. Consider changing all the task frames (e.g. using the random method) or excluding frames that are already used in a job. For the latter you'll need to remember segment size somewhere, as in CVAT it will be 0.

zhiltsov-max: I just realized that using "random" method will update all the jobs, including already finished ones. This can be a problem if we export these jobs after they are updated, as their honeypot frames will have no annotations.

Marishka17: @zhiltsov-max, I see several solutions for this problem:
1. Utilize GET `/api/tasks/<id>/data/meta` endpoint to obtain metadata for the whole task and use `segment_size` from the task details. In that case, we need to update CVAT logic to return the actual segment size (annotation frames + honeypots). It's not clear to me why we manually set the value to 0.
2. If we cannot return the actual `segment_size` from CVAT, we can define it on the Exchange Oracle side (in generall it will be `manifest.annotation.job_size` + `manifest.validation.val_size`) and then pass it to the Recording Oracle via the `annotation_meta` file. I don't think this is the desired solution because this logic should be encapsulated inside CVAT.
3. Revert the previous approach and obtain a validation layout for each job. This approach also looks undesired since it requires +N requests to CVAT that can be throttled.
4. Introduce a parameter in PATCH `/api/tasks/<id>/validation_layout` that will define whether jobs with the acceptance stage should be included in the job queryset in which honeypots should be shuffled.  

Which approach do you prefer?

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Marishka17: done

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Marishka17: Now points with `dm.Points.Visibility.absent` or `dm.Points.Visibility.hidden` visibilities are excluded from the GT annotations uploaded to CVAT GT jobs. Related frames are not excluded from the annotation dataset.

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
Marishka17: I also see a small problem we shuffle honeypots in jobs before resetting the assignee. That means that an assignee will be able to open a job on CVAT and define which frames were honeypots/see new ones (before Exchange oracle handles webhook from Recording oracle).

zhiltsov-max: > It's not clear to me why we manually set the value to 0

There are 2 reasons for this:
- the value would be different from what's requested by the user during task creation (e.g. 4 normal + 2 validation frames would become 6)
- there are not segments in the task, really. The frames come in the random order, they are not slices of some range. This indicates that the "segments" are special

I think the second approach is the best for now. 4 is also somewhat reasonable, but the suggested design looks clumsy to me. Maybe we could change it to a list of relevant job ids instead, but it has some scalability issues.

Marishka17: > I also see a small problem we shuffle honeypots in jobs before resetting the assignee. That means that an assignee will be able to open a job on CVAT and define which frames were honeypots/see new ones (before Exchange oracle handles webhook from Recording oracle).

Not relevant anymore

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -326,42 +388,127 @@ def build(self):
...
+        assert self._mean_gt_bbox_radius_estimation is not _unset
+
+        values = {
+            # We have at most 1 annotation per frame, so accuracy is either 0 or 1.
+            # Point set accuracy is controlled by iou_threshold,
+            # so configure it with the requested value.
+            "target_metric_threshold": 0.9,
+            "iou_threshold": self.manifest.validation.min_quality,
+            "oks_sigma": self._mean_gt_bbox_radius_estimation,
+            "use_bbox_size_for_points": False,
```
zhiltsov-max, line 484:
```suggestion
            "point_size_base": "image_size",
```

https://github.com/cvat-ai/cvat/pull/8634

Dzeranov:
> @Dzeranov, Could you please check why vercel deployment has failed? I have no rights

@Marishka17 rebase should help you with that.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -1388,23 +1556,20 @@ def build(self):
         # Data preparation
         self._extract_and_upload_rois()
         self._upload_task_meta()
+        self._prepare_gt_roi_dataset()
```
zhiltsov-max, line 1559: It's better to move this before the ROI extraction and uploading call to fail faster in case of errors, if possible.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -2091,7 +2253,19 @@ def _mangle_filenames(self):
...
-    def _prepare_job_params(self):
+        self._input_filename_to_nameless = {
+            roi_info.original_image_id: self._roi_filenames[roi_info.bbox_id]
+            for roi_info in self._roi_infos
+        }
+
+    @property
+    def _task_segment_size(self):
+        # Unlike other task types, here we use a grid of RoIs,
+        # so the absolute job size numbers from manifest are multiplied by the job size multiplier.
```
zhiltsov-max, line 2264:
```suggestion
        ```suggestion
        # Here we use a job size multiplier, because each image 
        # is supposed to be simple and the assignment is expected 
        # to take little time with the default job size.
```
```

File: packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py [hunk 22]
```diff
@@ -563,36 +610,200 @@ def clear_job_annotations(job_id: int) -> None:
...
+
+def update_quality_control_settings(
+    settings_id: int,
+    *,
+    target_metric_threshold: float,
+    target_metric: str = "accuracy",
+    max_validations_per_job: int = Config.cvat_config.cvat_max_validation_checks,
+    iou_threshold: float = Config.cvat_config.cvat_iou_threshold,
+    oks_sigma: float | None = None,
+    use_bbox_size_for_points: bool | None = None,
```
zhiltsov-max, line 726:
```suggestion
    point_size_base: str | None = None,
```

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -563,36 +610,200 @@ def clear_job_annotations(job_id: int) -> None:
...
+        "target_metric_threshold": target_metric_threshold,
+        "iou_threshold": iou_threshold,
+        "low_overlap_threshold": iou_threshold,  # used only for warnings
+    }
+
+    if oks_sigma is not None:
+        params["oks_sigma"] = oks_sigma
+
+    if use_bbox_size_for_points is not None:
+        params["use_bbox_size_for_points"] = use_bbox_size_for_points
```
zhiltsov-max, line 742:
```suggestion
    if point_size_base is not None:
        params["point_size_base"] = point_size_base
```

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
```diff
@@ -167,56 +128,86 @@ def _parse_gt(self):
...
+                    self._gt_stats[(cvat_task_id, val_frame)].failed_attempts += 1
+                else:
+                    self._gt_stats[(cvat_task_id, val_frame)].accepted_attempts += 1
+
+            # assess job quality
+            job_quality_report = job_id_to_quality_report[cvat_job_id]
+
+            accuracy = job_quality_report.summary.accuracy
+            if isinstance(accuracy, int):
+                assert accuracy == 0
```
zhiltsov-max, line 197: What does this check mean?

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
```diff
@@ -980,17 +389,121 @@ def process_intermediate_results(  # noqa: PLR0912
...
+
+                for val_frame in validation_frames:
+                    val_frame_stats = gt_stats[(cvat_task_id, val_frame)]
+                    if (
+                        val_frame_stats.failed_attempts >= Config.validation.gt_ban_threshold
+                        and not val_frame_stats.accepted_attempts
+                    ):
+                        cvat_task_id_to_failed_val_frames.setdefault(cvat_task_id, set()).add(
+                            val_frame
+                        )
```
zhiltsov-max, line 423: If I understand correctly, we can have the same GT frame in different tasks, and they can have different statistics, so that the same frame can be banned in 1 task, but not another. It doesn't look right, can we change gt stats to be escrow-wise?

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
Marishka17: It was used to identify situations where there are no valid gt frames in the job, resulting in an undefined quality. Probably it was not the best approach, replaced by `job_quality_report.summary.gt_count`

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
Marishka17: done

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
zhiltsov-max: `gt_count` is the number of gt annotations. I think it's better to check it using `validation_layout` instead. Also, I'm not sure accuracy has to be 0 in such cases.

Marishka17: > Also, I'm not sure accuracy has to be 0 in such cases.

Are you talking about the value returned by CVAT? In that case, it is.

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
```diff
@@ -980,17 +405,126 @@ def process_intermediate_results(  # noqa: PLR0912
...
+                    honeypots_mapping[honeypot]
+                    for honeypot in job_honeypots
+                    if honeypots_mapping[honeypot] in val_frame_ids
+                ]
+
+                # choose new unique validation frames for the job
+                available_validation_frames = list(
+                    set(non_disabled_frames) - set(validation_frames_to_replace)
+                )
+                rng = np.random.default_rng()
```
zhiltsov-max, line 488: Please use a specific known generator. We don't know when the default might change. MT19937 would be ok.

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
```diff
@@ -980,17 +405,126 @@ def process_intermediate_results(  # noqa: PLR0912
...
+                try:
+                    new_validation_frames = [
+                        int(f)
+                        for f in rng.choice(
+                            available_validation_frames,
+                            replace=False,
+                            size=len(validation_frames_to_replace),
+                        )
+                    ]
```
zhiltsov-max, line 497:
```suggestion
                    new_validation_frames = rng.choice(
                        available_validation_frames,
                        replace=False,
                        size=len(validation_frames_to_replace),
                    ).tolist()
```

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
zhiltsov-max: Ok. Actually, could you clarify how we can get this situation now? It was possible before frames could be changed in jobs, but now we always should have all the jobs verifiable, unless we have more than `val_size` images in the pool.

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
```diff
@@ -980,17 +405,126 @@ def process_intermediate_results(  # noqa: PLR0912
...
+                    "Logical error occurred while disabling validation frames "
+                    f"for the task({task_id}). Frames {intersection} "
+                    "are already disabled."
+                )
+
+            updated_disable_frames = task_validation_layout.disabled_frames + list(val_frame_ids)
+            non_disabled_frames = list(
+                set(task_validation_layout.validation_frames) - set(updated_disable_frames)
+            )
+
```
zhiltsov-max, line 459: It makes sense to check there that there are at least `frames_per_job_count` items in `non_disabled_frames`. In this case the task (and, probably, the escrow) becomes unverifiable.

Marishka17: It's checked when we try to select new N validation frames from the `non_disabled_frames`. In that case, we do not shuffle honeypots (updated_task_honeypot_real_frames is not changed)

packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py
Marishka17: Rigth now such a situation is possible: when recording oracle cannot select new N validation frames, we just update disabled frames in the task. That means that related honeypots are marked as deleted in related jobs, right? So it is possible to achieve the situation, when all validation frames are disabled. 

> but now we always should have all the jobs verifiable, while we have at least val_size images in the pool.

What should we do in case when all validation frames contain incorrect annotations?

zhiltsov-max: > What should we do in case when all validation frames contain incorrect annotations?

By the current policy, we stop the escrow annotation with the current annotations without checking them. This part is [here](https://github.com/humanprotocol/human-protocol/blob/34bbdf364473573e2af8bc8049ce6a71b7eff540/packages/examples/cvat/recording-oracle/src/handlers/process_intermediate_results.py#L562-L586).

> Rigth now such a situation is possible: when recording oracle cannot select new N validation frames, we just update disabled frames in the task. That means that related honeypots are marked as deleted in related jobs, right? So it is possible to achieve the situation, when all validation frames are disabled.

Discussed in https://github.com/humanprotocol/human-protocol/pull/2720#discussion_r1834014156

zhiltsov-max: If it's not possible to select N new validation frames, it means we don't have enough GT to validate any of the jobs, since each CVAT task has the same GT frames. The escrow annotation has to be finished. You stop any interactions with CVAT, no new annotation iterations should happen for any of the escrow jobs.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -326,42 +388,127 @@ def build(self):
...
-                # The task is fully created once 'update:task' or 'update:job' webhook is received.
+                # The task is fully created once 'update:task' webhook is received.
                 cvat_api.put_task_data(
                     cvat_task.id,
                     cloud_storage.id,
-                    filenames=job_filenames,
+                    filenames=data_subset,
                     sort_images=False,
+                    # use the same value for the chunk size as for the job size
+                    chunk_size=segment_size,
```
zhiltsov-max, line 410: Need to include validation frames for chunk sizes

Marishka17: Do we need to add an upper limit for the `chunk_size`?

zhiltsov-max: No, I don't think so. The parameter should be removed or redesigned later

packages/examples/cvat/recording-oracle/src/cvat/interface.py
```diff
@@ -0,0 +1,14 @@
+from pydantic import BaseModel
+
+
+class _AnnotationInfo(BaseModel):
+    accuracy: float | int
+
+
+class _FrameResult(BaseModel):
+    conflicts: list[dict]
+    annotations: _AnnotationInfo
```
zhiltsov-max, line 10: If `QualityReportData` is public, it's strange to make it's public components private.

packages/examples/cvat/recording-oracle/src/cvat/api_calls.py
```diff
@@ -0,0 +1,182 @@
...
+
+def get_task_quality_report(
+    task_id: int,
+    *,
+    max_waiting_time: int = 10 * 60,
+    sleep_interval: float = 0.5,
+) -> models.QualityReport:
+    logger = logging.getLogger("app")
+    report = get_last_task_quality_report(task_id)
+    if report and report.created_date > report.target_last_updated:
```
zhiltsov-max, line 87: The condition `report.created_date > report.target_last_updated` is always true

packages/examples/cvat/exchange-oracle/src/cvat/api_calls.py
```diff
@@ -563,36 +610,204 @@ def clear_job_annotations(job_id: int) -> None:
...
+        "low_overlap_threshold": iou_threshold,  # used only for warnings
+    }
+
+    if oks_sigma is not None:
+        params["oks_sigma"] = oks_sigma
+
+    if point_size_base is not None:
+        params["point_size_base"] = point_size_base
+
+    if match_empty_frames:
```
Marishka17, line 745: @zhiltsov-max
```suggestion
    if match_empty_frames is not None:
```

packages/examples/cvat/exchange-oracle/src/core/tasks/skeletons_from_boxes.py
```diff
@@ -18,6 +18,7 @@
 @frozen(kw_only=True)
 class RoiInfo:
     original_image_key: int
+    original_image_id: str
```
zhiltsov-max, line 21: I can't find where this field is used. Is it only for debugging?

Marishka17: It was used before refactoring but doesn’t seem to be needed now

Marishka17:
@zhiltsov-max, Is there anything else required from my side? Do you have any other comments or..?

zhiltsov-max:
@Marishka17, need to test more, no other actions needed so far.

packages/examples/cvat/exchange-oracle/src/handlers/completed_escrows.py
zhiltsov-max, line 1: Could you update the annotation downloading code here to use the v2 downloading protocol? Currently, it uses pre-fixed v1, so there are repeated error messages in the log, because the code tries to download annotations without receiving the "export compete" status.

zhiltsov-max:
Need to implement the idea from https://github.com/humanprotocol/human-protocol/pull/2720/commits/16c67b04042af7c435b5944d5c35368d24611066

zhiltsov-max (approved):

packages/examples/cvat/exchange-oracle/src/core/config.py
```diff
@@ -167,14 +170,30 @@ class CronConfig:
...
```
zhiltsov-max, line 182: @Marishka17, could you check this default value? Shouldn't it be like just 1k instead?

Marishka17: Initially, I thought that 1k jobs per task value was more suited for a sandbox. But okay, let's reduce that.
```suggestion
    cvat_max_jobs_per_task = int(os.environ.get("CVAT_MAX_JOBS_PER_TASK", 1000))
```

Dzeranov (approved):

Issue #2717, opened 2024-10-29
[SDK] Treat messages/files for encryption/decryption as binaries not strings


Dzeranov:
**Description**
Need to change the way of handling messages/files for encryption/decryption. Currently it's a string but for a better interoperability it should be a buffer.

**Motivation**
Current implementation won't work properly in case message to encrypt/decrypt is not a string (.zip for example).

**Implementation details** 
TBD by developer.

PR #2732, opened 2024-10-31
[SDK][Python] fix: encryption module now can work not only with strings

Dzeranov:
## Description

Encryption module now can work not only with strings

## Summary of changes

1. Both `sign_and_encrypt` and `sign` methods now can work both with `str` and `bytes` types.
2. `decrypt` method now returns `bytes` instead of `str`

## Related issues
Part of the #2717

**_NOTE:_** Don't merge this PR.

### DIFF
--- a/packages/sdk/python/human-protocol-sdk/human_protocol_sdk/encryption/encryption.py
+++ b/packages/sdk/python/human-protocol-sdk/human_protocol_sdk/encryption/encryption.py
@@ -40,7 +40,7 @@ Module
 ------
 """
 
-from typing import Optional, List
+from typing import Optional, List, Union
 from pgpy import PGPKey, PGPMessage
 from pgpy.constants import SymmetricKeyAlgorithm
 from pgpy.errors import PGPError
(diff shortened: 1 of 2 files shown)


portuu3 (approved):

dnechay (commented):
LGTM overall, 2 questions:
- does it make sense to add a unit test on encrypt/decrypt of actual `bytes`? Maybe some dumb object converted to bytes
- do we expect the same 1:1 behavior from JS sdk, i.e. `decrypt` to return string if encrypted message was string and `encrypt` to accept both? I thought for these methods we were going to use just bytes and potentially provide some utility methods like `encryptStrMessage`, `decryptStrMessage`

Dzeranov:
@dnechay 
1. I'll add it, this PR is not ready to be merged yet, simply to show what will be the behavior.
2. `decrypt` method returns `bytes` in this implementation for both `str` and `bytes`. As for input parameters, `pgpy` package doesn't care what is the data you need to encrypt, `Union[str, bytes]` is simply for a type hint here.

PR #2734, opened 2024-10-31
[HUM-37] clarify colors on statuses on my jobs table

KacperKoza343:
## Description

HUM-37


packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
```diff
@@ -14,7 +14,19 @@ const myJobSchema = z.object({
   escrow_address: z.string(),
   chain_id: z.number(),
   job_type: z.string(),
-  status: z.string(),
+  status: z.string().transform((value) => {
+    try {
+      return z
```
dnechay, line 19: We can use `z.enum(['ACTIVE', 'CANCELED', 'COMPLETED'])` here

dnechay (commented):
Nice!

Let's make another improvement here: we use hardcoded status values (e.g. `ACTIVE`) here and there, so would be better to move it to some sort of enum and use this enum in schema validation and other places.

packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
mpblocky: Added in new commit

dnechay (commented):
WRT to coloring status - lgtm

Could you please move status values to enum and refactor usage of hardcoded values?

packages/apps/human-app/frontend/src/shared/helpers/get-chip-status-color.ts
```diff
@@ -0,0 +1,21 @@
...
+      return colorPalette.secondary.main;
+    case MyJobStatusWithUnknown.COMPLETED:
+      return colorPalette.success.main;
+    case MyJobStatusWithUnknown.VALIDATION:
+      return colorPalette.error.light;
+    case MyJobStatusWithUnknown.CANCELED:
+    case MyJobStatusWithUnknown.UNKNOWN:
+    case MyJobStatusWithUnknown.EXPIRED:
+    case MyJobStatusWithUnknown.REJECTED:
+      return colorPalette.error.main;
```
mpblocky, line 19: @dnechay we don't have colors for those statuses specified in figma, is it okay like it is now?

dnechay: @mpblocky yes, it's fine

packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
```diff
@@ -9,12 +9,44 @@ import { createPaginationSchema } from '@/shared/helpers/create-pagination-schem
...
+export enum MyJobStatus {
+  ACTIVE = 'ACTIVE',
+  CANCELED = 'CANCELED',
+  COMPLETED = 'COMPLETED',
+  VALIDATION = 'VALIDATION',
+  EXPIRED = 'EXPIRED',
+  REJECTED = 'REJECTED',
+}
+
+// Repeated because of linting rule
```
mpblocky, line 21: FYI: linting rule prevented me from creating `MyJobStatusWithUnknown` with all the values based directly on the `MyJobStatus`. The `UNKNOWN` is a type for handling unknown statuses on frontend, hence why it's not in the filters list. 
Using utility types was not an option as there were places where `MyJobStatus` was used as a value.

dnechay: @mpblocky 
> The UNKNOWN is a type for handling unknown statuses on frontend, hence why it's not in the filters list.

That's legit, we don't need this value in filters list. At the same time, we don't need to have a second enum.
You can have `MyJobStatus` enum as it as and the define next:

```
const UNKNOWN_JOB_STATUS = 'UNKNOWN';

export type MyJobStatusWithUnknown =
  | `${MyJobStatus}`
  | typeof UNKNOWN_JOB_STATUS;
```
in case you want exact types.

At the same time, the only place where you need it - is `getChipStatusColor`, so instead of having this `MyJobStatusWithUnknown`, you can simply change `getChipStatusColor` to accept `status: string` and return default value when not active|completed|validation. Right now this `getChipStatusColor` look like a common utility that can be used anywhere else because it's in shared helpers, but why we use `MyJobStatusWithUnknown` type then? It makes more sense to remove enum usage here and use hardcoded value if it's shared utility.

If we don't plan to have it like shared - let's move it some other place close to "my jobs" and keep `MyJobStatusWithUnknown` with unknown for it and `throw` in `default` block of switch-case statement

packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
```diff
@@ -9,12 +9,44 @@ import { createPaginationSchema } from '@/shared/helpers/create-pagination-schem
...
 const myJobSchema = z.object({
   assignment_id: z.string(),
   escrow_address: z.string(),
   chain_id: z.number(),
   job_type: z.string(),
-  status: z.string(),
+  status: z.string().transform((value) => {
+    try {
+      return z
+        .enum([
```
dnechay, line 40: Shouldn't it be `.nativeEnum(MyJobStatus)`?
From what I see we can have at least `expired` jobs, so I assume all of them needed.

dnechay (commented):
Looks better with `enum`!

Let's get rid of duplicated enum values and fix validation for API response

packages/apps/human-app/frontend/src/api/services/worker/my-jobs-data.ts
mpblocky: This way we have two options, as we are using this enum in here:
`const buttonDisabled = d.status !== MyJobStatusWithUnknown.ACTIVE;`

1. delete `MyJobStatusWithUnknown` altogether and just use magic string `'UNKNOWN'` along with `MyJobsStatus` which will result in:
` const buttonDisabled = d.status !== MyJobStatus.ACTIVE && d.status !== UNKNOWN_JOB_STATUS;`

2. Keep the current solution.

Which would You prefer?

About the plan I'm not sure if we are going the use the chip anywhere else so it's entirely Your decision too :sweat_smile:

dnechay: @mpblocky option 1, but I don't get why we need to check for `d.status !== UNKNOWN_JOB_STATUS;`?

Button should be enabled only if `d.status === MyJobStatus.ACTIVE`, otherwise disabled (which is just `d.status !== MyJobStatus.ACTIVE`), so you don't have to check for unknown

mpblocky: True, now it works - must have something broken with my eslint since it complained about no overlap :thinking: Will do the 1st version

packages/apps/human-app/frontend/src/shared/helpers/get-chip-status-color.ts
```diff
@@ -0,0 +1,16 @@
+import { type ColorPalette } from '@/styles/color-palette';
+
+export function getChipStatusColor(status: string, colorPalette: ColorPalette) {
+  const normalizedStatus = status.toUpperCase();
+
+  switch (normalizedStatus) {
+    case 'ACTIVE':
```
dnechay, line 7: Why using hardcoded values and not values from enum here now (it was from enum in previous commits)?

This `getChipStatusColor` is placed in very common `shared/helpers`, but does every chip in app will have same set of statuses and colors? I suppose - no, so this utility takes place only for "jobs".

Let's do some kind of file/utility renaming, use enums instead of hardcoded values (it's their purpose)

Also not sure why we need to `const normalizedStatus = status.toUpperCase();` if we already do it while validating response schema

mpblocky: We had a misunderstanding then :sweat_smile: We discussed these changes [here](https://github.com/humanprotocol/human-protocol/pull/2734#discussion_r1836818687) and since the proposed argument type was `string` I assumed that we will have to normalize it and compare strings as well; the `string` type also indicated that it will be used in places other than `jobs` so I left it there. All fixed in new commit!

dnechay: True, I didn't make myself clear, sorry

> the string type also indicated that it will be used in places other than jobs so I left it there.

I meant the general case if we had many different chips, but falling into the same set of rules in terms of color, which is hard to imagine tbh because they rely on status value that in general is specific for an entity (job in this case)

Thank you for double-checking it and pointing what was confusing!

dnechay (commented):
@mpblocky please resolve merge conflicts and it's good to go

dnechay (approved):

PR #2739, opened 2024-10-31
[CVAT][Exchange Oracle] Job and assignments statuses to lowercase

Dzeranov:
## Description

Job/assignment statuses transformed to a lower case in API responses

## Summary of changes

To align with changes in #2610

### DIFF
--- a/packages/examples/cvat/exchange-oracle/src/schemas/exchange.py
+++ b/packages/examples/cvat/exchange-oracle/src/schemas/exchange.py
@@ -7,13 +7,13 @@ from strenum import StrEnum  # added in python 3.11
 from src.core.types import Networks, TaskTypes
 from src.utils.enums import BetterEnumMeta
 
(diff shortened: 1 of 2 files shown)


zhiltsov-max (approved):

packages/examples/cvat/exchange-oracle/src/schemas/exchange.py
```diff
@@ -7,13 +7,13 @@
 from src.core.types import Networks, TaskTypes
 from src.utils.enums import BetterEnumMeta
 
-DEFAULT_TOKEN = "HMT"  # noqa: S105 (it's  not a credential)
+DEFAULT_TOKEN = "hmt"  # noqa: S105 (it's  not a credential)
```
zhiltsov-max, line 10: I'm not sure about the token, as it's an abbreviation, unlike others.

PR #2763, opened 2024-11-08
Add debug.py for ease of local development

Bobronium:
## Context behind the change
- Add `debug.py` to both Exchange and Recording Oracles. It's equivalent to `run.py` except it also contains patches simplifying local development (before the common practice was to edit pieces of code for ease of local development). This should simplify this process even further.

## How has this been tested?
This has been tested locally. No changes to application code were made.

## Release plan
Doesn't involve changes to production code/setup.


packages/examples/cvat/exchange-oracle/debug.py
zhiltsov-max, line 1:
```
        "bucket_host": bucket_info.host_url.replace(
            "127.0.0.1",
            "172.22.0.1",
        ),
```

Please add a solution for this substitution in `_make_cvat_cloud_storage_params()` (`job_creation.py`) as well. Maybe it can be an env variable. It is needed to make minio reachable from CVAT.

Bobronium: Good idea. I've added two new ENV variables:

- `CVAT_IN_DOCKER`
- `CVAT_LOCAL_HOST`

When `CVAT_IN_DOCKER` is set, `debug.py` will substitute host in CVAT storage url with one from `CVAT_LOCAL_HOST`.

zhiltsov-max (approved):

zhiltsov-max:
@Dzeranov, please merge.

PR #2781, opened 2024-11-13
[Reputation Oracle] feat: get rid of sdk's storage module

### LINKED ISSUE #2719: [Reputation Oracle] Get rid of StorageModule
****Description
Remove StorageModule usage and replace it's usages.

Motivation
StorageModule from our SDK is deprecated and is going to be deleted soon.

Implementation details
TBD by developer
### _**Important note**_:
Make sure that downloading/uploading files works well not only with strings.

dnechay:
## Issue tracking
Resolves #2719 

## Context behind the change
Storage module in SDK is going to be deprecated, so we get rid of it here. At the same time existing storage utils have a bug in working with binaries: when doing `copyFileFromURLToBucket` it's being downloaded as string instead of binary (which is expected to be a `.zip`) and it leads to file in invalid format being copied; this PR also fixed that.

## How has this been tested?
- [x] unit tests
- [x] e2e using Minio: create fake route handler to call `copyFileFromURLToBucket` for some encrypted `.zip` file and make sure it's copied to proper bucket;
- [x] get copied file, decrypt it using JL private key, make sure that result is valid `.zip` file that can opened

## Release plan
Merge & deploy as is.
Only open question is if we need to fix some files that already uploaded as string instead of zip?!

Can't think of any

### DIFF
--- a/packages/apps/job-launcher/server/tsconfig.json
+++ b/packages/apps/job-launcher/server/tsconfig.json
@@ -7,7 +7,7 @@
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
     "allowJs": true,
-    "target": "ES2020",
+    "target": "ES2022",
     "sourceMap": true,
     "outDir": "./dist",
     "baseUrl": "./",
--- a/packages/apps/reputation-oracle/server/docker-compose.yml
+++ b/packages/apps/reputation-oracle/server/docker-compose.yml
@@ -26,12 +26,9 @@ services:
     ports:
       - 9001:9001
       - 9000:9000
-    environment:
-      MINIO_ROOT_USER: ${S3_ACCESS_KEY}
-      MINIO_ROOT_PASSWORD: ${S3_SECRET_KEY}
     entrypoint: 'sh'
     command:
-      -c "mkdir -p /data/reputation && minio server /data --console-address ':9001'"
+      -c "minio server /data --console-address ':9001'"
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:9000/minio/health/live"]
       interval: 5s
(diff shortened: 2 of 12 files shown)


packages/apps/job-launcher/server/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
     "allowJs": true,
-    "target": "ES2020",
+    "target": "ES2022",
```
dnechay, line 10: By some reason using `ErrorOptions` (which is part of es2022) worked for JL when it shouldn't. I spend many hours trying to understand why it happened and had no luck, probably something just messed up in global monorepo setup.

Set proper target to avoid future confusion

packages/apps/reputation-oracle/server/docker-compose.yml
```diff
@@ -45,7 +42,7 @@ services:
         condition: service_healthy
     entrypoint: >
       /bin/sh -c "
-      /usr/bin/mc config host add myminio http://minio:9000 ${S3_ACCESS_KEY} ${S3_SECRET_KEY};
+      /usr/bin/mc config host add myminio http://minio:9000 minioadmin minioadmin;
```
dnechay, line 45: This and removal of env variables need to easily work with it out of the box, aligned with job launcher

packages/apps/reputation-oracle/server/src/modules/storage/storage.service.spec.ts
```diff
@@ -108,6 +105,10 @@ describe('StorageService', () => {
     jest.spyOn(pgpConfigService, 'encrypt', 'get').mockReturnValue(true);
   });
 
+  beforeEach(() => {
+    downloadFileFromUrlSpy = jest.spyOn(StorageService, 'downloadFileFromUrl');
```
dnechay, line 109: Mocks in this test file are completely ruined and conflict with each other, so defining new spy before each test is needed to as a temp solution. We are going to revisit reputation oracle, tests will be part of that

Dzeranov (approved):

PR #2794, opened 2024-11-15
[HUM-72] fix: remove metamask error from walletconnect

mpblocky:
Added Wagmi adapter instead of ether adapter which removed the metamask window.web3 error.

HUM-72


packages/apps/human-app/frontend/src/contexts/wallet-connect.tsx
```diff
@@ -22,15 +32,13 @@ const metadata = {
   icons: env.VITE_DAPP_ICONS,
 };
 
-const ethersConfig = defaultConfig({
-  metadata,
-});
-createWeb3Modal({
-  ethersConfig,
-  chains,
+const networks: [AppKitNetwork, ...AppKitNetwork[]] = [polygon, polygonAmoy];
```
mpblocky, line 35: TODO

mpblocky:
@dnechay we need to discuss what steps should we take next. This PR removes ethers adapter and changes it to wagmi adapter. It also updates the `walletconnect` to `reown appkit`. This fixed the metamask throwing error with `window.web3`. We still use `ethers` for contract calls, to not dig through whole codebase. 

However You can still find that error when entering worker sign in, because of hcaptcha. We do not have any control over it's implementation, so fixing it is out of our reach.

What do You think about it? Should we still push to merge this PR, if the hcaptcha throws and still works? The same situation was with ethers adapter - metamask was throwing the error but everything still worked fine.

packages/apps/human-app/frontend/src/contexts/wallet-connect.tsx
```diff
@@ -94,49 +112,55 @@ export function WalletConnectProvider({
...
-              openModal,
-              signMessage: async (message: string) => {
-                try {
-                  const signature =
-                    await web3ProviderMutation.data.signer.signMessage(message);
-                  return signature;
-                } catch (error) {
-                  throw new JsonRpcError(error);
+    <WagmiProvider config={wagmiAdapter.wagmiConfig}>
+      <QueryClientProvider client={client}>
```
dnechay, line 116: Do you know why we need `QueryClientProvider` here?

dnechay:
@mpblocky 

> We still use ethers for contract calls, to not dig through whole codebase

If I understand correctly, `wagmi` requires `viem` as peerDependency, so we actually might need to install it. If we do - why not migrating to it from `ethers`? Maybe as a separate PR if you prefer, but at least we will have consistent libraries used

Do we plan to cleanup changes introduced in https://github.com/humanprotocol/human-protocol/pull/2816 &  https://github.com/humanprotocol/human-protocol/pull/2819 as part of this issue?

dnechay:
@mpblocky 

> We still use ethers for contract calls, to not dig through whole codebase

If I understand correctly, `wagmi` requires `viem` as peerDependency, so we actually might need to install it. If we do - why not migrating to it from `ethers`? Maybe as a separate PR if you prefer, but at least we will have consistent libraries used

> However You can still find that error when entering worker sign in, because of hcaptcha. We do not have any control over it's implementation, so fixing it is out of our reach.

It's not our scope of responsibility. If on our end hCaptcha is properly configured and works as expected but throws some error from inside due to outdated packages or something - we shouldn't care until they introduce some fix that we can catch up with 

> What do You think about it? Should we still push to merge this PR, if the hcaptcha throws and still works?

Yes. What we use in our app and what they load in their iframe - are two different things

> The same situation was with ethers adapter - metamask was throwing the error but everything still worked fine.

Could you please elaborate on this? I probably miss some context from the past

packages/apps/human-app/frontend/package.json
```diff
@@ -44,7 +42,9 @@
     "react-imask": "^7.4.0",
     "react-number-format": "^5.3.4",
     "react-router-dom": "^6.22.0",
+    "viem": "2.21.43",
```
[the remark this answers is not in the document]
mpblocky: yep, works without it! Sadly the docs are outdated in a lot of places

mpblocky:
@dnechay 

> If I understand correctly, wagmi requires viem as peerDependency, so we actually might need to install it. If we do - why not migrating to it from ethers? Maybe as a separate PR if you prefer, but at least we will have consistent libraries used

Yes, I paid attention to that too - that's why I mentioned digging through codebase; right now we needed to get rid of the error and if You want we can migrate to `wagmi` from `ethers` in another PR. @BartoszSolkaBD 

> Could you please elaborate on this? I probably miss some context from the past

The ethers adapter was working fine with that error present, hence our doubts whether we should migrate the adapter to wagmi or not

dnechay:
@mpblocky I would prefer to not have different adapters in the code to avoid surprises and to ease maintenance, so would be nice to migrate it, yes.
Indeed, it would be better to do it in a separate PR (or bunch of small PRs with similar APIs migrated) so it's easier to test if things keep working as expected

yarn.lock
```diff
@@ -1972,6 +1971,11 @@
     "@ethersproject/properties" "^5.7.0"
     "@ethersproject/strings" "^5.7.0"
 
+"@fastify/busboy@^2.0.0":
```
dnechay, line 1974: Don't forget to run `yarn install` from the root in order to have proper packages resolutions, this one shouldn't be there

dnechay:
@mpblocky sorry for confusion [here](https://github.com/humanprotocol/human-protocol/pull/2794#discussion_r1850665226), actually `viem` is required as peer dependency of `wagmi` package
Could you please install it:
```
// to install
yarn workspace @human-protocol/human-app-frontend add viem@^2.21.44
// to respect resolutions
yarn install
```

dnechay (approved):

PR #2824, opened 2024-11-20
[HUM-73]: fix: remove refresh token race condition

mpblocky:
HUM-73

### DIFF
--- a/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
+++ b/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
@@ -1,0 +1,31 @@
+import { signInSuccessResponseSchema } from '@/api/services/worker/sign-in';
+import { apiPaths } from '@/api/api-paths';
+import { browserAuthProvider } from '@/shared/helpers/browser-auth-provider';
+
+export const fetchTokenRefresh = async () => {
+  const response = await fetch(apiPaths.worker.obtainAccessToken.path, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      // eslint-disable-next-line camelcase -- camel case defined by api
+      refresh_token: browserAuthProvider.getRefreshToken(),
+    }),
+  });
+
+  if (!response.ok) {
+    throw new Error('Failed to refresh token');
+  }
+
+  const data: unknown = await response.json();
+
+  const refetchAccessTokenSuccess = signInSuccessResponseSchema.parse(data);
+
+  browserAuthProvider.signIn(
+    refetchAccessTokenSuccess,
+    browserAuthProvider.authType
+  );
+
+  return refetchAccessTokenSuccess;
+};
--- a/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
+++ b/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
@@ -18,17 +18,12 @@ export const fetchTokenRefresh = async (baseUrl: string) => {
   );
 
   if (!response.ok) {
-    throw new Error('Failed to refresh token');
+    return null;
   }
 
   const data: unknown = await response.json();
 
   const refetchAccessTokenSuccess = signInSuccessResponseSchema.parse(data);
-
-  browserAuthProvider.signIn(
-    refetchAccessTokenSuccess,
-    browserAuthProvider.authType
-  );
 
   return refetchAccessTokenSuccess;
 };
--- a/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
+++ b/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
@@ -2,17 +2,20 @@ import { signInSuccessResponseSchema } from '@/api/services/worker/sign-in';
 import { apiPaths } from '@/api/api-paths';
 import { browserAuthProvider } from '@/shared/helpers/browser-auth-provider';
 
-export const fetchTokenRefresh = async () => {
-  const response = await fetch(apiPaths.worker.obtainAccessToken.path, {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
-    },
-    body: JSON.stringify({
-      // eslint-disable-next-line camelcase -- camel case defined by api
-      refresh_token: browserAuthProvider.getRefreshToken(),
-    }),
-  });
+export const fetchTokenRefresh = async (baseUrl: string) => {
+  const response = await fetch(
+    `${baseUrl}${apiPaths.worker.obtainAccessToken.path}`,
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        // eslint-disable-next-line camelcase -- camel case defined by api
+        refresh_token: browserAuthProvider.getRefreshToken(),
+      }),
+    }
+  );
 
   if (!response.ok) {
     throw new Error('Failed to refresh token');
--- a/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
+++ b/packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
@@ -1,6 +1,6 @@
-import { signInSuccessResponseSchema } from '@/api/services/worker/sign-in';
 import { apiPaths } from '@/api/api-paths';
 import { browserAuthProvider } from '@/shared/helpers/browser-auth-provider';
+import { signInSuccessResponseSchema } from '@/api/services/worker/sign-in/schema';
 
 export const fetchTokenRefresh = async (baseUrl: string) => {
   const response = await fetch(
--- a/packages/apps/human-app/frontend/src/api/fetcher.ts
+++ b/packages/apps/human-app/frontend/src/api/fetcher.ts
@@ -71,6 +71,9 @@ export function createFetcher(defaultFetcherConfig?: {
   options?: RequestInit | (() => RequestInit);
   baseUrl: FetcherUrl | (() => FetcherUrl);
 }) {
+  let isRefreshing = false;
+  let refreshPromise: Promise<SignInSuccessResponse | undefined> | null = null;
+
   async function fetcher<SuccessInput, SuccessOutput>(
     url: string | URL,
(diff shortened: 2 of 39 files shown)


mpblocky:
This approach was taken as it was necessary to preserve explicit token refresh through the hook. This solution is also the least impactful on whole codebase which avoids unforseen regression.

The easiest way to test it is to:
1. add buttons in profile page to call for some authenticated url
2. add refresh token button (using refresh hook)
3. save current access token somewhere
4. refresh token
5. turn on 3g throttling
6. add old access token to force refresh
7. click multiple times on the button
8. optionally click on refresh token button

This way in the network tab we can clearly see that there is always one in-flight refresh token request.

packages/apps/human-app/frontend/src/api/fetcher.ts
```diff
@@ -71,6 +71,46 @@ export function createFetcher(defaultFetcherConfig?: {
   options?: RequestInit | (() => RequestInit);
   baseUrl: FetcherUrl | (() => FetcherUrl);
 }) {
+  let isRefreshing = false;
+  let refreshPromise: Promise<SignInSuccessResponse | undefined> | null = null;
+
+  async function refreshToken(): Promise<
```
dnechay, line 77: Do we need to have it inside of `createFetcher` or can move outside?
We don't use `baseUrl` or `options`, `refreshToken` relies on `apiClient.fetcher` to make its request (which looks like circular dependency).

Would be nice to move it outside and to separate file in order to remove circular dependency.

packages/apps/human-app/frontend/src/api/fetcher.ts
```diff
@@ -71,6 +71,46 @@ export function createFetcher(defaultFetcherConfig?: {
...
+    { access_token: string; refresh_token: string } | null | undefined
+  > {
+    if (!isRefreshing) {
```
dnechay, line 80: You don't need both `isRefreshing` and promise ref, just promise ref is enough.

Let's refactor this function to something more clean, i.e. moving anonymous function with actual logic to separate function and keep `refreshToken` only to call it and handle "single running version" logic.

You can find [example in the fiddle](https://jsfiddle.net/dmitry_hmt_ai/dw056keq/16/)

dnechay (commented):
Approach lgtm! Let's make it a bit cleaner :)

packages/apps/human-app/frontend/src/api/fetcher.ts
mpblocky: Yes, that's actually something I wanted to get rid of using plain fetch, good catch!

packages/apps/human-app/frontend/src/api/fetcher.ts
mpblocky: Sure!

packages/apps/human-app/frontend/src/api/fetcher.ts
```diff
@@ -71,6 +68,25 @@ export function createFetcher(defaultFetcherConfig?: {
...
+  let refreshPromise: Promise<SignInSuccessResponse | undefined> | null = null;
+
+  async function refreshToken(): Promise<
+    { access_token: string; refresh_token: string } | null | undefined
+  > {
+    if (!refreshPromise) {
+      try {
+        refreshPromise = fetchTokenRefresh();
+      } catch (error) {
+        browserAuthProvider.signOut({ triggerSignOutSubscriptions: true });
```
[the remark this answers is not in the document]
mpblocky: Done as in the fiddle, can You check?

mpblocky:
@dnechay I also removed the dependency cycle

packages/apps/human-app/frontend/src/api/fetch-refresh-token.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!response.ok) {
+    throw new Error('Failed to refresh token');
+  }
+
+  const data: unknown = await response.json();
+
+  const refetchAccessTokenSuccess = signInSuccessResponseSchema.parse(data);
+
+  browserAuthProvider.signIn(
```
dnechay, line 28: Function is called `fetchTokenRefresh`, so it's assumed to have only one purpose: fetch new token, parse response and return it.

The logic with `browserAuthProvider.signIn` (and also `browserAuthProvider.signOut` for errors, that is missed while refactoring) should be in `refreshToken` that calls `fetchTokenRefresh`

packages/apps/human-app/frontend/src/api/fetcher.ts
```diff
@@ -71,6 +69,22 @@ export function createFetcher(defaultFetcherConfig?: {
   options?: RequestInit | (() => RequestInit);
   baseUrl: FetcherUrl | (() => FetcherUrl);
 }) {
+  let refreshPromise: Promise<SignInSuccessResponse | undefined> | null = null;
+
+  async function refreshToken(): Promise<
```
dnechay, line 74: As per [this comment](https://github.com/humanprotocol/human-protocol/pull/2824/files#r1853643880), now we can move it out of `createFetcher` and below it will still
```
return fetcher
```

packages/apps/human-app/frontend/src/api/fetcher.ts
```diff
@@ -153,32 +167,10 @@ export function createFetcher(defaultFetcherConfig?: {
...
-        );
-      } catch {
-        browserAuthProvider.signOut({ triggerSignOutSubscriptions: true });
-        return;
-      }
+      const refetchAccessTokenSuccess = await refreshToken();
 
       const newHeaders = appendHeader(fetcherOptionsWithDefaults, {
-        Authorization: `Bearer ${refetchAccessTokenSuccess.access_token}`,
+        Authorization: `Bearer ${refetchAccessTokenSuccess?.access_token}`,
```
dnechay, line 173: If success - it should be always defined and have `access_token`.
So there are two options (will depend on `refreshToken` implementation):
- if `refreshToken` throws in case of failed refresh - we should have `try..catch` and `return` in `catch` block (like it was)
- if `refreshToken` returns `null` in case of refresh not successful, we simply
```
const refreshTokenResult = await refreshToken();
if (!refreshTokenResult) {
  return;
}
```

packages/apps/human-app/frontend/src/api/services/common/use-access-token-refresh.ts
```diff
@@ -32,19 +30,11 @@ export function useAccessTokenRefresh() {
...
-              body: JSON.stringify({
-                // eslint-disable-next-line camelcase -- camel case defined by api
-                refresh_token: browserAuthProvider.getRefreshToken(),
-              }),
-            },
-          }
-        );
+        const refetchAccessTokenSuccess = await apiClient.refreshToken();
+
+        if (!refetchAccessTokenSuccess) {
```
dnechay, line 35: Don't forget to update code here according to new implementation of `refreshToken`

packages/apps/human-app/frontend/src/api/fetcher.ts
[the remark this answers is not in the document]
dnechay: @mpblocky this part looks better now! Let's make sure success/error cases are handled properly :)

packages/apps/human-app/frontend/src/api/fetcher.ts
mpblocky: Just to clarify: we want to actually separate `fetcher` from `refresh` completely? Go back from `apiClient.fetcher` to `apiClient` and use explicit `refreshToken` rather than `apiClient.refreshToken`?

mpblocky: Sure, sorry for that

dnechay: Thanks for jumping into with clarifications, let me share my though process :)

From naming `createFetcher` - I would expected that such "factory function" creates and returns just a "fetcher" instance that I would later user to make some requests, this is why I initially though about moving it out (+the fact that `refreshToken` doesn't use anything from `createFetcher` options at all, so why it would be there)

> Go back from apiClient.fetcher to apiClient and use explicit refreshToken rather than apiClient.refreshToken?

See, here you have different naming - `apiClient`, and when having such naming it actually makes sense to have: `apiClient.fetch()` and `apiClient.refreshToken()`

So no need to move it out, we can just remove that confusion by:
- using `baseUrl` in `fetchTokenRefresh()` instead of `env.VITE_API_URL`, since we create `apiClient` with specific config
- rename `createFetcher` to `createApiClient`, because it's not just generic "fetcher", it's highly coupled to the API it uses
- return `{ fetch, refreshToken }`

This way we will be explicit on what we have, wdyt?

dnechay: @mpblocky Thanks for jumping into with clarifications, let me share my though process :)

From naming `createFetcher` - I would expected that such "factory function" creates and returns just a "fetcher" instance that I would later user to make some requests, this is why I initially though about moving it out (+the fact that `refreshToken` doesn't use anything from `createFetcher` options at all, so why it would be there)

> Go back from apiClient.fetcher to apiClient and use explicit refreshToken rather than apiClient.refreshToken?

See, here you have different naming - `apiClient`, and when having such naming it actually makes sense to have: `apiClient.fetch()` and `apiClient.refreshToken()`

So no need to move it out, we can just remove that confusion by:
- using `baseUrl` in `fetchTokenRefresh()` instead of `env.VITE_API_URL`, since we create `apiClient` with specific config
- rename `createFetcher` to `createApiClient`, because it's not just generic "fetcher", it's highly coupled to the API it uses
- return `{ fetch, refreshToken }`

This way we will be explicit on what we have, wdyt?

mpblocky: Both approaches make sense actually - it boils down to if we want to have explicit usage of `env` in fetchRefreshToken or have a singular place where we take it from

dnechay: True, so feel free to pick any of them, just not the mix 😅

mpblocky: lets stay with the current state - there would be additional changes needed and with the current code I'm not sure how much changes would be needed :sweat_smile:

dnechay (approved):

PR #2831, opened 2024-11-20
[CVAT] Improve quality setup for point annotation task

zhiltsov-max:
## Context behind the change

- Improved target quality setting for the point annotation task
- Improved the reasoning comment

### DIFF
--- a/packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
+++ b/packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
@@ -511,10 +511,28 @@ class PointsTaskBuilder(SimpleTaskBuilder):
         assert self._mean_gt_bbox_radius_estimation is not _unset
 
         values = {
-            # We have at most 1 annotation per frame, so accuracy is either 0 or 1.
+            # We have at most 1 annotation per frame, so accuracy on each frame is either 0 or 1,
+            # regardless of the points count. For job accuracy we'll have:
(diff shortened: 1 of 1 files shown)


Dzeranov (approved):

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -511,10 +511,28 @@ def _setup_quality_settings(self, task_id: int, **overrides) -> None:
...
+            # For each frame, as we have just 1 annotation, quality is computed as mean
+            # point quality on the frame with frame matching threshold.
             # Point set accuracy is controlled by iou_threshold,
-            # so configure it with the requested value.
-            "target_metric_threshold": 0.9,
+            # so configure it with the requested quality value.
+            #
+            # TODO: consider adding a quality option to count points as separate,
+            # or implement and use the mean IoU as the target metric in CVAT.
+            "target_metric_threshold": 0.95,  # some big number close to 1
```
Bobronium, line 535: Why does it have to be big number close to one? 

From my understanding any number in range (0, 1] would do. If it's the case, why not just pick one (since we expect the accuracy to be 1 anyway)?

zhiltsov-max: It would be ok to pick 1, but I'm just not sure that 1 will work well on all the steps of the pipeline.

PR #2722, opened 2024-10-29
[CVAT] Add polygons support

Bobronium:
## Description
Add initial Polygons support to Exchange and Recording Oracles.
Validation is implemented through converting polygons to bounding boxes and using existing `bbox_iou` function. 
This is not accurate by any means, but should be enough until we move validation to CVAT.

## Summary of changes
- Add required types for polygons support

## How test the changes
Considering you're running `docker-compose.dev.yml`:
- Upload [cheemsdoge_manifest.json](https://github.com/user-attachments/files/17633307/cheemsdoge_manifest.json) to http://127.0.0.1:9010/manifests/cheemsdoge_manifest.json
- Upload [cheemsdoge_gt.json](https://github.com/user-attachments/files/17633305/cheemsdoge_gt.json) to http://127.0.0.1:9010/launcher
- And upload these to http://127.0.0.1:9010/launcher/polygons-data:
	- [cheems.jpg](https://github.com/user-attachments/assets/9d2cf3b6-bf5d-430c-a1a4-f1924035a869)
	- [cheemsdoge_gt.jpg](https://github.com/user-attachments/assets/a4696e96-d06a-485b-9a97-177ba16e0b07)

Then, go through the usual process of creating, annotating and submitting annotations. You should see annotations results uploaded to http://127.0.0.1:9010/recor-data/:9010/cheemsdoge_manifest.json@80002/resulting_annotations.zip

PR #2832, opened 2024-11-22
[Exchange Oracle] Add `iou_threshold` for polygon tasks and enable single shape mode

Bobronium:
## Context behind the change
Allow `iou_threshhold` setting to be configured for polygons. Default value is 0.5 which should be sufficient for the most tasks. Increasing the value will require more precision and will likely cause more false-negatives. It might make sense to increase it for segmentation tasks, though experimentation is needed.

## How has this been tested?
Test process was similar to one described in #2722.

## Release plan; Potential risks; What to monitor; Rollback plan
Doesn't introduce and breaking changes, DB migrations nor public API changes.


packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
```diff
@@ -550,6 +550,12 @@ def build(self):
         return super().build()
 
 
+class PolygonTaskBuilder(SimpleTaskBuilder):
+    def _setup_quality_settings(self, task_id: int, **overrides) -> None:
+        values = {"iou_threshold": Config.core_config.polygons_iou_threshold, **overrides}
```
zhiltsov-max, line 555:
```suggestion
        values = {"iou_threshold": Config.core_config.polygons_iou_threshold}
        values.update(**overrides)
```
The code can fail if there is `iou_threshold` in `overrides`.

packages/examples/cvat/exchange-oracle/src/core/config.py
```diff
@@ -258,6 +258,9 @@ class CoreConfig:
     min_roi_size_h = int(os.environ.get("MIN_ROI_SIZE_H", 300))
     "Minimum absolute ROI size for image_boxes_from_points and image_skeletons_from_boxes tasks"
 
+    polygons_iou_threshold = float(os.environ.get("POLYGONS_IOU_THRESHOLD", 0.8))
```
zhiltsov-max, line 261: Please move it into the `CvatConfig` class, next to other similar variables.

packages/examples/cvat/exchange-oracle/src/handlers/job_creation.py
Bobronium: Python allows duplicate keys in dict literal, the last key will remain so result would be identical in this case.

```python
print({"foo": 1, "foo": 2})  # {"foo": 2}
print({"foo": 1, **{"foo": 2}})  # {"foo": 2}
```

You might have confused this with function arguments expansion in function calls (usually `**kwargs`), in this case, python indeed doesn't allow multiple arguments to be present.

packages/examples/cvat/exchange-oracle/src/core/config.py
Bobronium: Good catch. Done.

packages/examples/cvat/exchange-oracle/src/core/config.py
```diff
@@ -195,6 +195,9 @@ class CvatConfig:
     cvat_iou_threshold = float(os.environ.get("CVAT_IOU_THRESHOLD", 0.8))
     cvat_oks_sigma = float(os.environ.get("CVAT_OKS_SIGMA", 0.1))
 
+    cvat_polygons_iou_threshold = float(os.environ.get("CVAT_POLYGONS_IOU_THRESHOLD", 0.5))
```
zhiltsov-max, line 198: Please update `.env.template` with the new variable

Bobronium: Done

zhiltsov-max (approved):

Issue #2638, opened 2024-10-15
[Reputation Oracle] Refactor `processPaidWebhooks` method


Dzeranov:
**Description**
Need to rework `processPaidWebhooks` cron job.
Desired behaviour:
In case `await escrowClient.complete()` is successful we don't need to try completing the escrow again.
In case `webhookService.sendWebhook()` fails, we want to launch a cron job only for resending failed webhook(s).

My suggestion here is to implement some kind of a webhook queue because we are sending webhooks to multiple instances (JL, ExO, RecO) and we want to track which notification failed to be sent.

**Motivation**
With current implementation in case escrow has been successfully completed but webhook wasn't sent, cron job will be launched again trying to update escrow status to `completed` and this operation will fail causing webhook to be stuck.

**Implementation details** 
To be provided by the assignee.

eugenvoronov:
### Implemented: 
- [x] **Update `WebhookIncomingEntity` model:**
   - Rename `WebhookIncomingEntity` to `WebhookEntity`.
   - Add an enum field `type` to `WebhookEntity` with values: `in` and `out`.
   - Add a `failedReason` field to store the reason for failed webhooks.
   -  Add a `url` field to store the webhook url.

- [x] **Cron Module to track `in` webhooks:**
   - Monitor `WebhookEntity` entries with `type: in` and status `pending`.
   - For each webhook, check the current escrow status:
     - If escrow is incomplete, invoke `escrowClient.complete()`.
     - If escrow fails, increment `retriesCount` and retry the operation.

- [x] **Create webhooks:**
   - Generate three `WebhookEntity` instances with `type: out` and status `pending`.

- [x] **Update `in` webhook status:**
   - Update the status of the `WebhookEntity` with `type: in` to `completed`.

- [x] **Cron module to track `out` webhooks:**
   - Monitor `WebhookEntity` entries with `type: out ` and status `pending`.
   - For each `out` webhook:
     - Use `webhookService.sendWebhook()` to send the notification.
     - If sending fails, increment `retriesCount` and store the `failedReason`.
     - Upon successful delivery, update the related `in` webhook status to `completed`.
- [x] **Update migrations**  
- [x] **Update unit tests**

PR #2834, opened 2024-11-25
[Reputation Oracle] Refactoring Cron Module

eugenvoronov:
## Issue tracking
This change is being tracked via a [Reputation Oracle] Refactor processPaidWebhooks method
[#2638](https://github.com/humanprotocol/human-protocol/issues/2638), the discussion in comment https://github.com/humanprotocol/human-protocol/pull/2658#discussion_r1830811808, which outlines the need for restructuring the webhook handling logic in the cron-job service.
The requirement is to ensure that the `try..catch block` is fully contained within the `webhookService` while the `cron-job` service should only handle job execution and locking.

## Context behind the change
In the previous implementation, the `cron-job` service was responsible for handling the webhook logic, including error handling with `try..catch`. This made the service do more than it should, as its main responsibility is to schedule and ensure a lock to avoid parallel job execution.

This pull request refactors the code by moving the `try..catch` logic into the `webhookService`. The `cron-job` service will now solely handle job execution scheduling.

## How has this been tested?
I have tested the changes locally by running the cron job and ensuring that the `webhookService` properly handles errors through the `try..catch` block. Additionally, I improved unit tests.


packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -187,61 +124,7 @@ export class CronJobService {
...
-            `Error processing escrow completion tracking. Error ID: ${errorId}, Escrow completion tracking ID: ${escrowCompletionTrackingEntity.id}, Reason: ${failureDetail}, Message: ${err.message}`,
-          );
-          await this.escrowCompletionTrackingService.handleEscrowCompletionTrackingError(
-            escrowCompletionTrackingEntity,
-            failureDetail,
-          );
-          continue;
-        }
-      }
+      await this.webhookService.processPendingEscrowCompletion();
```
dnechay, line 127: Should `processPendingEscrowCompletion` be part of `EscrowCompletiontTracking` service?

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -271,130 +154,12 @@ export class CronJobService {
...
-          this.logger.error(
-            `Error processing escrow completion tracking. Error ID: ${errorId}, Escrow completion tracking ID: ${escrowCompletionTrackingEntity.id}, Reason: ${failureDetail}, Message: ${err.message}`,
-          );
-          await this.escrowCompletionTrackingService.handleEscrowCompletionTrackingError(
-            escrowCompletionTrackingEntity,
-            failureDetail,
-          );
-        }
-      }
+      await this.webhookService.processPaidEscrowCompletion();
```
dnechay, line 157: Should it be in `EscrowCompletiontTracking` service?

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.ts
eugenvoronov: Yeah, I thought about that, I ran into a cyclic dependency issue where the Escrow module uses the Webhook module, and the Webhook module uses the Escrow module. Do you have any ideas on how to refactor the process to avoid cyclic dependencies?

eugenvoronov:
> It's a bit confusing that we have methods to process escrow completion in `webhook` service. IMU `webhook` is for working with webhooks, not to have escrow completion or payments logic.
> 
> However, if we move aforementioned logic to `EscrowCompletionService` - we might get circular dependency between it and webhooks service due to:
> 
> * `processPendingIncomingWebhooks` creates `escrowCompletionTracking` entities
> * `processPaidEscrowCompletion` creates outgoing webhooks
>   So we might need to find a workaround for that, probably by splitting incoming/outgoing webhooks services

Yeah, I thought about that, I ran into a cyclic dependency issue where the Escrow module uses the Webhook module, and the Webhook module uses the Escrow module. I will figured out how to solve it.

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.spec.ts
```diff
@@ -1,830 +1,369 @@
...
+      );
 
-      expect(createUniqueSpy).toHaveBeenCalledWith({
-        cronJobType: CronJobType.ProcessPendingIncomingWebhook,
-      });
-      expect(result).toEqual(cronJobEntity);
+      expect(cronJobRepository.findOneByType).toHaveBeenCalledWith(
+        CronJobType.ProcessPendingIncomingWebhook,
+      );
+      expect(cronJobRepository.createUnique).toHaveBeenCalled();
```
dnechay, line 82: Would be nice to still check that it has been called with proper arguments

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.spec.ts
```diff
@@ -1,830 +1,369 @@
...
-      expect(result).toEqual(cronJobEntity);
+      const result = await service.startCronJob(
+        CronJobType.ProcessPendingIncomingWebhook,
+      );
 
-      jest.useRealTimers();
+      expect(cronJobRepository.findOneByType).toHaveBeenCalledWith(
+        CronJobType.ProcessPendingIncomingWebhook,
+      );
+      expect(cronJobRepository.updateOne).toHaveBeenCalledWith(updatedCronJob);
```
dnechay, line 105: Under the hood `startCronJob` updates `startedAt = new Date()`, so this condition might be flaky if tests run slowly.

Let's either fake timers or use `expect.objectContaining` and not check the exact value, but up to some second

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.spec.ts
```diff
@@ -1,830 +1,369 @@
...
+      cronJobRepository.updateOne.mockResolvedValue(completedCronJob as any);
 
       const result = await service.completeCronJob(cronJobEntity);
 
-      expect(updateOneSpy).toHaveBeenCalled();
-      expect(cronJobEntity.completedAt).toEqual(mockDate);
-      expect(result).toEqual(cronJobEntity);
-
-      jest.useRealTimers();
+      expect(cronJobRepository.updateOne).toHaveBeenCalledWith(
```
dnechay, line 162: This `expect` is flaky now (I got it failed to 1ms when running on local), please consider to keep fake timers or use `expect.objectContaining`

packages/apps/reputation-oracle/server/src/modules/escrow-completion-tracking/escrow-completion-tracking.module.ts
```diff
@@ -5,11 +5,19 @@ import { TypeOrmModule } from '@nestjs/typeorm';
 import { EscrowCompletionTrackingEntity } from './escrow-completion-tracking.entity';
```
dnechay, line 5: Maybe we rename this module to `escrow-completion`? It works not only with `escrow-completion-tracking` entity and does some other things, so should be cleaner

packages/apps/reputation-oracle/server/src/modules/webhook/webhook-outgoing.service.ts
```diff
@@ -138,19 +90,44 @@ export class WebhookService {
...
+        const { url, payload } = webhookEntity;
+
+        await this.sendWebhook(url, payload);
+      } catch (err) {
+        const errorId = uuidv4();
+        const failureDetail = `${ErrorWebhook.PendingProcessingFailed} (Error ID: ${errorId})`;
+        this.logger.error(
+          `Error processing pending outgoing webhook. Error ID: ${errorId}, Webhook ID: ${webhookEntity.id}, Reason: ${failureDetail}, Message: ${err.message}`,
+        );
+        await this.handleWebhookOutgoingError(webhookEntity, failureDetail);
```
dnechay, line 126: `handleWebhookOutgoingError` and similar methods in other classes now don't have to be `public` and can be tested as part of `process*` method tests

packages/apps/reputation-oracle/server/src/modules/webhook/webhook-incoming.service.ts
```diff
@@ -0,0 +1,119 @@
...
+    webhookEntity.chainId = dto.chainId;
+    webhookEntity.escrowAddress = dto.escrowAddress;
+    webhookEntity.status = WebhookIncomingStatus.PENDING;
+    webhookEntity.waitUntil = new Date();
+    webhookEntity.retriesCount = 0;
+
+    webhookEntity =
+      await this.webhookIncomingRepository.createUnique(webhookEntity);
+
+    if (!webhookEntity) {
```
dnechay, line 52: No need in this check

packages/apps/reputation-oracle/server/src/modules/webhook/webhook-incoming.module.ts
```diff
@@ -5,26 +5,27 @@ import { TypeOrmModule } from '@nestjs/typeorm';
...
+    EscrowCompletionTrackingModule,
   ],
   controllers: [WebhookController],
   providers: [
     Logger,
-    WebhookService,
+    WebhookIncomingService,
     WebhookIncomingRepository,
-    WebhookOutgoingRepository,
+    EscrowCompletionTrackingRepository,
```
dnechay, line 27: Should it be service and not repository?

I might need to check deeper how nest.js modules system works, but seems that we don't need that at all because it imports `EscrowCompletionTrackingModule` above

packages/apps/reputation-oracle/server/src/modules/escrow-completion-tracking/escrow-completion-tracking.service.ts
```diff
@@ -70,4 +93,177 @@ export class EscrowCompletionTrackingService {
...
+            )
+          ).webhookUrl,
+          (
+            await OperatorUtils.getLeader(
+              chainId,
+              await escrowClient.getExchangeOracleAddress(escrowAddress),
+            )
+          ).webhookUrl,
+          (
+            await OperatorUtils.getLeader(
```
dnechay, line 198: It was disabled in `cron`, was it uncommented intentionally?

dnechay (commented):
Wasn't able to check all the test files deeply, but overall looks good

Dropped some comments, will check test file deeper tomorrow

packages/apps/reputation-oracle/server/src/modules/escrow-completion-tracking/escrow-completion-tracking.service.ts
eugenvoronov: Yes, about 2 weeks ago @Dzeranov asked me to uncomment it. @Dzeranov is it correct?

Dzeranov: True. 
But I think you can comment sending webhook to the recording oracle back. I checked CVAT oracles src and found out that they're expecting this webhook only on Exchange Oracle.

eugenvoronov: Okay, thanks

packages/apps/reputation-oracle/server/src/modules/webhook/webhook-incoming.module.ts
eugenvoronov: Yes, you are right, we don't need to provide `EscrowCompletionTrackingModule` here.

dnechay (commented):
LGTM overall, last few comments

packages/apps/reputation-oracle/server/src/database/migrations/1731562977435-AddWebhookOutgoingAndEscrowCompletionTrackingTables.ts
```diff
@@ -46,7 +46,7 @@ export class AddWebhookOutgoingAndEscrowCompletionTrackingTables1731562977435
       `ALTER TYPE "hmt"."cron-jobs_cron_job_type_enum" RENAME TO "cron-jobs_cron_job_type_enum_old"`,
     );
     await queryRunner.query(
-      `CREATE TYPE "hmt"."cron-jobs_cron_job_type_enum" AS ENUM('process-pending-incoming-webhook', 'process-pending-outgoing-webhook', 'process-pending-escrow-completion-tracking', 'process-paid-escrow-completion-tracking')`,
+      `CREATE TYPE "hmt"."cron-jobs_cron_job_type_enum" AS ENUM('process-pending-incoming-webhook', 'process-pending-outgoing-webhook', 'process-pending-escrow-completion', 'process-paid-escrow-completion')`,
```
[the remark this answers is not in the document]
eugenvoronov: Yes, you are right, I didn't undo this changes, thanks

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.spec.ts
```diff
@@ -1,830 +1,396 @@
...
 
       await service.processPendingIncomingWebhooks();
 
-      expect(createEscrowCompletionTrackingMock).toHaveBeenCalled();
-      expect(updateOneMock).not.toHaveBeenCalledWith({
-        id: mockEntity.id,
-        status: EscrowCompletionTrackingStatus.COMPLETED,
-      });
-      expect(completeCronJobMock).toHaveBeenCalled();
+      expect(service.completeCronJob).toHaveBeenCalled();
```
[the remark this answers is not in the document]
eugenvoronov: I think will be better to rename it: `should complete the cron job after processing`

eugenvoronov:
Comments were resolved and unit tests were improved @dnechay

packages/apps/reputation-oracle/server/src/modules/webhook/webhook-outgoing.service.spec.ts
```diff
@@ -0,0 +1,292 @@
...
+    it('should successfully create outgoing webhook with valid DTO', async () => {
+      jest
+        .spyOn(webhookOutgoingRepository, 'createUnique')
+        .mockResolvedValue(webhookEntity as WebhookOutgoingEntity);
+
+      await webhookOutgoingService.createOutgoingWebhook(payload, url);
+
+      expect(webhookOutgoingRepository.createUnique).toHaveBeenCalledWith(
+        expect.objectContaining({
+          hash,
```
dnechay, line 134: Why checking only these two fields and not full object?

dnechay (dismissed):
LGTM

One non-blocking comment

dnechay (approved):

PR #2843, opened 2024-11-26
[HUM-74] hide not supported statuses for available tasks table filter

adrian-oleskiewicz:
## Issue tracking

HUM-74
## Context behind the change

Implemented filter for not supported statuses for available jobs table
HUM-81 - replaced SVG refresh icons to correct sizes

## Release plan

simple deploy


dnechay (commented):
Sorry for not clarifying it earlier: when on "available jobs" tab (no matter desktop or mobile) there is no need in `status` filter at all, this tab should always fetch only `active` jobs

Thanks for icons improvement!

adrian-oleskiewicz:
> Sorry for not clarifying it earlier: when on "available jobs" tab (no matter desktop or mobile) there is no need in `status` filter at all, this tab should always fetch only `active` jobs
> 
> Thanks for icons improvement!

Hey, now worries :) I removed this filter

dnechay (approved):

PR #2847, opened 2024-11-26
[Dashboard] fix: properly filter dashboard stats values

dnechay:
## Issue tracking
N/A

## Context behind the change
When getting values from cache it returns `undefined` when no value is there, so we have to properly check if the stat is missing.

Also updated docker-compose files for human-app and dashboard server:
- removed `version` because it's not necessary anymore
- added `name`, so when you have `redis` started in one of servers already, starting it in another one doesn't kill existing (they run on different ports anyways)

## How has this been tested?
- [x] on local: trigger stats endpoint to check it returns empty array when no data is there
- [x] e2e: check UI is able to display data with missing points

## Release plan
Merge

N/A

### DIFF
--- a/packages/apps/dashboard/server/docker-compose.yml
+++ b/packages/apps/dashboard/server/docker-compose.yml
@@ -1,4 +1,4 @@
-version: '3.8'
+name: 'dashboard-server'
 
 services:
   redis:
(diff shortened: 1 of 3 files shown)


Dzeranov (approved):
lgtm

Issue #2804, opened 2024-11-18
[HUMAN App] Duplicate oracle entries when Amoy and Sepolia networks are enabled


portuu3:
**Describe the bug**
If Amoy and Sepolia networks are enabled oracles active in two networks appears twice.

**Expected behavior**
It should appear only once

**Screenshots**
![Image](https://github.com/user-attachments/assets/0dc7269f-2236-45e7-bd67-4636d5ae86a3)

dnechay:
Taking into account that there might be oracles with same address, but different URLs in different chains, we decided to hold the development of this and discuss how we approach this from UX perspective. Once we have an agreement - will share it here

dnechay:
After discussing internally with the team we came up to next conclusions:
- we might want to introduce a bigger UI/UX redesign, where users will see just a list of available jobs (and that's a different story)
- we are not going to have something like described in this issue in production, so we can hold it a bit and decide later if we going to make a redesign or implement some sort of "network selector" to address this issue

PR #2861, opened 2024-11-28
[Human App] Added oracles filter

eugenvoronov:
## Issue tracking  
The progress of this issue is being tracked via internal bug reports related to duplicate oracles being displayed in the frontend when fetching data from the backend. [HUMAN App] Duplicate oracle entries when Amoy and Sepolia networks are enabled
[#2804](https://github.com/humanprotocol/human-protocol/issues/2804)

### Goal of the change:  
The goal of this pull request is to filter and display only unique oracles in the frontend based on a combination of their `url` and `address`. This prevents duplicate oracles from being rendered when the backend provides multiple entries with the same `url` and `address` but different `chainId` values.

### Code changes:  
1. **Introduction of a `Map` to filter unique oracles:**
   - A `Map` object is used to filter out duplicate oracles from the response based on a unique key combining `url` and `address`.
   - This ensures that only one oracle is displayed even if multiple entries with the same `url` and `address` are present in the response.

2. **Key generation logic:**
   - A unique key is generated for each oracle using the following pattern:  
     `const key = `${oracle.url}-${oracle.address}`;`
   - If the key already exists in the `Map`, the duplicate entry is ignored.

3. **Updated `getOracles` function:**  
   - The oracles are fetched from the backend and processed through the `Map` to ensure only unique entries are retained.  

### Steps taken:
**Manual testing:**
   - Loaded the application in a local development environment and verified that only unique oracles are displayed.
   - Tested different scenarios where:
     - Oracles with the same `url` but different `chainId` are returned.
     - Oracles with unique `url` and `address` combinations are returned.

### DIFF
--- a/packages/apps/human-app/frontend/src/api/services/worker/oracles.ts
+++ b/packages/apps/human-app/frontend/src/api/services/worker/oracles.ts
@@ -68,12 +68,20 @@ export async function getOracles({
       signal
     );
 
-    oracles = oracles.concat(
-      results.map((oracle) => ({
-        ...oracle,
-        name: oracleUrlToNameMap.get(oracle.url) ?? '',
-      }))
-    );
+    // Add new results and filter duplicates based on `url` and `address`
+    const uniqueOraclesMap = new Map<string, Oracle>();
+
+    results.forEach((oracle) => {
+      const key = `${oracle.url}-${oracle.address}`;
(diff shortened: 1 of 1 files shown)


portuu3:
I think it should be done in the backend

dnechay:
We will implement it on BE

PR #2867, opened 2024-12-02
[HDM-25] double footer scroll fix

adrian-oleskiewicz:
## Issue tracking

HDM-25: Please change table footer for mobile view so that there will be no second scroll bar. Additionally
 improved the display of: “No escrows launched yet”.
## Context behind the change

Simple css improvements to hide double scroll and improve UX for mobile
## Release plan

normal deploy


dnechay (approved):
Change lgtm
One question: can we get it as it mentioned in design? (not sure if possible via mui)
Screenshot 2024-12-02 at 17 31 58

When open on small screen (screenshot is just for example) - it has single scroll-bar which might be not intuitive to use
Screenshot 2024-12-02 at 17 34 01

If no - what we have is good enough :)

adrian-oleskiewicz:
@dnechay It was hard to achieve same pixel perfect design but I have very close solution
Zrzut ekranu 2024-12-3 o 09 57 19

dnechay (commented):
Nice work!

Probably we could use
```
position: 'relative',
right: '-20px',
```
remove gaps in `toolbar`, and `justify-self: center` on `displayedRows` to get something more similar

Taking your css as a base a provided some options to designer, let's see what they pick and then adjust a bit

Overall lgtm

adrian-oleskiewicz:
> Nice work!
> 
> Probably we could use
> 
> ```
> position: 'relative',
> right: '-20px',
> ```
> 
> on `selectLabel`, remove gaps in `toolbar`, and `justify-self: center` on `displayedRows` to get something more similar
> 
> Taking your css as a base a provided some options to designer, let's see what they pick and then adjust a bit
> 
> Overall lgtm

Thanks! That is really working nicely. 

Zrzut ekranu 2024-12-3 o 14 34 30

dnechay (approved):
🙌🏻

Issue #2851, opened 2024-11-26
[Dashboard server] Move `/stats/networks-available` to `/config/networks-available`


portuu3:
**Description**
Move `/stats/networks-available` to `/config/networks-available`

PR #2874, opened 2024-12-03
[Dashboard] Added config controller

eugenvoronov:
## Issue tracking  
The request for this change originated from a need to restructure configuration paths. This issue is tracked in [Dashboard server] Move /stats/networks-available to /config/networks-available
[#2851](https://github.com/humanprotocol/human-protocol/issues/2851).

## Context behind the change  
The `/stats/available-networks` endpoint has been moved to `/config/available-networks` to better reflect its purpose as part of the configuration layer rather than statistical data.  

### Changes made:  
- Updated the routing logic to point to `/config/available-networks`.  
- Modified any references to the old endpoint in the following files:  
  - `dashboard/server/src/modules/stats` ➡️ `dashboard/server/src/common/config`  
  - Added new controller `dashboard/server/src/common/config/config.controller.ts``

## How has this been tested?  
- **Manual Testing**:  
  Manually tested the application via the browser (Swagger)  and API tools (Postman) to confirm:  
  - `GET /config/available-networks`

### Pre-release:  
- Communicate endpoint change to relevant teams (frontend, API consumers).  
- Update internal and external API documentation.

### DIFF
--- a/packages/apps/dashboard/server/src/common/config/config.controller.ts
+++ b/packages/apps/dashboard/server/src/common/config/config.controller.ts
@@ -1,0 +1,26 @@
+import { ApiOperation, ApiResponse, ApiTags } from '@nestjs/swagger';
+import { Controller, Get, HttpCode } from '@nestjs/common';
+
+import { ChainId } from '@human-protocol/sdk';
+import { NetworkConfigService } from '../../common/config/network-config.service';
+
+@ApiTags('Config')
+@Controller('/config')
+export class ConfigController {
+  constructor(private readonly networkConfigService: NetworkConfigService) {}
+
+  @Get('/available-networks')
+  @HttpCode(200)
+  @ApiOperation({
+    summary: 'Get available networks with recent usage',
+    description: 'Endpoint to return networks filtered by recent activity.',
+  })
+  @ApiResponse({
+    status: 200,
+    description: 'Networks retrieved successfully',
+    type: Array<ChainId>,
+  })
+  public async getAvailableNetworks(): Promise<ChainId[]> {
+    return this.networkConfigService.getAvailableNetworks();
+  }
+}
--- a/packages/apps/dashboard/server/src/common/config/config.module.ts
+++ b/packages/apps/dashboard/server/src/common/config/config.module.ts
@@ -3,6 +3,7 @@ import { EnvironmentConfigService } from './env-config.service';
 import { RedisConfigService } from './redis-config.service';
 import { NetworkConfigService } from './network-config.service';
 import { S3ConfigService } from './s3-config.service';
+import { ConfigController } from './config.controller';
 
 @Global()
 @Module({
@@ -12,6 +13,7 @@ import { S3ConfigService } from './s3-config.service';
     NetworkConfigService,
     S3ConfigService,
   ],
+  controllers: [ConfigController],
   exports: [
     RedisConfigService,
     EnvironmentConfigService,
--- a/packages/apps/dashboard/server/src/modules/stats/stats.controller.ts
+++ b/packages/apps/dashboard/server/src/modules/stats/stats.controller.ts
@@ -10,16 +10,11 @@ import {
 import { HmtDailyStatsResponseDto } from './dto/hmt.dto';
 import { DateValidationPipe } from '../../common/pipes/date-validation.pipe';
 import { HmtGeneralStatsDto } from './dto/hmt-general-stats.dto';
-import { ChainId } from '@human-protocol/sdk';
-import { NetworkConfigService } from '../../common/config/network-config.service';
 
 @ApiTags('Stats')
 @Controller('/stats')
 export class StatsController {
-  constructor(
-    private readonly statsService: StatsService,
-    private readonly networkConfigService: NetworkConfigService,
-  ) {}
+  constructor(private readonly statsService: StatsService) {}
 
   @Get('/hmt-price')
   @HttpCode(200)
@@ -131,20 +126,5 @@ export class StatsController {
   ): Promise<HmtDailyStatsResponseDto> {
     const results = await this.statsService.hmtDailyStats(from, to);
     return { from, to, results };
-  }
-
-  @Get('/available-networks')
-  @HttpCode(200)
-  @ApiOperation({
-    summary: 'Get available networks with recent usage',
-    description: 'Endpoint to return networks filtered by recent activity.',
-  })
-  @ApiResponse({
-    status: 200,
-    description: 'Networks retrieved successfully',
-    type: Array<ChainId>,
-  })
-  public async getAvailableNetworks(): Promise<ChainId[]> {
-    return this.networkConfigService.getAvailableNetworks();
   }
 }


packages/apps/dashboard/server/src/modules/stats/stats.service.ts
```diff
@@ -178,8 +178,7 @@ export class StatsService implements OnModuleInit {
       totalHolders: 0,
       totalTransactions: 0,
     };
-    const availableNetworks =
-      await this.networkConfigService.getAvailableNetworks();
+    const availableNetworks = await this.networksService.getOperatingNetworks();
```
dnechay, line 181:
```suggestion
    const operatingNetworks = await this.networksService.getOperatingNetworks();
```
and below

packages/apps/dashboard/server/src/modules/networks/networks.service.ts
```diff
@@ -0,0 +1,90 @@
...
+export class NetworksService {
+  private readonly logger = new Logger(NetworksService.name);
+  constructor(
+    @Inject(CACHE_MANAGER) private cacheManager: Cache,
+    private readonly envConfigService: EnvironmentConfigService,
+  ) {}
+
+  public async getOperatingNetworks(): Promise<ChainId[]> {
+    const cachedNetworks = await this.cacheManager.get<ChainId[]>(
+      AVAILABLE_NETWORKS_CACHE_KEY,
```
dnechay, line 23:
```suggestion
      OPERATING_NETWORKS_CACHE_KEY,
```
And probably cache-key itself as well

packages/apps/dashboard/server/src/modules/networks/networks.service.ts
```diff
@@ -0,0 +1,90 @@
...
+        this.logger.error(
+          `Error processing network ${networkKey} (Chain ID: ${chainId}): ${error.message}`,
+        );
+      }
+    }
+
+    await this.cacheManager.set(
+      AVAILABLE_NETWORKS_CACHE_KEY,
+      availableNetworks,
+      this.envConfigService.networkAvailableCacheTtl,
```
dnechay, line 86: Let's rename `networkAvailableCacheTtl` and it's env counterpart to have `operating` word

packages/apps/dashboard/server/src/modules/networks/networks.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+import { NetworksService } from './networks.service';
+import { ChainId } from '@human-protocol/sdk';
+
+@ApiTags('Networks')
+@Controller('/networks')
+@UsePipes(new ValidationPipe({ transform: true }))
+export class NetworksController {
+  constructor(private readonly networksService: NetworksService) {}
+
+  @Get('/available-networks')
```
dnechay, line 18:
```suggestion
  @Get('/operating')
```
You already have `networks` in `@Controller`, so full path is `networks/available-networks` and looks a bit excessive

packages/apps/dashboard/server/src/modules/networks/networks.controller.ts
```diff
@@ -0,0 +1,32 @@
...
+  @HttpCode(200)
+  @ApiOperation({
+    summary: 'Get available networks with recent usage',
```
dnechay, line 21:
```suggestion
    summary: 'Get recently used networks',
```

dnechay (commented):
LGTM, some style comments

As a scout rule, could you please also fix `networkAvailableCacheTtl`? It should be in MS, but returns seconds, and also add `NETWORKS_AVAILABLE_CACHE_TTL` to `validationSchema`, so it's indeed `number` when we get it from env and not `string`

eugenvoronov:
Added unit tests

packages/apps/dashboard/server/src/common/config/env-config.service.ts
```diff
@@ -93,10 +93,10 @@ export class EnvironmentConfigService {
     );
   }
 
-  get networkAvailableCacheTtl(): number {
+  get networkOperatingCacheTtl(): number {
     return this.configService.get<number>(
```
dnechay, line 97: We should multiple the result by `1000` to return value in milliseconds as it's expected by `cacheManager`

Also need to add `NETWORKS_OPERATING_CACHE_TTL` to `validationSchema` when initializing config, so it's actually a `number`, not `string` when read from env

dnechay (changes requested):

dnechay:
Also please resolve conflicts and update PR name & description to correspond to latest state of the code

eugenvoronov:
@dnechay Added e2e tests

packages/apps/dashboard/server/src/modules/networks/networks.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    const availableNetworks = [];
+
+    for (const chainId of Object.values(MAINNET_CHAIN_IDS)) {
+      const networkConfig = NETWORKS[chainId];
+      if (!networkConfig) continue;
+
+      const statisticsClient = new StatisticsClient(networkConfig);
+      try {
+        const [hmtData, escrowStats] = await Promise.all([
+          statisticsClient.getHMTDailyData({ from: filterDate }),
```
dnechay, line 48: You have to keep
```
from: new Date(Math.floor(filterDate.getTime() / 1000) * 1000)
```
here, because under the hood for querying subgraph it uses seconds (and [does implicit conversion](https://github.com/humanprotocol/human-protocol/blob/32eda6b952c6cc70c4561cca32a8a17736768ba3/packages/sdk/typescript/human-protocol-sdk/src/statistics.ts#L509)), so if you don't `Math.floor` here - it fails with error like
```
Error processing network 56: An error occurred while interacting with Ethereum: Invalid value provided for argument `from`: Float(1731492900.033)
```

It's very confusing that we have such conversion here, but it's better to keep it as it was and fix SDK as separate PR

For ref: https://github.com/humanprotocol/human-protocol/issues/2912

packages/apps/dashboard/server/src/modules/networks/networks.service.ts
```diff
@@ -0,0 +1,79 @@
...
+          statisticsClient.getEscrowStatistics({ from: oneMonthAgo }),
```
dnechay, line 49: Same as above for `filterDate`

packages/apps/dashboard/server/test/app.e2e-spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        .get('/networks/operating')
+        .expect(200);
+
+      expect(Array.isArray(response.body)).toBe(true);
+      expect(response.body.every((id: any) => typeof id === 'number')).toBe(
+        true,
+      );
+    });
+
+    it('should retrieve cached operating networks on subsequent requests', async () => {
```
dnechay, line 53: You can remove this test in favor of next one: `should retrieve from cache when it exists`

packages/apps/dashboard/server/test/app.e2e-spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        expect.any(Array),
+        expect.any(Number),
+      );
+    });
+
+    it('should retrieve from cache when it exists', async () => {
+      const mockData = [1, 2, 3];
+      await cacheManager.set('operating-networks', mockData);
+
+      jest.spyOn(cacheManager, 'get');
```
dnechay, line 82: In e2e tests you don't need to spy, you only need to check it's actually working

packages/apps/dashboard/server/test/app.e2e-spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        .expect(200);
+
+      const secondResponse = await request(app.getHttpServer())
+        .get('/networks/operating')
+        .expect(200);
+
+      expect(secondResponse.body).toEqual(firstResponse.body);
+    });
+
+    it('should set the cache when it does not exist', async () => {
```
dnechay, line 65: You can remove this test, it's not e2e since it spies on cacheManager

dnechay (changes requested):
Sorry, in previous review I mentioned wrong test suite, it should have been "unit" instead of "e2e", but good to see it's added here as well! Let's clean it up a bit

Also could you please add `e2e.test` file to `.gitignore`? It's not really convenient to set it up locally every time

A1exKH (changes requested):
Code LGTM. Please address comments about tests.

eugenvoronov:
Thanks! Comments were resolved @A1exKH @dnechay

packages/apps/dashboard/server/src/modules/networks/networks.service.ts
```diff
@@ -0,0 +1,81 @@
...
+      const networkConfig = NETWORKS[chainId];
+      if (!networkConfig) continue;
+
+      const statisticsClient = new StatisticsClient(networkConfig);
+      try {
+        const [hmtData, escrowStats] = await Promise.all([
+          statisticsClient.getHMTDailyData({
+            from: new Date(Math.floor(filterDate.getTime() / 1000) * 1000),
+          }),
+          statisticsClient.getEscrowStatistics({ from: oneMonthAgo }),
```
dnechay, line 51: We need to round `from` for `getEscrowStatistics` as well

dnechay (commented):
@eugenvoronov you forgot to add `Math.floor` in the second place

packages/apps/dashboard/server/.gitignore
```diff
@@ -38,4 +38,6 @@ lerna-debug.log*
 /redis_data
 
 .env.development
-.env.production
\ No newline at end of file
+.env.production
+
+e2e.test
```
dnechay, line 43: The format of env file name is `.env${NODE_ENV}` and for tests that run via `jest` `NODE_ENV = test` ([ref](https://jestjs.io/docs/environment-variables#node_env)), so

I would suggest to do one of two:
1.
```
.env.test
```
if we want it just automatically

or
2.
```
.env.e2e-test
```
and in `package.json` change the command to be
```
test:e2e": " NODE_ENV=e2e-test jest --config ./test/jest-e2e.json
```

dnechay (commented):
LGTM

@eugenvoronov last comment about ignored env file for tests

ansaqib:
@A1exKH - You look to have got interest in the project - Request you to explain your interests

A1exKH (approved):
LGTM.

dnechay (approved):

Issue #2778, opened 2024-11-12
[Dashboard] Support leaderboard sorting by reputation and fee on BE


dnechay:
**Description**
Right now leaderboard table fetches pre-ordered data (sorted by stake amount) from BE, displays all the results and allows to sort them by different columns, but this sorting is done fully on UI, so it might not meet user expectations. Instead when users initiates sorting of the table by column other than "Stake" - we should make a request to BE to get leaders sorted by new criteria, not sort an existing list.

**Motivation**
In order to allow users to see leaders sorted by different criteria.

**Implementation details** 
TBD by developer

PR #2904, opened 2024-12-12
[Dashboard|SDK] Implemented leaders sorting

eugenvoronov:
## Issue tracking
This issue is being tracked in the following GitHub issue: [Dashboard] Support leaderboard sorting by reputation and fee on BE
[#2778](https://github.com/humanprotocol/human-protocol/issues/2778)

## Context behind the change
The changes in this pull request aim to improve how leaderboards are sorted in the application. Previously, sorting was done only on the client-side, leading to performance issues and mismatches with user expectations. Now, the sorting is handled server-side to ensure that the results are accurately sorted according to the selected criteria.

Key changes:
1. **Backend (Details controller and service)**:
   - Updated `/leaders` controller to accept sorting parameters (`orderBy`, `orderDirection`, `first`, and `skip`).
   - Implemented `getLeader` method in the Details Service to handle sorting by various criteria (e.g., fee, reputation, amountStaked, role).
   - Added pagination and reputation-based filtering in the service layer.
   - Modified the `assignReputationsToLeaders` method to ensure default reputation values are assigned when reputation data is missing.

2. **SDK updates**:
   - Extended the `LeaderFilter` in both Python and TypeScript SDKs to accept additional parameters such as `roles`, `minAmountStaked`, and `orderBy`. In order to effectively filter data at the subgraph level and request the data we need.
   - Updated the `GET_LEADERS_QUERY` GraphQL query to support these new parameters.
   - Modified the `getLeaders` method in `OperatorUtils` to handle the new filtering and sorting logic.

---

### Unit Tests:
- Implemented unit tests for the **Details Service** to cover scenarios such as:
  - Fetching and returning leaders with reputations.
  - Handling missing reputation data gracefully.
  - Sorting leaders by reputation.
  - Handling errors when fetching reputations.
  - Supporting pagination when sorting by reputation.

### Manual Testing:
- **Backend**:
  - Tested the `/leaders` endpoint to ensure it handles sorting by multiple criteria such as fee, reputation, amount staked, and role.
  - Verified pagination works as expected when sorting.
  - Checked default filtering logic for `minAmountStaked` and `roles`.

### Frontend
Implement sorting on Dashboard UI side.

### Suggestion
It would be good to unify the list of roles, since the spelling varies from service to service.

### Identified issue: Multi-chain sorting inaccuracy
During development, I noticed a sorting issue when querying leaders across multiple chain IDs. When retrieving data from multiple networks (e.g. getAvailableNetworks return more than one `chainId`), the sorting of the aggregated results becomes incorrect. This happens because sorting is performed separately within each subgraph on the SDK side, not across all queried chains simultaneously and together.

Since the subgraph only sorts data within its own chain-specific dataset, fetching results from multiple chain IDs involves sending separate requests. After receiving these sorted lists, we flatten (concatenate) them in the application, causing the global sorting to be inaccurate.

A possible solution is to implement back-end sorting after aggregating results from multiple chain IDs. However, this approach may be inefficient.


packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -148,40 +153,88 @@ export class DetailsService {
     return result;
   }
 
-  public async getLeadersByChainId(
+  public async getLeaders(
```
dnechay, line 156: Maybe for convenience we make pagination options as object?

```
type GetLeadersPaginationOptions = {
  orderBy?: LeadersOrderBy;
  orderDirection?: OrderDirection;
  first?: number;
  skip?: number,
}
...
public async getLeaders(chainId: ChainId, {
  orderBy,
  orderDirection,
  first,
  skip
}: GetLeadersPaginationOptions = {})
```
It's fine to have `chainId` as required argument because there is `ChainId.ALL` option we can explicitly use

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -148,40 +153,88 @@ export class DetailsService {
...
+          orderBy,
+          orderDirection,
+          first,
+          skip,
+        );
+        const [leaders, reputations] = await Promise.all([
+          OperatorUtils.getLeaders(filter),
+          this.fetchReputations(id),
+        ]);
+        return { leaders, reputations };
```
dnechay, line 180: Maybe we assign reputations to leaders here and operate only on `leaders` below?

packages/sdk/typescript/human-protocol-sdk/src/graphql/queries/operator.ts
```diff
@@ -26,22 +26,32 @@ const LEADER_FRAGMENT = gql`
...
 
 export const GET_LEADERS_QUERY = (filter: ILeadersFilter) => {
-  const { role } = filter;
+  const { role, roles, minAmountStaked } = filter;
 
   const WHERE_CLAUSE = `
     where: {
-      ${role ? `role: $role` : ''}
+      ${minAmountStaked ? `amountStaked_gt: $minAmountStaked` : ''}
+      ${roles ? `role_in: $roles` : ''}
```
dnechay, line 34: Should it be one or another for `role`/`roles`?

Ideally would be nice to keep just `roles`, so when you need one - you provide a single value. However it's a breaking change and we should not just remove `role` if we expect somebody else might be using it, but IMU:
- we are probably only consumers of this particular API
- it's used only on dashboard

So maybe we can safely remove `role` as for now, wdyt?

packages/sdk/typescript/human-protocol-sdk/src/graphql/queries/operator.ts
```diff
@@ -26,22 +26,32 @@ const LEADER_FRAGMENT = gql`
 `;
 
 export const GET_LEADERS_QUERY = (filter: ILeadersFilter) => {
-  const { role } = filter;
+  const { role, roles, minAmountStaked } = filter;
 
   const WHERE_CLAUSE = `
     where: {
-      ${role ? `role: $role` : ''}
+      ${minAmountStaked ? `amountStaked_gt: $minAmountStaked` : ''}
```
dnechay, line 33: Should be including end?
```suggestion
      ${minAmountStaked ? `amountStaked_gte: $minAmountStaked` : ''}
```

packages/sdk/typescript/human-protocol-sdk/src/operator.ts
```diff
@@ -108,6 +108,11 @@ export class OperatorUtils {
   public static async getLeaders(filter: ILeadersFilter): Promise<ILeader[]> {
     let leaders_data: ILeader[] = [];
 
+    const first =
+      filter.first !== undefined ? Math.min(filter.first, 1000) : 10;
```
dnechay, line 112: Do we need some checks for negative values?

packages/sdk/typescript/human-protocol-sdk/test/operator.test.ts
```diff
@@ -211,15 +212,27 @@ describe('OperatorUtils', () => {
       const gqlFetchSpy = vi.spyOn(gqlFetch, 'default').mockResolvedValueOnce({
         leaders: [mockLeaderSubgraph, mockLeaderSubgraph],
       });
-      const filter = { chainId: ChainId.LOCALHOST, role: 'role' };
+      const filter: ILeadersFilter = {
+        chainId: ChainId.LOCALHOST,
+        role: 'role',
+      };
+      const first =
```
dnechay, line 219: Are these three lines useless? There are not `first`, `skip` and `orderDirection` in the object above

packages/sdk/typescript/human-protocol-sdk/test/operator.test.ts
```diff
@@ -237,17 +250,30 @@ describe('OperatorUtils', () => {
...
```
dnechay, line 257: Same here: do we need these calculated based on `filter.<param>`?

packages/sdk/typescript/human-protocol-sdk/test/operator.test.ts
```diff
@@ -263,15 +289,28 @@ describe('OperatorUtils', () => {
       const gqlFetchSpy = vi.spyOn(gqlFetch, 'default').mockResolvedValueOnce({
         leaders: [mockLeaderSubgraph, mockLeaderSubgraph],
       });
-      const filter = { chainId: ChainId.LOCALHOST, role: 'role' };
+
+      const filter: ILeadersFilter = {
+        chainId: ChainId.LOCALHOST,
+        role: 'role',
+      };
+      const first =
```
dnechay, line 297: Same here: do we need it based on `filter`?

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/gql/operator.py
```diff
@@ -31,12 +31,24 @@
 def get_leaders_query(filter: LeaderFilter):
     return """
 query GetLeaders(
-    $role: String
+    $role: String,
```
dnechay, line 34: Same not about having both `role/roles` as for TS sdk

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/gql/operator.py
```diff
@@ -45,6 +57,10 @@ def get_leaders_query(filter: LeaderFilter):
 """.format(
         leader_fragment=leader_fragment,
         role_clause="role: $role" if filter.role else "",
+        amount_staked_clause=(
+            "amountStaked_gt: $minAmountStaked" if filter.min_amount_staked else ""
```
dnechay, line 61:
```suggestion
            "amountStaked_gte: $minAmountStaked" if filter.min_amount_staked else ""
```

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/operator/operator_utils.py
```diff
@@ -44,19 +44,45 @@ class LeaderFilter:
...
+
         self.chain_id = chain_id
         self.role = role
+        self.role = role
+        self.roles = roles
+        self.min_amount_staked = min_amount_staked
+        self.order_by = order_by
+        self.order_direction = order_direction
+        self.first = min(first, 1000)
+        self.skip = skip
```
dnechay, line 85: Do we need some checks on negative values?

packages/sdk/python/human-protocol-sdk/test/human_protocol_sdk/operator/test_operator_utils.py
```diff
@@ -116,10 +124,32 @@ def test_get_leaders_when_job_types_is_none(self):
 
             leaders = OperatorUtils.get_leaders(filter)
 
+            print("Query used:", get_leaders_query(filter))
+            print(
```
dnechay, line 128: Do we need these `print`s?

packages/apps/dashboard/server/src/modules/details/dto/details-pagination.dto.ts
```diff
@@ -9,9 +9,59 @@ import {
...
+    enum: OrderDirection,
+    default: OrderDirection.DESC,
+  })
+  @IsEnum(OrderDirection)
+  @IsIn(Object.values(OrderDirection))
+  @IsOptional()
+  public orderDirection?: OrderDirection;
+
+  @ApiPropertyOptional({
+    minimum: 0,
```
dnechay, line 44:
```suggestion
    minimum: 1,
```
?

packages/apps/dashboard/server/src/modules/details/dto/details-pagination.dto.ts
```diff
@@ -9,9 +9,59 @@ import {
...
+    default: 10,
+  })
+  @Type(() => Number)
+  @IsNumber()
+  @Min(0)
```
dnechay, line 49:
```suggestion
  @Min(1)
```

packages/apps/dashboard/server/src/modules/details/dto/details-pagination.dto.ts
```diff
@@ -9,9 +9,59 @@ import {
...
+  public first?: number = 10;
+
+  @ApiPropertyOptional({
+    minimum: 0,
+    default: 0,
+  })
+  @Type(() => Number)
+  @IsNumber()
+  @Min(0)
+  @Max(1000)
```
dnechay, line 61: Maybe let's make a comment that we do not expect more than that? It looks a bit weird that we are limiting `skip` :)

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -148,40 +153,88 @@ export class DetailsService {
...
-    }
 
-    allLeadersData.sort((a, b) =>
-      BigInt(a.amountStaked) >= BigInt(b.amountStaked) ? -1 : 1,
-    );
-    if (take && take > 0) {
-      allLeadersData = allLeadersData.slice(0, take);
+    this.assignReputationsToLeaders(leaders, reputations);
+
+    if (orderBy === LeadersOrderBy.REPUTATION) {
```
dnechay, line 195: In case if we need to get list of leaders sorted by reputation we need to:
- first: get all reputations sorted asc/desc
- take the page of N addresses from sorted reputations
- retrieve leaders details for these N addresses
- return the results

Retrieving leaders by addresses can be done in 3 ways:
1. by using `OperatorUtils.getLeader` for each address
2. by adding `addresses` filter to `OperatorUtils.getLeaders`
3. implementing new util like `OperatorUtils.getLeadersByIds`

The first option is not effective and it might be expensive to send many queries to subgraph.
The second one seems a bit dirty because it becomes a "super-method", so I would do it as a 3rd one, where we can also limit the number of `ids` to get to some reasonable value (e.g. based on GQL max query size) and that might be less than max value of `1000` allowed for page

As an alternative and as option to simplify this PR - we can remove the "feature" of sorting by "reputation"

wdyt, @portuu3 @eugenvoronov ?

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -148,40 +153,88 @@ export class DetailsService {
...
+            : rankB - rankA;
+        })
+        .slice(skip ?? 0, (skip ?? 0) + (first ?? leaders.length));
     }
-    const leaders = allLeadersData.map((leader) =>
-      plainToInstance(LeaderDto, leader, {
-        excludeExtraneousValues: true,
-      }),
-    );
+    return leaders;
```
dnechay, line 206: If `orderBy` is one of the leader fields (i.e. not "reputation"), then when querying leaders for all chains we need to:
1. Get N (as requested page size) from each chain
2. Merge all results to a single array
3. Sort this array per order criteria
4. Slice it to N results

This way you will address they issue of the order when retrieving data from multiple chains.

As an alternative, I can suggest to remove the option to get leaders across all chains and make `chainId` required parameter (that shouldn't be `ALL`/`-1`). I don't see a scenario for the user when they might want to see "leaders" across all chains

wdyt, @portuu3 ?

dnechay (commented):
Few extra questions:

1. Are these `md` files updates automatic? Could you please shed some light on how/when we run it?
2. What about a UI counterpart? This PR focuses on BE side, are we going to implement actual logic to fetch values using "sort criteria" selected by user in a different one?

eugenvoronov:
> 2\. What about a UI counterpart? This PR focuses on BE side, are we going to implement actual logic to fetch values using "sort criteria" selected by user in a different one?

1. For python SDK:
```
cd /human-protocol/packages/sdk/python/human-protocol-sdk
make generate-doc-md
```

For typescript SDK:
```
cd /human-protocol/packages/sdk/typescript/human-protocol-sdk
yarn build:doc
```

2. Yes, as I mentioned in PR description, we need to link the filtering functions in the dashboard to the server part. But for now, the changes are fully compatible with the current version of the client.

packages/sdk/typescript/human-protocol-sdk/src/interfaces.ts
```diff
@@ -35,9 +35,12 @@ export interface ILeaderSubgraph
   reputationNetworks?: { address: string }[];
 }
 
-export interface ILeadersFilter {
+export interface ILeadersFilter extends IPagination {
   chainId: ChainId;
   role?: string;
```
portuu3, line 40: lets remove role

packages/sdk/typescript/human-protocol-sdk/src/graphql/queries/operator.ts
portuu3: Yes we can do it

packages/apps/dashboard/server/src/modules/details/details.service.ts
portuu3: lgtm

packages/apps/dashboard/server/src/modules/details/details.service.ts
portuu3: It is a good point, but I think we need reputation ordering.
In any case max items in GQL is 1000, I don't think we will have 1000 operators at least for now, so it could be addresses in one request to subgraph and other to reputation oracle and cross the data.

packages/apps/dashboard/server/src/modules/details/details.service.ts
eugenvoronov: @dnechay I agree to include a pagination object)

Could you clarify what you mean by the ChainId.ALL option? Using ChainId.ALL at the SDK level results in an error. I only see this option:
```
const chainIds = chainId === ChainId.ALL
    ? MAINNET_CHAIN_IDS
    : chainId
    ? [chainId]
    : await this.networkConfig.getAvailableNetworks();
```

packages/apps/dashboard/server/src/modules/details/dto/details-pagination.dto.ts
eugenvoronov: Yes, you are right, I am adding this restriction by mistake.

packages/sdk/typescript/human-protocol-sdk/src/graphql/queries/operator.ts
eugenvoronov: Yes, agree

packages/apps/dashboard/server/src/modules/details/details.service.ts
eugenvoronov: @dnechay Before creating this PR, I attempted to implement option 2. However, I found that when requesting a list of leaders by addresses from the subgraph in a specific order, the subgraph does not preserve the order and disrupts it.

Copy of the message to Adrian: 
> I tested a slightly more optimized approach for filtering leaders. The idea was to first request data filtered by reputation points from the Reputation Oracle, and then query the SDK for the leaders by their addresses, fetching only the necessary ones in the already defined order. Unfortunately, when selecting data from the subgraph using addresses, the order wasn’t preserved (because subgraph didn't save the order of records), so this approach ended up being ineffective.
I’ve reverted to the original plan: retrieving the full list of leaders first from subgraph and then sorting them on the dashboard side.
[4:07](https://humanprotocol.slack.com/archives/D04AE6252QL/p1733922466909459)
This is how I wanted to get them from subgraph, i've passed the list of addresses of leaders received from Reputation Oracle.
export const GET_LEADERS_QUERY = (filter: ILeadersFilter) => {
  const { role,, addresses } = filter;

  const WHERE_CLAUSE = `
    where: {
      ${addresses ? `address_in: $addresses` : ''}
      ${role ? `role: $role` : ''}
    }
  `;

  return gql`
    query getLeaders(
      $addresses: [String]
      $role: String
      $first: Int
      $skip: Int
      $orderBy: String
      $orderDirection: String
    ) {
      leaders(
        ${WHERE_CLAUSE}
        first: $first
        skip: $skip
        orderBy: $orderBy
        orderDirection: $orderDirection
      ) {
        ...LeaderFields
      }
    }
    ${LEADER_FRAGMENT}
  `;
};

eugenvoronov: This may be the best and easiest way we can do this:

> It is a good point, but I think we need reputation ordering. In any case max items in GQL is 1000, I don't think we will have 1000 operators at least for now, so it could be addresses in one request to subgraph and other to reputation oracle and cross the data. @dnechay

packages/apps/dashboard/server/src/modules/details/details.service.ts
dnechay: @eugenvoronov 
Screenshot 2024-12-16 at 15 05 42

But from [this discussion](https://github.com/humanprotocol/human-protocol/pull/2904#discussion_r1881926672) my understanding is that no need to allow it, only specific chain to make leaders retrieval simpler

eugenvoronov:
@dnechay @portuu3 comments were resolved, ready for review 

I want to point out that when users choose to sort leaders by reputation, leaders with a null reputation are currently not displayed. I haven't added logic in the `assignReputationsToLeaders` method to include these leaders in the final list yet in order not to complicate this method.. If this behavior is incorrect, please let me know, and I will adjust the logic accordingly.

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -148,40 +153,51 @@ export class DetailsService {
...
+    );
 
-    let allLeadersData: ILeader[] = [];
-    for (const id of chainIds) {
-      const leadersData = await OperatorUtils.getLeaders({ chainId: id });
-      allLeadersData = allLeadersData.concat(
-        leadersData.filter((leader) => leader.amountStaked > 0 && leader.role),
-      );
-    }
+    const [rawLeaders, reputations] = await Promise.all([
```
portuu3, line 167: I think this is not going to work. Lets say we want to get 4 leaders ordered by reputation, then:
- OperatorUtils.getLeaders(filter) returns leaders A, B, C, D.
- fetchReputations return B, C, E, F.
Then we would be only returning B and C, when we should be getting B,C,E,F

portuu3 (changes requested):

packages/apps/reputation-oracle/server/src/modules/reputation/reputation.repository.ts
```diff
@@ -32,23 +36,41 @@ export class ReputationRepository extends BaseRepository<ReputationEntity> {
...
+      ...(first && { take: first }),
     });
   }
+
+  private static mapOrderBy(orderBy: ReputationOrderBy): string {
+    const orderByMap = {
+      [ReputationOrderBy.CREATED_AT]: 'createdAt',
+      [ReputationOrderBy.REPUTATION_POINTS]: 'reputationPoints',
+    };
+    return orderByMap[orderBy] || 'createdAt';
```
dnechay, line 74:
```suggestion
    return orderByMap[orderBy] || orderByMap[ReputationOrderBy.CREATED_AT];
```

packages/apps/reputation-oracle/server/src/modules/reputation/reputation.dto.ts
```diff
@@ -53,6 +57,34 @@ export class ReputationGetAllQueryDto {
...
+    default: SortDirection.DESC,
+  })
+  @IsEnum(SortDirection)
+  @IsIn(Object.values(SortDirection))
+  @IsOptional()
+  public orderDirection?: SortDirection;
+
+  @ApiPropertyOptional({ type: Number })
+  @IsOptional()
+  @Transform(({ value }) => Number(value))
```
dnechay, line 81: Should `first` and `skip` have `Min(N)` for validation purpose?

dnechay (changes requested):
Please pay attention to [Adrian's comment above](https://github.com/humanprotocol/human-protocol/pull/2904/files?show-viewed-files=true&file-filters%5B%5D=#r1888505834), for that case solution doesn't work.

In case we order by leader property - we need to get N leaders and then retrieve reputations for them (aka all).
In case we order by reputation - we need to get N reputations and then get corresponding leaders

packages/apps/reputation-oracle/server/src/modules/reputation/reputation.dto.ts
eugenvoronov: Yes, agree

eugenvoronov:
```
@ApiPropertyOptional({ type: Number })
  @IsOptional()
  @Min(1)
  @Transform(({ value }) => Number(value))
  public first?: number;

  @ApiPropertyOptional({ type: Number })
  @IsOptional()
  @Min(0)
  @Transform(({ value }) => Number(value))
  public skip?: number;
```

eugenvoronov: What do you think? @dnechay

eugenvoronov:
> Please pay attention to [Adrian's comment above](https://github.com/humanprotocol/human-protocol/pull/2904/files?show-viewed-files=true&file-filters%5B%5D=#r1888505834), for that case solution doesn't work.
> 
> In case we order by leader property - we need to get N leaders and then retrieve reputations for them (aka all). In case we order by reputation - we need to get N reputations and then get corresponding leaders

In case `order by reputation` there may be a case where there is no leader that matches the reputation. In our last discussion, Adrian suggested:

Sorting by `role`, `amountStaked`, `fee`:
1. Fetch required amount of leaders sorted by (role, amountStaked, fee).
2. Fetch all reputations for oracle roles.
3. Cross the data.
4. Return the list

Sorting by `reputation`:
1. Fetch all leaders.
2. Fetch all reputations for oracle roles.
3. Cross the data.
4. Order them based on desired order.
@dnechay what do you think?
Return only required amount of elements.

packages/sdk/typescript/human-protocol-sdk/test/operator.test.ts
```diff
@@ -14,14 +14,15 @@ import {
 } from '../src/graphql/queries/operator';
```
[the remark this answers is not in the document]
eugenvoronov: I will add it now

dnechay:
@eugenvoronov 

> Sorting by role, amountStaked, fee:

That part looks good (later we might want to get reputations for specific addresses in case there are too many returned)

> In case order by reputation there may be a case where there is no leader that matches the reputation

Could you please provide an example of this? I'm curious how this is possible
But if we expect it to happen, then:
1. Fetch all reputations for oracle roles ordered by reputation (you already to it by `reputationPoints: desc`)
2. Fetch max N leaders (1000 in our case)
3. Cross the data
4. Return only required amount of elements

eugenvoronov:
> @eugenvoronov
> 
> > Sorting by role, amountStaked, fee:
> 
> That part looks good (later we might want to get reputations for specific addresses in case there are too many returned)
> 
> > In case order by reputation there may be a case where there is no leader that matches the reputation
> 
> Could you please provide an example of this? I'm curious how this is possible But if we expect it to happen, then:
> 
> 1. Fetch all reputations for oracle roles ordered by reputation (you already to it by `reputationPoints: desc`)
> 2. Fetch max N leaders (1000 in our case)
> 3. Cross the data
> 4. Return only required amount of elements

I think this is only possible if not all leaders in the subgraph are indexed.

eugenvoronov:
@dnechay 
I improved the logic for the `reputation` sorting case, following the description:
> Fetch all reputations for oracle roles ordered by reputation (you already to it by reputationPoints: desc)
Fetch max N leaders (1000 in our case)
Cross the data
Return only required amount of elements


In the case described by Adrian https://github.com/humanprotocol/human-protocol/pull/2904#discussion_r1888505834, if there is no leader to whom the reputation belongs, then the leader will not be returned (please, check this test case `should return sorted leaders by reputation`).

dnechay:
@eugenvoronov 

> In the case described by Adrian https://github.com/humanprotocol/human-protocol/pull/2904#discussion_r1888505834, if there is no leader to whom the reputation belongs, then the leader will not be returned

I see now, I thought you were talking about different scenario, where there might be a "reputation object with address", but "leader entity" doesn't exist at all (i.e. in the chain, not just in response), but now I see that we talk about responses in `getLeaders`.

To avoid that we should implement the API where we can get leaders by array of addresses (right now we can do it only one by one), so the flow will be:
1. You will get N reputation entities already sorted
2. You get N addresses from reputation entities response
3. You take these N addresses and fetch leaders details
4. You merge the data

But as discussed above - we don't need to implement it now as since we do not expect to have more than 1000 leaders, so now we can get this 1000 and use it for both cases. Later if there are more than 1000 - we will fix that. That's my understanding of what we agreed on

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -149,40 +155,62 @@ export class DetailsService {
...
+      first,
+    );
+
+    const [rawLeaders, reputations] = await Promise.all([
+      OperatorUtils.getLeaders(filter),
+      this.fetchReputations(
+        chainId,
+        orderBy,
+        orderDirection,
+        orderBy !== LeadersOrderBy.REPUTATION && first,
```
dnechay, line 175: Probably it should be vice-versa:
```suggestion
        orderBy === LeadersOrderBy.REPUTATION ? first : undefined,
```

so:
- when `orderBy` **_is "by reputation"_**, you get first `N` and then cross them with `getLeaders` response
- when `orderBy` **_is  any leader prop_** (i.e. not "by reputation") - we take all reputations to cross the with N leaders

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -149,40 +155,62 @@ export class DetailsService {
...
-      const reputations = await this.fetchReputations(id);
-      this.assignReputationsToLeaders(leaders, reputations, id);
-    }
+  private createLeadersFilter(
+    chainId: ChainId,
+    orderBy: LeadersOrderBy,
+    orderDirection: OrderDirection,
+    first?: number,
+  ): ILeadersFilter {
+    const isReputationOrder = orderBy === LeadersOrderBy.REPUTATION;
```
dnechay, line 199: Style hint - we could refactor it a bit to improve readability and get rid of many ternaries:
```
const leadersFilter: ILeadersFilter = {
  chainId,
  minAmountStaked: MIN_AMOUNT_STAKED,
  roles: [
    SubgraphOracleRole.JOB_LAUNCHER,
    SubgraphOracleRole.EXCHANGE_ORACLE,
    SubgraphOracleRole.RECORDING_ORACLE,
    SubgraphOracleRole.REPUTATION_ORACLE,
  ],
};

if (orderBy === LeadersOrderBy.REPUTATION) {
  leadersFilter.first = MAX_LEADERS_COUNT;
} else {
  Object.assign(leadersFilter, {
    orderBy,
    orderDirection,
    first,
  });
}

return leadersFilter;
```

Also it will fix `first` value for the case when it's `LeadersOrderBy.REPUTATION` because now if we pass `first = 10` - then we can get to issue with missing leaders, we should get `MAX_LEADERS_COUNT` in this case

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -240,19 +283,34 @@ export class DetailsService {
...
+    orderBy: LeadersOrderBy,
+    first: number,
+  ): LeaderDto[] {
+    const leaderMap = new Map(leaders.map((l) => [l.address.toLowerCase(), l]));
+
+    if (orderBy === LeadersOrderBy.REPUTATION) {
+      return reputations
+        .map((rep) => {
+          const leader = leaderMap.get(rep.address.toLowerCase());
+          if (leader) {
```
dnechay, line 295: @portuu3 @eugenvoronov can you think about such case? IMU it can't be that we have a reputation record but don't have leader address in chain. If it's not returned by subgraph then it's probably an issue and in this case I would:
- write a log that we don't have leader for reputation record
- skip this leader for response
- do not substitute this "missing leader" with another one

i.e. in code it will look like
```
reputations.slice(0, first)
  .map(/*existing logic*/)
  .filter(Boolean)
```

In any case - would be nice to drop a comment in the code about what & why so somebody new doesn't try to break it w/o full context

packages/apps/dashboard/server/src/modules/details/details.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    expect(result).toEqual([
+      expect.objectContaining({
+        address: '0x456',
+        role: 'Job Launcher',
+        reputation: 'low',
+      }),
+    ]);
+  });
+
+  it('should return sorted leaders by reputation', async () => {
```
dnechay, line 107: This case covered in the next one - `should return first 3 sorted leaders by reputation` - so can be removed

packages/apps/dashboard/server/src/modules/details/details.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+  OperatorUtils: {
+    getLeaders: jest.fn(),
+  },
+}));
+
+const MOCK_MAX_LEADERS_COUNT = 5;
+
+jest.mock('../../common/constants/leader', () => ({
+  ...jest.requireActual('../../common/constants/leader'),
+  MAX_LEADERS_COUNT: MOCK_MAX_LEADERS_COUNT,
```
dnechay, line 27: By some reason we don't run tests for dashboard in CI 🤔 
This mock fails
Screenshot 2024-12-18 at 17 32 58

Probably because of hoisting of `jest.mock` ([read this for ref](https://dev.to/jobber/serious-jest-making-sense-of-hoisting-253i))

There are some workarounds for it like e.g. using `var` instead of `const` and etc.

packages/apps/dashboard/server/src/modules/details/details.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    const result = await service.getLeaders(ChainId.POLYGON_AMOY, {
+      orderBy: LeadersOrderBy.REPUTATION,
+      orderDirection: OrderDirection.DESC,
+    });
+
+    expect(result[0].address).toBe('0xB');
+    expect(result[1].address).toBe('0xC');
+  });
+
+  it('should return first 3 sorted leaders by reputation', async () => {
```
dnechay, line 135: This test case fails

packages/apps/dashboard/server/src/modules/details/details.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    const mockReputations = [
+      { address: '0xB', reputation: 'hign' },
+      { address: '0xC', reputation: 'hign' },
+      { address: '0xD', reputation: 'medium' },
+      { address: '0xF', reputation: 'medium' },
+      { address: '0xE', reputation: 'low' },
+    ];
+
+    const getLeadersSpy = jest.spyOn(OperatorUtils, 'getLeaders');
+    getLeadersSpy.mockResolvedValue(mockLeaders as ILeader[]);
```
dnechay, line 152: Let's use
```
getLeadersSpy.mockImplementation(async ({ first }) => {
      return mockLeaders.slice(0, first) as ILeader[];
    });
```

because it helps to disclose issue with invalid filter passed

dnechay (commented):
Could you please also add dashboard tests to CI?

packages/apps/dashboard/server/src/modules/details/details.service.ts
eugenvoronov: You know, since we have `orderBy === LeadersOrderBy.REPUTATION ? first : undefined,`  in `this.fetchReputations`, I don't think we really need that `.slice(0, first)`.

eugenvoronov: You know, since we have `orderBy === LeadersOrderBy.REPUTATION ? first : undefined`, in `this.fetchReputations`, I don't think we really need that `.slice(0, first)`.

dnechay: Good point! ☝🏻

dnechay (approved):
LGTM

dnechay (dismissed):

packages/apps/dashboard/server/src/common/constants/reputation.ts
```diff
@@ -0,0 +1,7 @@
+import { ReputationLevel } from '../enums/reputation';
+
+export const REPUTATION_RANK: Record<ReputationLevel, number> = {
```
portuu3, line 3: can't find a reference to this

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -237,22 +282,56 @@ export class DetailsService {
...
+   *   - Create a map of reputation records for quick lookup.
+   *   - Update each leader's reputation if found, otherwise set it to LOW.
+   * - Return the list of updated leaders.
+   */
   private assignReputationsToLeaders(
     leaders: LeaderDto[],
     reputations: { address: string; reputation: string }[],
-    chainId: ChainId,
-  ) {
+    orderBy: LeadersOrderBy,
```
portuu3, line 306: I think ordering shouldn't be here. As the method name says, here we should just map leaders and their reputations.

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -237,22 +282,56 @@ export class DetailsService {
...
+  ): LeaderDto[] {
+    const leaderMap = new Map(leaders.map((l) => [l.address.toLowerCase(), l]));
+
+    if (orderBy === LeadersOrderBy.REPUTATION) {
+      return reputations
+        .map((rep) => {
+          const leader = leaderMap.get(rep.address.toLowerCase());
+          if (!leader) {
+            this.logger.warn(
```
portuu3, line 315: We should not be logging this. This is something that might happen if an operator stops running a service and unstakers, removes the role, etc.

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -237,22 +282,56 @@ export class DetailsService {
...
+   *   - If no leader is found:
+   *     - Log a warning indicating a missing leader for the specific reputation record.
+   *     - This likely indicates an issue where the subgraph did not return the expected leader data.
+   *     - Skip processing the record without substituting a different leader.
+   *   - Return the top `first` leaders based on reputation.
+   * - If not ordered by REPUTATION:
+   *   - Create a map of reputation records for quick lookup.
+   *   - Update each leader's reputation if found, otherwise set it to LOW.
+   * - Return the list of updated leaders.
+   */
```
portuu3, line 302: I think we don't need the comment

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -237,22 +282,56 @@ export class DetailsService {
...
   private assignReputationsToLeaders(
```
portuu3, line 303: The reputation mapping should be done by chainId

portuu3 (changes requested):

packages/apps/dashboard/server/src/modules/details/details.service.ts
eugenvoronov: Why should this be done?

portuu3: Because an operator might have different reputation score on each network

packages/apps/dashboard/server/src/modules/details/details.service.ts
```diff
@@ -149,40 +155,64 @@ export class DetailsService {
...
+      chainId,
+      orderBy,
+      orderDirection,
+      first,
+    );
+
+    const [rawLeaders, reputations] = await Promise.all([
+      OperatorUtils.getLeaders(filter),
+      this.fetchReputations(
+        chainId,
```
eugenvoronov, line 172: Get reputation data for particular chain id

packages/apps/dashboard/server/src/modules/details/details.service.ts
eugenvoronov: Yes, but to `assignReputationsToLeaders` method we pass `leaders` retrieved from particular `chainId`, and same for `reputation`. https://github.com/humanprotocol/human-protocol/pull/2904/files#r1891605978

eugenvoronov: And `LeaderDto` also contain `chainId` param.

portuu3: You are right, we agreed on removing all chain ids fetch. Sorry

packages/apps/dashboard/server/src/common/constants/reputation.ts
eugenvoronov: Yes this can be removed

packages/apps/dashboard/server/src/modules/details/dto/details-pagination.dto.ts
```diff
@@ -9,9 +9,47 @@ import {
...
-import { ChainId } from '@human-protocol/sdk';
+import { ChainId, OrderDirection } from '@human-protocol/sdk';
 
 import { IsRoleValid } from './validation/role-validation';
+import { LeadersOrderBy } from 'src/common/enums/leader';
+
+export class LeadersPaginationDto {
+  @ApiProperty({ enum: ChainId })
+  @IsEnum(ChainId)
+  @IsIn(Object.values(ChainId).filter((id) => id !== ChainId.ALL))
```
dnechay, line 20: Please remove the `All Networks` option on UI in scope of this PR and as a default one make the first in the list of available options, otherwise this PR will break leaders table

dnechay:
@eugenvoronov also please resolve merge conflicts

packages/apps/dashboard/server/src/modules/details/dto/details-pagination.dto.ts
eugenvoronov: Done

dnechay (dismissed):

dnechay (approved):
@portuu3 tested locally, seems working as expected

portuu3 (approved):

PR #2908, opened 2024-12-12
HDM-33: added EVM  validation for address search

adrian-oleskiewicz:
## Issue tracking

HDM-33
## Context behind the change

I created helper function to test if address starts with 0x and has 40 characters and implemented check logic for search input. Button is blocked until correct address is provided and there is tooltip added. 
## How has this been tested?

manually
## Release plan

normal deply


dnechay (commented):
LGTM overall

One question before approve: can we get that tooltip visible also on mobile devices?

adrian-oleskiewicz:
> LGTM overall
> 
> One question before approve: can we get the tooltip visible also on mobile devices?

I implemented custom tooltip component so we can reuse it across the app

dnechay (approved):

Issue #2931, opened 2024-12-17
[Job Launcher] Check that the user has a credit card during job creation


portuu3:
**Description**
Check that the user has a credit card during job creation, so we can slash him in case of abuse.
This check does not apply to whitelisted users.

**Motivation**
Improve security

PR #2934, opened 2024-12-17
[Job Launcher] Check that the user has a credit card during job creation

flopez7:
## Issue tracking
#2931

## Context behind the change
Check that the user has a credit card during job creation, so we can slash him in case of abuse.
This check does not apply to whitelisted users

## How has this been tested?
Deployed locally and launched tests

## Release plan
None

None

### DIFF
--- a/packages/apps/job-launcher/server/src/modules/job/job.service.spec.ts
+++ b/packages/apps/job-launcher/server/src/modules/job/job.service.spec.ts
@@ -117,6 +117,7 @@ import { CronJobRepository } from '../cron-job/cron-job.repository';
 import { CronJobType } from '../../common/enums/cron-job';
 import { QualificationService } from '../qualification/qualification.service';
 import { NetworkConfigService } from '../../common/config/network-config.service';
+import { WhitelistService } from '../whitelist/whitelist.service';
 
 const rate = 1.5;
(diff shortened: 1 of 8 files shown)


packages/apps/job-launcher/server/src/modules/job/job.service.spec.ts
```diff
@@ -601,11 +620,26 @@ describe('JobService', () => {
       getUserBalanceMock.mockResolvedValue(userBalance);
 
       await expect(
-        jobService.createJob(userId, JobRequestType.FORTUNE, fortuneJobDto),
+        jobService.createJob(userMock, JobRequestType.FORTUNE, fortuneJobDto),
       ).rejects.toThrow(
         new ControlledError(ErrorJob.NotEnoughFunds, HttpStatus.BAD_REQUEST),
       );
     });
+
```
portuu3, line 628: Missing:
- Should create a job if the user is whitelisted.
- Should create a job id the user is not whitelisted and has a credit card.

How is possible that other test cases work if:
- isUserWhitelisted is not mocked
- getDefaultPaymentMethod is not mocked

portuu3 (changes requested):

portuu3 (approved):

Issue #2938, opened 2024-12-18
[Job Launcher] Show most recent jobs first


portuu3:
**Description**
Show the most recent jobs first for all statuses

**Motivation**
Improve UX

PR #2943, opened 2024-12-19
[Job Launcher] Show most recent jobs first

flopez7:
## Issue tracking
#2938

## Context behind the change
Show the most recent jobs first for all statuses

## How has this been tested?
Deployed and tested locally

## Release plan
None

None


portuu3 (changes requested):
I think you are missing sort field
Can you add an ordering button also?

portuu3 (approved):

PR #2946, opened 2024-12-20
[Reputation Oracle] feat: add support for batched payouts

### LINKED ISSUE #2903: [Reputation Oracle] Add support for batched payments when more than 100 escrow recipients
**Description**
After all jobs from escrows are completed and we have to pay the users that worked on them, we call `bulkPayOut` method of the Escrow, but under the hood it has a limit of `100` recipients that can be exceeded in reality when there are more users that successfully finished some jobs and should get paid for that. Right now reputation oracle fails to pay due to aforementioned limit and we have to prepare batches and pay manually, but it should be automatic.

**Motivation**
Automatic execution of payments when there are more recipients for bounty than supported by escrow

**Implementation details** 
Extend "Escrow completion" process by introducing new status - `awaiting_payments` and modify process to work the next way:
- when in `pending` status, it should create batches of recipients for further processing and set process to `awaiting_payments`
- when in `awaiting_payments`, it should process all batches for payments by manually creating a transaction and using its hash to track status in blockchain to avoid "replays" (aka double-payouts)
- after all batches are paid - set process to `paid` and process it w/o changes

dnechay:
## Issue tracking
Resolves #2903 

## Context behind the change
Functionality to create raw transaction and calculate its hash is placed to SDK because it should encapsulate contract interaction, but sending raw transaction can be done manually in the service.

When processing batches, we have to store nonce in the DB in order to handle different edge-cases by providing it for retry.

When running `processPaidEscrowCompletion` we now have to check also for `Partial` escrow status because when paying in batches `bulkPayOut` is called with `forceComplete=false`, so escrow won't be in `Paid` and won't autocomplete.

## How has this been tested?
**TS SDK**
- [x] unit tests
- [x] create example script and call `createBulkPayoutTransaction` with testnet credentials, send `rawTransaction`, verify that real transaction hash matches the one calculated by lib method
- [x] call `createBulkPayoutTransaction` with old nonce and make sure that real transaction fails with `NONCE_EXPIRED` error

**Python SDK**
- [x] unit tests
- [x] e2e test

**RepO**
Locally, while connecting to local DB and testnet blockchain, implemented "test" endpoint to trigger it and manipulate different parts of the `EscrowCompletionService` class
- [x] call `createEscrowPayoutsBatch`, make sure that entity is created with correct properties
- [x] call  `createEscrowPayoutsBatch` with replay; make sure that replay fails due to unique index
- [x] simulate errors in `processPayoutsBatch`, call `processAwaitingPayouts`; make sure `escrowCompletionEntity` gets updated retries count
- [x] have valid payouts batch in db, call `processAwaitingPayouts`, make sure that upon successful payout tx `escrowCompletionEntity` gets `PAID` status
- [x] unit tests

## Release plan
- [ ] 1. Wait till cron executions finishes on RepO
- [ ] 2. Stop RepO service
- [ ] 3. Clear `cron-jobs` table
- [ ] 4. Run DB migrations
- [ ] 4. Deploy and start new version of RepO service

In case we run a migration w/ service running - it can lead to some jobs get stuck. It can be solved by removing their "lock" in `cron-jobs` table.


packages/apps/reputation-oracle/server/src/database/migrations/1734697642348-escrowAwaitingPayouts.ts
```diff
@@ -0,0 +1,209 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
Dzeranov, line 1: Did you autogenerate a migration against an empty(new) database?

packages/apps/reputation-oracle/server/src/database/database.module.ts
```diff
@@ -22,6 +22,7 @@ import { UserQualificationEntity } from '../modules/qualification/user-qualifica
 import { WebhookIncomingEntity } from '../modules/webhook/webhook-incoming.entity';
 import { WebhookOutgoingEntity } from '../modules/webhook/webhook-outgoing.entity';
 import { EscrowCompletionEntity } from '../modules/escrow-completion/escrow-completion.entity';
+import { EscrowPayoutsBatchEntity } from 'src/modules/escrow-completion/escrow-payouts-batch.entity';
```
Dzeranov, line 25: Please, use relative path.

packages/apps/reputation-oracle/server/src/database/migrations/1734697642348-escrowAwaitingPayouts.ts
dnechay: Yup, messed it up while working on schema 😬 
Fixed in https://github.com/humanprotocol/human-protocol/pull/2946/commits/a6c2701f8c737967cc5a566436dc32f4c57119dc

packages/apps/reputation-oracle/server/src/database/database.module.ts
dnechay: Fixed in https://github.com/humanprotocol/human-protocol/pull/2946/commits/9e7990737a0132b35cf282fda7833902468176a2

Dzeranov (approved):
lgtm, lets wait for @portuu3

packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
```diff
@@ -116,21 +127,49 @@ export class EscrowCompletionService {
...
-            escrowCompletionEntity.finalResultsHash,
+          );
+
+          /**
+           * When creating payout batches we need to guarantee deterministic result,
+           * so order it first.
+           */
+          const payoutBatches = _.chunk(
+            _.orderBy(calculatedPayouts, 'address', 'asc'),
+            ESCROW_BULK_MAX_COUNT,
```
portuu3, line 142: According to the contract pay outs length must be less than 100. So ESCROW_BULK_MAX_COUNT should 99

portuu3 (changes requested):
TS sdk docs missing

packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
portuu3: Agree, but a contract upgrade means a long process for such a small change. Let's have it mind for future updates.

portuu3 (approved):
lgtm

PR #2965, opened 2024-12-29
[CVAT] Update quality metric to precision

zhiltsov-max:
## Context behind the change

Depends on https://github.com/cvat-ai/cvat/pull/8888
Includes / Depends on #2978

These changes make the annotation quality more understandable from the annotator viewpoint in the skeleton point annotation tasks. Simple precision has a flaw that if not enough annotations is provided, only the provided ones will be checked. This is workarounded by considering empty frames annotated as "empty", which makes it impossible to provide not enough annotations. As the result, we'll obtain `Precision = TP / max(job_anns, gt_anns)`. This change will increase the quality metrics in assignments in general, as accuracy includes both GT as user annotations in the denominator, while precision includes only the user annotations. 

- Changed quality metric from accuracy to precision for skeleton point annotation tasks
- Changed CVAT project quality setting `match_empty_frames` to the new `empty_is_annotated`


zhiltsov-max:
Need to upgrade CVAT SDK to 2.23.2 once released.

Issue #2928, opened 2024-12-17
[Reputation Oracle] Refactor error handling


Dzeranov:
**Description**
Current error handling mechanism needs to be reworked:
1. Errors need to be standardised, with the ability to create custom error classes and rely on `instanceof` operator.
2. Error context should be in the error object itself (e.g. userId etc.).
3. Error messages should not contain any dynamic values.
4. Handle database errors on repository level, do not throw httpexceptions on service level.

**Motivation**
1. Cleaner and easier to maintain codebase.
2. Improve observability and troubleshooting.

**Roadmap**
Refactor error handling in the following modules:
- [x] [Auth](https://github.com/humanprotocol/human-protocol/pull/2960)
- [x] [Cron-job](https://github.com/humanprotocol/human-protocol/pull/3062)
- [x] [Escrow-completion](https://github.com/humanprotocol/human-protocol/pull/3057)
- [x] ~~Health~~
- [x] [KYC](https://github.com/humanprotocol/human-protocol/pull/2937)
- [x] [Payout](https://github.com/humanprotocol/human-protocol/pull/3062) 
- [x] [Qualification](https://github.com/humanprotocol/human-protocol/pull/2944)
- [x] [Reputation](https://github.com/humanprotocol/human-protocol/pull/2947)
- [x] [Sendgrid](https://github.com/humanprotocol/human-protocol/pull/3062)
- [x] [Storage](https://github.com/humanprotocol/human-protocol/pull/3062)
- [x] [User](https://github.com/humanprotocol/human-protocol/pull/3042)
- [x] [Web3](https://github.com/humanprotocol/human-protocol/pull/3062)
- [x] [Webhook](https://github.com/humanprotocol/human-protocol/pull/2968)

PR #2968, opened 2025-01-03
[Reputation Oracle] Error handling for webhook module

flopez7:
## Issue tracking
#2928

## Context behind the change
- Refactored error handling for webhook module.
- Removed some unused imports.

## How has this been tested?
- Setup a local postgres database, apply migrations.
- Deploy reputation oracle locally.
- Send webhook with an invalid event type.

## Release plan
None

None

### DIFF
--- a/packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
+++ b/packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
@@ -21,7 +21,6 @@ import { ErrorEscrowCompletion } from '../../common/constants/errors';
 import { ControlledError } from '../../common/errors/controlled';
 import { WebhookOutgoingService } from '../webhook/webhook-outgoing.service';
 import { isDuplicatedError } from '../../common/utils/database';
-import { WebhookErrorMessage } from '../webhook/webhook.error';
 
 @Injectable()
 export class EscrowCompletionService {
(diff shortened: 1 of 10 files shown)


packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
```diff
@@ -206,7 +204,7 @@ export class EscrowCompletionService {
         for (const url of callbackUrls) {
           if (!url) {
             throw new ControlledError(
-              ErrorWebhook.UrlNotFound,
+              WebhookErrorMessage.URL_NOT_FOUND,
```
dnechay, line 207: IMO `WebhookErrorMessage` should be used for error message only in `webhook` module itself (i.e. where webhook-related errors are thrown). Here we attempt to retrieve callback urls for creating webhooks and it's different piece of business logic. I would suggest hardcode `not found` error message here for now and later we refactor it together with other errors for this service, wdyt?

dnechay (commented):
LGTM, one minor suggestion

packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
flopez7: Agree, it makes sense to hardcode the error message here for now to keep this part of the code simple while we continue with the refactor. I'm also on board with revisiting and centralizing error handling for this service, along with related modules, as part of a broader refactor later. I'll add a comment in the code to indicate this is a temporary solution and will be refactored

dnechay (approved):

PR #2978, opened 2025-01-08
[CVAT] Auto gt ban threshold

zhiltsov-max:
## Context behind the change

Depends on #2976

- GT ban threshold is changed to 1 - manifest quality instead of a configuration constant


packages/examples/cvat/recording-oracle/alembic/versions/1e89224ad721_add_gt_labels_in_gt_stats.py
```diff
@@ -0,0 +1,40 @@
+"""add_gt_labels_in_gt_stats
```
Dzeranov, line 1: Looks like a duplication of:

https://github.com/humanprotocol/human-protocol/blob/develop/packages/examples/cvat/recording-oracle/alembic/versions/1e89224ad721_add_gt_labels_in_gt_stats.py

packages/apps/reputation-oracle/server/src/database/database.module.ts
```diff
@@ -42,7 +42,7 @@ import { EscrowPayoutsBatchEntity } from '../modules/escrow-completion/escrow-pa
         typeOrmLoggerService.setOptions(
           loggerOptions && loggerOptions[0] === 'all'
             ? 'all'
-            : (loggerOptions as LoggerOptions) ?? false,
+            : ((loggerOptions as LoggerOptions) ?? false),
```
zhiltsov-max, line 45: @Dzeranov, probably, some linter ignore is needed for this line. It gets autoformatted.

packages/examples/cvat/recording-oracle/alembic/versions/1e89224ad721_add_gt_labels_in_gt_stats.py
zhiltsov-max: It was from the dependency PR. No more shown after merging with develop.

Dzeranov (approved):

PR #2979, opened 2025-01-09
[HUM-138]: chore: refactor shared helpers

mpblocky:
HUM-138

- Removed unused shared helpers
- Redistributed shared helpers


mpblocky:
@dnechay ready for review

dnechay (commented):
Changes in PR lgtm. Let's do some extra steps in helpers refactor:
1. `defaultErrorMessage` and `jsonRpcErrorHandler` can be placed in some common place
2. There are some schemas in `helpers` folder and it's counter-intuitive to look for them there. Taking into account how they are used it would be better to move them to e.g. `shared/schemas`
3. `wait` should be better named `delay`, because from the function named `wait` you might want to expect that you pass not just ms value, but some promise or a callback. By similar reason `sleep` is not good, because in other different languages it means "pause the whole process", which is not the case for JS. So `delay` is believed to be the most common name for it. At the same time helper time can be renamed from `async` to `timers` because it's actually an "timer" utility, not something to do with async flows

mpblocky:
@dnechay I made changes for #2 and #3, however I'm not sure about #1. Which location would You prefer?

1. `shared/helpers/error.ts`
2. `shared/errors/[files]`?

dnechay:
> @dnechay I made changes for #2 and #3, however I'm not sure about #1. Which location would You prefer?
> 
> 1. `shared/helpers/error.ts`
> 2. `shared/errors/[files]`?

My logic is next: in `helpers` you would place something that by some unknown reason you might export to a separate library that can be published and easily used elsewhere. These error helpers don't fit this definition and look way too specific for this app/UI, so I would place them to something like `shared/errors/handlers.ts` (don't forget about `index.ts` as well) and also rename `defaultErrorMessage` -> `getErrorMessageForError`

mpblocky:
@dnechay sounds good, added changes :)

PR #2987, opened 2025-01-10
[HUM-127]: chore: refactor page card components

mpblocky:
## Issue tracking
HUM-127

## How has this been tested?
- `withLayoutBackground` - manually checked on mocked page, how do each of `page-card*.tsx` look
- manual checks for other changes

## Release plan
normal deploy

N/A


packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card-loader.tsx
```diff
@@ -0,0 +1,41 @@
...
+  cardMaxWidth = '100%',
+}: {
+  cardMaxWidth?: string;
+  withLayoutBackground?: boolean;
+}) {
+  const { isDarkMode } = useColorMode();
+  const { setGrayBackground } = useBackgroundColorStore();
+
+  useEffect(() => {
+    if (withLayoutBackground) {
```
adrian-oleskiewicz, line 19: This prop is not used so can be removed than we can get rid of this // eslint-disable-next-line

mpblocky:
@dnechay ready for review

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card-error.tsx
```diff
@@ -1,41 +1,13 @@
...
-  width: '100%',
-  background: constDarkColorPalette.paper.main,
-  [breakpoints.mobile]: {
-    background: constDarkColorPalette.backgroundColor,
-  },
-};
+import { commonDarkPageCardStyles, commonPageCardStyles } from './consts';
 
 export function PageCardError({
   errorMessage,
```
dnechay, line 13: Would be nice to move types for this FC props out of function signature and have meaningful names for them, like:
```
type CommonProps = {
  cardMaxWidth?: string;
  withLayoutBackground?: boolean;
};

type ErrorMessageProps = CommonProps & {
  errorMessage: string;
};

type ChildErrorProps = CommonProps & {
  children: React.ReactElement;
};

PageCardError(props: ErrorMessageProps | ChildErrorProps)
```
or something else

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
```diff
@@ -197,7 +168,7 @@ export function PageCard({
               >
                 <ArrowBackIcon fontSize="inherit" />
               </IconWrapper>
-            ) : null}
+            )}
             {!hiddenCancelButton && (
               <Button
                 onClick={() => {
```
dnechay, line 174: This and others anonymous function can be moved to some `handle<btn>Click` (e.g. `handleCancelButtonClick`) for better readability: ideally JSX in `return` should carry only view-related code, so you don't care about business part and can think just of UI

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
```diff
@@ -245,7 +216,7 @@ export function PageCard({
             }}
             xs={12}
           >
-            {backArrowPath && !hiddenArrowButton ? (
+            {backArrowPath && !hiddenArrowButton && (
               <IconWrapper
                 onClick={goBack.bind(null, backArrowPath)}
```
dnechay, line 221: This bind can be done once, no need to do it on every render

mpblocky: Good catch!

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card-loader.tsx
```diff
@@ -0,0 +1,41 @@
+import { Grid } from '@mui/material';
+import { useEffect } from 'react';
+import { useBackgroundColorStore } from '@/shared/hooks/use-background-store';
+import { Loader } from '@/shared/components/ui/loader';
+import { useColorMode } from '@/shared/hooks/use-color-mode';
+import { commonDarkPageCardStyles, commonPageCardStyles } from './consts';
+
+export function PageCardLoader({
```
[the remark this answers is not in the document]
mpblocky: I checked the impact of that prop and it doesn't make any change, whether it's set to true or false in both modes. Will delete that!

packages/apps/human-app/frontend/src/modules/auth/providers/require-auth.tsx
```diff
@@ -3,7 +3,7 @@ import { createContext } from 'react';
 import { useAuth } from '@/modules/auth/hooks/use-auth';
 import { routerPaths } from '@/router/router-paths';
 import type { AuthenticatedUserContextType } from '@/modules/auth/context/auth-context';
-import { PageCardLoader } from '@/shared/components/ui/page-card-loader';
+import { PageCardLoader } from '@/shared/components/ui/page-card';
```
migace, line 6: why do we have a PageCardLoader component in the `page-card` file? I think the name of the component should be exactly like the name of the file

packages/apps/human-app/frontend/src/modules/homepage/components/choose-sign-up-account-type.tsx
```diff
@@ -27,10 +27,10 @@ export function ChooseSignUpAccountType() {
 
   return (
     <PageCard
-      backArrowPath={backToWelcomeStage}
-      cancelRouterPathOrCallback={backToWelcomeStage}
+      arrowButtonNavigationTarget={backToWelcomeStage}
```
migace, line 30: I do not like this naming, maybe `backNavigation` and `closeNavigation` instead `cancelButtonNavigationTarget`?

mpblocky: Yes, this sounds more concise, will do that

mpblocky:
@dnechay this one is ready for re-review :)

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
```diff
@@ -95,34 +62,35 @@ export function PageCard({
...
       return;
     }
-    if (typeof pathOrCallback === 'string') {
-      navigate(pathOrCallback);
+    if (typeof navigationTarget === 'string') {
+      navigate(navigationTarget);
       return;
     }
-    navigate(pathOrCallback);
+    navigate(navigationTarget);
```
dnechay, line 74: Looks like we don't need last line in this fn
```suggestion
    if (navigationTarget instanceof Function) {
      navigationTarget();
    } else {
      navigate(navigationTarget);
    }
```

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
```diff
@@ -95,34 +62,35 @@ export function PageCard({
...
     <Grid
       container
       sx={{
-        ...(isDarkMode ? commonStylesDark : commonStyles),
+        ...(isDarkMode ? commonDarkPageCardStyles : commonPageCardStyles),
         padding: isMobile ? '0 2rem 7.25rem 2rem' : '2rem 2rem 7.7rem 2rem',
       }}
     >
-      {!hiddenCancelButton && (
+      {!hideCancelButton && (
```
dnechay, line 93: As mentioned [here](https://github.com/humanprotocol/human-protocol/pull/2987#discussion_r1913414369), let's use `showCancelButton` name for this prop with default `true` and pass `showCancelButton={false}` where needed

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
```diff
@@ -178,14 +138,14 @@ export function PageCard({
...
                 display: 'flex',
                 width: '100%',
-                justifyContent: backArrowPath ? 'space-between' : 'flex-end',
+                justifyContent: backNavigation ? 'space-between' : 'flex-end',
                 alignItems: 'center',
               },
             }}
           >
-            {backArrowPath && !hiddenArrowButton ? (
+            {backNavigation && !hideBackButton && (
```
dnechay, line 146: 1. Let's please rename `hideBackButton` to `showBackButton` with `true` as default
2. `backNavigation && !hideBackButton` - this condition controls arrow button displaying in two places, so we can move it above like:
```
const _showBackButton = backNavigation && showBackButton;
/// later use it like
{_showBackButton && (<IconWrapper....
```

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
```diff
@@ -95,34 +62,35 @@ export function PageCard({
...
-      navigate(pathOrCallback);
+    if (typeof navigationTarget === 'string') {
+      navigate(navigationTarget);
       return;
     }
-    navigate(pathOrCallback);
+    navigate(navigationTarget);
+  };
+
+  const handleArrowButton = () => {
```
dnechay, line 77:
```suggestion
  const handleBackButton = () => {
```

dnechay (commented):
LGTM overall
@mpblocky few small style comments

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
dnechay: There is a type above, it will be
```
<PageCard
  hideCancelButton
  ...
>
```
vs
```
<PageCard
  showCancelButton={false}
...
```
I agree that it looks more attractive as `hideCancelButton`, but at the same time you can make `showCancelButton` as `false` by default and use it as
```
<PageCard
  showCancelButton
...
```

There are some examples from frameworks we use about similar stuff: [react-query](https://github.com/humanprotocol/human-protocol/blob/62b1e8b13884577b4688c576eb926eb044891d6a/packages/apps/human-app/frontend/src/main.tsx#L54), [material-ui](https://github.com/humanprotocol/human-protocol/blob/62b1e8b13884577b4688c576eb926eb044891d6a/packages/apps/staking/src/pages/Dashboard/index.tsx#L19), [recharts](https://github.com/humanprotocol/human-protocol/blob/62b1e8b13884577b4688c576eb926eb044891d6a/packages/apps/dashboard/ui-2024/src/components/Home/SmallGraph.tsx#L106), so it's a common practice to use "positive" statement instead of "negative"

mpblocky: Alright, I will revert to that then :)

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
mpblocky: Actually I cannot do that; `navigate` has multiple definitions and eslint does not let me do that:

```
    (to: To, options?: NavigateOptions): void;
    (delta: number): void;
}
``` 
where 

```
export type To = string | Partial<Path>;
```

So we have to leave it as is.

mpblocky:
@dnechay applied suggestions, I think it looks good now :)

packages/apps/human-app/frontend/src/shared/components/ui/page-card/page-card.tsx
dnechay: That seems reasonable from `useNavigate` developers.

From what I see, we use `-1` only as default value for `backNavigation`, so IMU we can get rid of it in `NavigationTarget` type, leaving it as `string | (() => void);`

Then we will be able to change the code to something like:
```
const followNavigationTarget = (navigationTarget: NavigationTarget) => {
    if (navigationTarget instanceof Function) {
      navigationTarget();
    } else if (typeof navigationTarget === 'string') {
      navigate(navigationTarget);
   } else {
      navigate(-1);
   }
```

Looks cleaner, wdyt?

dnechay: @mpblocky 
That seems reasonable from `useNavigate` developers.

From what I see, we use `-1` only as default value for `backNavigation`, so IMU we can get rid of it in `NavigationTarget` type, leaving it as `string | (() => void);`

Then we will be able to change the code to something like:
```
const followNavigationTarget = (navigationTarget: NavigationTarget) => {
    if (navigationTarget instanceof Function) {
      navigationTarget();
    } else if (typeof navigationTarget === 'string') {
      navigate(navigationTarget);
   } else {
      navigate(-1);
   }
```

Looks cleaner, wdyt?

mpblocky: It looks pretty similar and we are taking away to go back one page. I can do that, but are You sure You want this change?

dnechay: I don't see a case when we might want to pass `-2` or different numeric value. IMU you either want a default "go back" behavior (i.e. to exact previous route, which is `-1`), or you specific some specific one with exact path or cb.

If the above is correct - then yes, I prefer to change the implementation to something explicit that won't cause confusion with `navigate` typings again :)

dnechay: @mpblocky I don't see a case when we might want to pass `-2` or different numeric value. IMU you either want a default "go back" behavior (i.e. to exact previous route, which is `-1`), or you specific some specific one with exact path or cb.

If the above is correct - then yes, I prefer to change the implementation to something explicit that won't cause confusion with `navigate` typings again :)

mpblocky: Sounds good :)

mpblocky:
All done @dnechay

dnechay (approved):
Nice!

PR #2991, opened 2025-01-10
[HUM-128]: chore: split hcaptcha components to separate files

mpblocky:
## Issue tracking
HUM-128

## Context behind the change
Refactor:
- used `useImperativeHandle` to expose captcha's `reset` function
- changed names to more meaningful

## How has this been tested?
Checked if captcha behaves the same as before the changes; if it works correctly for login, if it resets after filling the form and clicking "forgot password"

## Release plan
normal deploy

N/A


mpblocky:
@dnechay ready for review

packages/apps/human-app/frontend/src/shared/components/hcaptcha/h-captcha.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+export function CustomHCaptcha({
+  setCustomHCaptchaToken,
+  error,
+}: CustomHCaptchaProps) {
+  const { isDarkMode } = useColorMode();
+  const captchaRef = useRef<HCaptcha>(null);
+
+  useEffect(() => {
+    if (error instanceof FetchError) {
```
dnechay, line 20: `error` is used here not for actual error handling or displaying error message, but for controlling captcha flow where you reset it.

In place where you use `CustomHCaptcha` you might want to reset it even if you don't have `FetchError`, but have more control over it, so instead of relying on `error` property it should be something different.

Potential options are:
1. have some property like `attemptId` and rely on `useEffect` with `captchaRef.current?.resetCaptcha();` call that triggers every time `attemptId` is changed
2. implement `useImperativeHandle` and provide specific API with `reset()` method

In the first scenario it might be a bit confusing on why you need to have this `attemptId` and the flow with `reset` is implicit, so I would suggest going with the second option because it's clear.

packages/apps/human-app/frontend/src/shared/components/hcaptcha/h-captcha-form.tsx
```diff
@@ -0,0 +1,42 @@
...
+  name: string;
+  error?: unknown;
+}
+
+export function HCaptchaForm({ name, error }: HCaptchaFormProps) {
+  const { colorPalette } = useColorMode();
+  const { setValue, formState } = useFormContext<Record<string, unknown>>();
+  const customCaptchaRef = useRef<CustomHCaptchaRef>(null);
+
+  function onVerify(token: string) {
```
mgacek-blockydevs, line 19: name of the function does not tell what do

mpblocky:
@dnechay PR ready!

dnechay (approved):
Nice work here! Looks much better

PR #2998, opened 2025-01-13
HUM-89 - oracles table data refactor

adrian-oleskiewicz:
## Issue tracking

HUM-89
## Context behind the change

1. Files structure refactor
2. Split code (SRP): components, utils
## How has this been tested?

- checked display of oracles table both mobile and desktop
- checked if selecting oracles works
- checked if filtering works
## Release plan

normal deploy

N/A


packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/utils/should-navigate-to-registration.ts
```diff
@@ -0,0 +1,14 @@
+import type { Oracle } from '@/modules/worker/services/oracles';
```
mpblocky, line 1: I would keep one name for the folders `utils` -> `helpers`

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/components/oracles-table-mobile.tsx
```diff
@@ -74,7 +77,7 @@ export function OraclesTableMobile({
           <TableButton
             fullWidth
             onClick={() => {
```
migace, line 79: I would prefer move it over return statement

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/helpers/get-oracles-table-columns.tsx
```diff
@@ -0,0 +1,61 @@
+import { t } from 'i18next';
+import { Grid } from '@mui/material';
+import type { MRT_ColumnDef } from 'material-react-table';
+import type { Oracle } from '@/modules/worker/services/oracles';
+import { EvmAddress } from '@/modules/worker/components/jobs/evm-address';
+import { Chips } from '@/shared/components/ui/chips';
+import { TableButton } from '@/shared/components/ui/table-button';
+import { type JobType } from '@/modules/smart-contracts/EthKVStore/config';
+
+export const getOraclesTablesColumns = (
```
migace, line 10: maybe it will be better to replace it by hook and use useMemo here?

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/helpers/navigate-to-oracle-jobs.ts
```diff
@@ -0,0 +1,12 @@
+import { type NavigateFunction } from 'react-router-dom';
+import type { Oracle } from '@/modules/worker/services/oracles';
+import { routerPaths } from '@/router/router-paths';
+
+export const navigateToOracleJobs = (
```
migace, line 5: `handleHCaptchaNavigation` -> Both components perform very similar logic, but they have completely different names. Maybe we could use the same convention?

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/helpers/should-navigate-to-registration.ts
```diff
@@ -0,0 +1,14 @@
...
+
+interface RegistrationResult {
+  oracle_addresses: string[];
+}
+
+export const shouldNavigateToRegistration = (
+  oracle: Oracle,
+  registrationData?: RegistrationResult
+): boolean => {
+  if (!oracle.registrationNeeded) return false;
```
migace, line 11:
```
if (true) {
  return true
}
```

it is not necessary here....

```
return !!oracle.registrationNeeded && !registrationData?.oracle_addresses?.includes(oracle.address);
```

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/components/oracles-table-mobile.tsx
adrian-oleskiewicz: I moved this logic to separate file

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/components/oracles-table-item-mobile.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { TableButton } from '@/shared/components/ui/table-button';
+import { EvmAddress } from '@/modules/worker/components/jobs/evm-address';
+import { ListItem } from '@/shared/components/ui/list-item';
+import type { JobType } from '@/modules/smart-contracts/EthKVStore/config';
+import { useColorMode } from '@/shared/hooks/use-color-mode';
+import { type Oracle } from '@/modules/worker/services/oracles';
+
+interface OraclesTableItemMobileProps {
+  oracle: Oracle;
+  selectOracle: (oracle: Oracle) => void;
```
mgacek-blockydevs, line 13: I don't like that `OraclesTableItemMobile` knows about oracle, and have possiblity to set it. I think, that change name like `oracle` and `onClick` will be better here

mgacek-blockydevs: the second problem, you are doing `props drilling` to pass it: 

```
const selectOracle = (oracle: Oracle) => {
    handleOracleNavigation({
      oracle,
      siteKey: user.site_key,
      navigate,
      registrationData: registrationInExchangeOraclesResult,
    });
  };
  ```
  
  What do you think about call this function in this component directly? 
  You do the same think with `oraclesQueryDataResult` maybe it would be a better idea to call function `useGetOracles` inside of this component? We use react query, so we cache data for it
  
  Maybe `OraclesTable` is not necessary, and you could move content of this file directly into `JobsDiscoveryPage`?

File: packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/components/oracles-table-mobile.tsx [hunk 7]
```diff
@@ -0,0 +1,49 @@
...
+      <Grid
+        container
+        sx={{ width: '100%', justifyContent: 'center', alignItems: 'center' }}
+      >
+        <Loader />
+      </Grid>
+    );
+  }
+
+  if (isOraclesDataError) {
```
mgacek-blockydevs, line 34: you know it before call OracleTable component, so I think you can show it in `JobsDiscoveryPage` component

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/components/oracles-table-mobile.tsx
```diff
@@ -0,0 +1,49 @@
...
+  selectOracle,
+  oraclesQueryDataResult,
+}: OraclesTableMobileProps) {
+  const {
+    data: oraclesData,
+    isError: isOraclesDataError,
+    isPending: isOraclesDataPending,
+  } = oraclesQueryDataResult;
+
+  if (isOraclesDataPending) {
```
mgacek-blockydevs, line 23: same like below

packages/apps/human-app/frontend/src/modules/worker/components/jobs-discovery/components/oracles-table-item-mobile.tsx
adrian-oleskiewicz: I moved this logic out of this component.
All logic from related with useGetOracles() hook is now handled in <OraclesTable />

I also deleted <OraclesTableMobile /> component

mpblocky:
@dnechay ready!

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/components/oracles-table.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  if (isMobile && isOraclesDataError) {
+    return <NoRecords />;
+  }
+
+  return isMobile ? (
+    <Stack flexDirection="column">
+      {(oraclesData ?? []).map((oracle) => (
+        <OraclesTableItemMobile
+          key={`${oracle.address}-${oracle.name}`}
```
dnechay, line 46: For uniqueness we might need to use combination of `oracle.address` and `oracle.chainId` (it's part of the response already)

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/components/oracles-table.tsx
```diff
@@ -0,0 +1,58 @@
...
+    );
+  }
+
+  if (isMobile && isOraclesDataError) {
+    return <NoRecords />;
+  }
+
+  return isMobile ? (
+    <Stack flexDirection="column">
+      {(oraclesData ?? []).map((oracle) => (
```
dnechay, line 44: Maybe somewhere about we have `const _oracleData = oraclesData ?? []` and use it here and below?
Also might make sense to show `NoRecords` if no oracles, wdyt?

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/components/oracles-table-desktop.tsx
```diff
@@ -0,0 +1,55 @@
...
+} from 'material-react-table';
+import { createTableDarkMode } from '@/shared/styles/create-table-dark-mode';
+import { useColorMode } from '@/shared/contexts/color-mode';
+import { useOraclesTableColumns } from '../hooks/use-oracles-table-columns';
+import { type Oracle } from '../hooks';
+
+interface OraclesTableDesktopProps {
+  isOraclesDataPending: boolean;
+  isOraclesDataError: boolean;
+  oraclesData: Oracle[] | undefined;
```
dnechay, line 13: Maybe `oraclesData: Oracle[]`? You already have `oracleData ?? []` in the parent component

dnechay (commented):
LGTM overall

Left some comments and please resolve merge conflict, I will carefully test it later

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/components/oracles-table.tsx
mpblocky: What `chainId` should this have?

```
const H_CAPTCHA_ORACLE: Oracle = {
  address: env.VITE_H_CAPTCHA_ORACLE_ADDRESS,
  jobTypes: env.VITE_H_CAPTCHA_ORACLE_TASK_TYPES,
  role: env.VITE_H_CAPTCHA_ORACLE_ROLE,
  url: env.VITE_H_CAPTCHA_ORACLE_ANNOTATION_TOOL,
  name: 'hCaptcha',
  registrationNeeded: false,
};
```

Should I do some check like `isTestnet ? amoy.chainId : polygon.chainId`?

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/components/oracles-table.tsx
mpblocky: Sounds good!

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/components/oracles-table.tsx
dnechay: Good question!
hCaptcha is available only on polygon & amoy yes, so the check above is good enough

dnechay: @mpblocky Good question!
hCaptcha is available only on polygon & amoy yes, so the check above is good enough

dnechay (approved):

PR #2999, opened 2025-01-13
[Reputation Oracle] hCaptcha token validation moved to the `HCaptchaGuard`

### LINKED ISSUE #2995: [Reputation Oracle] Implement `hCaptcha` guard
**Description**
Implement `hCaptcha` guard instead of checking token's validity in a service.

**Motivation**
`hCaptcha` token validation is not a part of a business logic, it should be handled on a controller level.

**Implementation details** 
1. Implement `hCaptcha` guard.
2. Remove `hCaptcha` token validations (except for labeling part) from servies.

Dzeranov:
## Issue tracking

To close #2995 

## Context behind the change

hCaptcha token validation moved to the `HCaptchaGuard` to remove code duplications and simplify services logic.

 
## How has this been tested?

- [x] Reputation Oracle locally deployed and manually tested all the affected endpoints

## Release plan

- [x] Add a value for `HUMAN_APP_EMAIL` in the environment.

N/A

Please, do not merge until [this PR ](https://github.com/humanprotocol/human-protocol/pull/2977)is merged.


packages/apps/reputation-oracle/server/src/common/config/auth-config.service.ts
```diff
@@ -61,4 +61,11 @@ export class AuthConfigService {
       86400,
     );
   }
+
+  /**
+   * Human APP email.
+   */
+  get humanAppEmail(): string {
+    return this.configService.getOrThrow<string>('HUMAN_APP_EMAIL');
```
dnechay, line 69: Would it be useful to add email validation for this env variable to `envValidator` we have?

packages/apps/reputation-oracle/server/src/modules/user/user.service.ts
```diff
@@ -369,24 +364,8 @@ export class UserService {
 
   public async registrationInExchangeOracle(
     user: UserEntity,
-    { hCaptchaToken, oracleAddress }: RegistrationInExchangeOracleDto,
+    { oracleAddress }: RegistrationInExchangeOracleDto,
```
dnechay, line 367: Maybe change it to
```suggestion
    oracleAddress: string,
```
and pass it as a string in controller?

packages/apps/reputation-oracle/server/src/common/guards/hcaptcha.ts
```diff
@@ -0,0 +1,57 @@
...
+  public async canActivate(context: ExecutionContext): Promise<boolean> {
+    const request: Request = context.switchToHttp().getRequest();
+
+    const { body } = request;
+    const hCaptchaToken = body['h_captcha_token'];
+
+    // TODO: Remove 27-46 lines once we figure out how to replace human app user
+    if (request.path === '/auth/signin') {
+      const email = body['email'];
+      // Need to validate email because guards being called before any interceptors or pipes
```
dnechay, line 29: As discussed offline, no need to validate email here since it will be validated later in route pipes, so comment can be removed

dnechay (commented):
LGTM, left some suggestions

After moving captcha verification logic to its own guard we can clearly see that e.g. `SignInDto` is not proper type of `AuthService.signIn` because it contains excessive information like captcha token. Ideally, DTO should be used when transferring data (e.g. for request and response bodies and proper conversion should be done in place) and services should have their own types, but it's a separate topic, not related to this PR

Dzeranov:
> LGTM, left some suggestions
> 
> After moving captcha verification logic to its own guard we can clearly see that e.g. `SignInDto` is not proper type of `AuthService.signIn` because it contains excessive information like captcha token. Ideally, DTO should be used when transferring data (e.g. for request and response bodies and proper conversion should be done in place) and services should have their own types, but it's a separate topic, not related to this PR

Agree with that but let it be a part of another refactoring phase.

dnechay (approved):

packages/apps/reputation-oracle/server/src/common/guards/hcaptcha.ts
```diff
@@ -0,0 +1,56 @@
...
+    private readonly hCaptchaService: HCaptchaService,
+    private readonly authConfigSerice: AuthConfigService,
+  ) {}
+  public async canActivate(context: ExecutionContext): Promise<boolean> {
+    const request: Request = context.switchToHttp().getRequest();
+
+    const { body } = request;
+    const hCaptchaToken = body['h_captcha_token'];
+
+    // TODO: Remove 27-45 lines once we figure out how to replace human app user
```
portuu3, line 26: why don't we verify that user has the role as we had before? @Dzeranov

PR #3000, opened 2025-01-13
[HUM-134]: chore: redistribute shared wallet connect files

mpblocky:
## Issue tracking
HUM-134

## Context behind the change
- Redistributed files of wallet connect

## How has this been tested?
- checked if wallet connect login is possible

## Release plan
normal deploy

N/A


packages/apps/human-app/frontend/src/shared/contexts/wallet-connect.tsx
```diff
@@ -102,52 +72,60 @@ export function WalletConnectProvider({
...
+          signMessage,
+          initializing,
+        } satisfies WalletConnectContextConnectedAccount)
+      : ({
+          isConnected: false,
+          web3ProviderMutation,
+          openModal,
+          signMessage,
+          initializing,
+        } satisfies WalletConnectContextDisconnectedAccount);
```
mgacek-blockydevs, line 115: I'm not sure... chainId should be on both based on the type like address

packages/apps/human-app/frontend/src/shared/contexts/wallet-connect.tsx
```diff
@@ -102,52 +72,60 @@ export function WalletConnectProvider({
...
+      if (web3ProviderMutation.data) {
+        try {
+          const signature =
+            await web3ProviderMutation.data.signer.signMessage(message);
+          return signature;
+        } catch (error) {
+          throw new JsonRpcError(error);
+        }
+      }
+      return Promise.resolve(undefined);
```
dnechay, line 90: You don't need this line in `async` function: it resolves with `undefined` if other is not specified

dnechay (commented):
Nice work on types and optimization!

I would suggest to have `shared/contexts/wallet-connect` folder with internal files like:
- `wallet-connect.ts`
- `types.ts`
- `use-wallet-connect.ts` - hook
- `index.ts` - from where you re-export hook, context and types

wdyt?

mpblocky:
Yes, we can definitely do that to shrink the 2nd iteration PR, as in the other PRs

packages/apps/human-app/frontend/src/modules/auth-web3/hooks/use-connected-wallet.tsx
```diff
@@ -1,5 +1,5 @@
 import { useContext } from 'react';
-import { AuthWeb3Context } from '@/modules/auth-web3/providers/require-wallet-connect';
+import { AuthWeb3Context } from '@/shared/contexts/wallet-connect/require-wallet-connect';
```
dnechay, line 2:
```suggestion
import { AuthWeb3Context } from '@/shared/contexts/wallet-connect';
```

packages/apps/human-app/frontend/src/modules/auth-web3/hooks/use-connected-wallet.tsx
```diff
@@ -1,5 +1,5 @@
...
 
 export const useConnectedWallet = () => {
```
dnechay, line 4: Should we move this `useConnectedWallet` hook to `@/shared/contexts/wallet-connect` folder? It seems to be just a wrapper for using `AuthWeb3Context` that is declared there

packages/apps/human-app/frontend/src/shared/contexts/wallet-connect/wallet-connect.tsx
```diff
@@ -0,0 +1,133 @@
...
+      if (web3ProviderMutation.data) {
+        try {
+          const signature =
+            await web3ProviderMutation.data.signer.signMessage(message);
+          return signature;
+        } catch (error) {
+          throw new JsonRpcError(error);
+        }
+      }
+      return Promise.resolve(undefined);
```
dnechay, line 90:
```suggestion
```
You don't need to return resolved promise here

packages/apps/human-app/frontend/src/modules/auth-web3/hooks/use-connected-wallet.tsx
mpblocky: Yep, sounds good, I will also rename that context because the name is misleading a bit

mpblocky:
@dnechay fixed!

dnechay (approved):

PR #3005, opened 2025-01-13
[HUM-132]: chore: separate homepage state hook from context

mpblocky:
## Issue tracking
HUM-132

## Context behind the change
- Separate hook from context

## How has this been tested?
- checked if home page state is correctly provided (e.g. "how it works" in navbar) and functions depending on it are behaving same as before

## Release plan
normal deploy

N/A

### DIFF
--- a/packages/apps/human-app/frontend/src/shared/types/homepage-state.ts
+++ b/packages/apps/human-app/frontend/src/shared/types/homepage-state.ts
@@ -1,1 +1,0 @@
-export type HomePageStateType = 'welcome' | 'chooseSignUpAccountType';
--- a/packages/apps/human-app/frontend/src/shared/types/homepage-state.ts
+++ b/packages/apps/human-app/frontend/src/shared/types/homepage-state.ts
@@ -1,0 +1,1 @@
+export type HomePageStateType = 'welcome' | 'chooseSignUpAccountType';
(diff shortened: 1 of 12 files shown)


packages/apps/human-app/frontend/src/shared/types/homepage-state.ts
```diff
@@ -0,0 +1 @@
+export type HomePageStateType = 'welcome' | 'chooseSignUpAccountType';
```
mpblocky, line 1: Should this become an enum?

mgacek-blockydevs: no, it is ok

mpblocky:
@dnechay PR ready!

dnechay (commented):
Nice optimization of context with `useMemo`

Do we really need to move `useHomePageState` to `shared/hooks`? IMO it will only clutter `shared/hooks` folder.

I would suggest to have `shared/contexts/homepage-state` folder with internal files like:
- `homepage-state.ts`
- `types.ts` (seems to be excessive for now, `HomePageStateType` can be places in context file itself because not used anywhere else so there is no need to even export it)
- `use-homepage-state.ts` - hook
- `index.ts` - from where you re-export hook and context

wdyt?

mpblocky:
Yes, we can definetly do that to shrink the 2nd iteration PR :)

mpblocky:
@dnechay I'm not sure about keeping hook in there; do You want to keep context hooks near the context and some more general hooks in `shared/hooks`?

dnechay:
@mpblocky

> do You want to keep context hooks near the context and some more general hooks in shared/hooks?

Yes: IMO it's cleaner for our use case, because we don't have many different contexts that we use at the same time in different components and if we co-locate these hooks near to their contexts - it's easier to find things.

In case you anticipate there might be some complex use-cases - then I would keep them in `shared/hooks/context`

I believe it's better to keep things simple and straightforward till the moment you actually see there is a complex use-case and you need to refactor for it

mpblocky:
@dnechay yes, this sounds good, just wanted to be sure we are on the same page :)

mpblocky:
@dnechay fixed :)

dnechay (approved):

PR #3006, opened 2025-01-13
[HUM-131]: chore: redistribute code from color mode context

mpblocky:
## Issue tracking
HUM-131

## Context behind the change
- Redistributed files of color mode context

## How has this been tested?
- checked if the dark mode is switching correctly

## Release plan
normal deploy

N/A


packages/apps/human-app/frontend/src/shared/helpers/dark-mode.ts
```diff
@@ -0,0 +1,17 @@
+const MODE_LOCAL_STORAGE_KEY = 'mode';
+
+export const setModeInLocalStorage = (mode: 'dark' | 'light') => {
+  localStorage.setItem(MODE_LOCAL_STORAGE_KEY, mode);
+};
+
+export const isDarkInModeLocalStorage = () => {
```
mpblocky, line 7: Should we create an enum for modes?

packages/apps/human-app/frontend/src/shared/helpers/dark-mode.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const isDarkColorModeEnabledInLocalStorage = () => {
+  const mode = localStorage.getItem(MODE_LOCAL_STORAGE_KEY);
+  if (mode === 'dark') {
+    return true;
+  }
+  return false;
+};
+
+export const isColorModeStateSavedInLocalStorage = () => {
```
mpblocky, line 15: I'm not sure if the naming is right.

mgacek-blockydevs: yeah, sounds better for me

mpblocky:
@dnechay PR ready

packages/apps/human-app/frontend/src/shared/helpers/dark-mode.ts
```diff
@@ -0,0 +1,17 @@
+const MODE_LOCAL_STORAGE_KEY = 'mode';
```
dnechay, line 1: It shouldn't be a `shared` helper since it's used only in `ColorMode` context.

It's fine to keep these functions in file with the context itself (or next to it as e.g. `color-mode-settings.ts`

dnechay (commented):
Let's keep everything relaetd to `ColorMode` context in one folder: `shared/contexts/color-mode`

mpblocky:
Sure!

mpblocky:
@dnechay everything addressed :)

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode-settings.ts
```diff
@@ -0,0 +1,17 @@
+const MODE_LOCAL_STORAGE_KEY = 'mode';
+
+export const saveColorMode = (mode: 'dark' | 'light') => {
```
dnechay, line 3: Would be nice to have some `enum ColorMode` declared in this file and use it here and there

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode.tsx
```diff
@@ -39,18 +32,38 @@ interface ColorModeProviderProps {
...
-  const [isDarkMode, setIsDarkMode] = useState<boolean>(
-    isDarkInModeLocalStorage()
-  );
+export function ColorModeProvider({
+  children,
+}: Readonly<ColorModeProviderProps>) {
+  const [isDarkMode, setIsDarkMode] = useState<boolean>(isDarkColorMode());
+
+  useEffect(() => {
+    const handleColorModeChange = (matches: boolean) => {
```
dnechay, line 41: Maybe we can move `handleColorModeChange` and `runColorMode` function out of the provider component itself? Either to somewhere above or to some other file in `color-mode` context folder?

dnechay (commented):
LGTM overall, left some suggestions

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode-settings.ts
mpblocky: I actually thought about it but it seemed a bit excessive 😅 Do You want me to do it anyway?

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode.tsx
mpblocky: Sure :)

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode-settings.ts
dnechay: It might seem excessive because used only in this folder, however if you don't use hardcoded strings and use enums for that - it's easier to get in different places if it relates to enum or something else, so would be nice, yes :)

mpblocky:
@dnechay fixed!

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode-handlers.ts
```diff
@@ -0,0 +1,32 @@
...
+  setIsDarkMode(matches);
+  if (matches) {
+    document.body.classList.add('dark-mode');
+  } else {
+    document.body.classList.remove('dark-mode');
+  }
+};
+
+export const runColorMode = (
+  fn: (matches: boolean) => void
```
dnechay, line 19: Tbh, the naming in this function is confusing, I would suggest to change it to
```
export const addColorSchemePrefsListener = (
  onChange: (prefersDark: boolean) => void
): (() => void) => {
  const query = window.matchMedia('(prefers-color-scheme: dark)');
  onChange(query.matches);

  const listener = (event: MediaQueryListEvent) => {
    onChange(event.matches);
  };
  query.addEventListener('change', listener);

  return () => {
    query.removeEventListener('change', listener);
  };
};
```

also it always returns `(() => void)`, no need in `undefined`

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode-handlers.ts
```diff
@@ -0,0 +1,32 @@
+import { hasColorMode } from './color-mode-settings';
+
+export const handleColorModeChange = (
```
dnechay, line 3: Sorry, I was wrong: now I see that this function depends on `setIsDarkMode` that is not some common function from settings, but state setter function. So it is better to declare it at the same place where `setIsDarkMode` defined - in `ColorModeProvider` itself for simplicity. So with `runColorMode` refactor mentioned below it will look like:

```
const [isDarkMode, setIsDarkMode] = useState<boolean>(isDarkColorMode());

const handleColorSchemePrefsChange = (prefersDarkScheme: boolean) => {
  if (hasColorMode()) {
    return;
  }

  setIsDarkMode(prefersDarkScheme);
  if (prefersDarkScheme) {
    document.body.classList.add('dark-mode');
  } else {
    document.body.classList.remove('dark-mode');
  }
};

useEffect(() => {
  const unsubscribe = addColorSchemePrefsListener(
    handleColorSchemePrefsChange
  );

  return () => {
    unsubscribe();
  };
}, []);
```

wdyt?

dnechay (commented):
Please resolve merge conflicts

packages/apps/human-app/frontend/src/shared/contexts/color-mode/color-mode-handlers.ts
mpblocky: Yes, that seems better :D I thought that was intentional :)

PR #3007, opened 2025-01-14
Hum 92 use register address

adrian-oleskiewicz:
## Issue tracking

HuM-92
## Context behind the change

Refactor hook to align with SRP

## How has this been tested?

checked if worker profile correctly uses the hook
## Release plan

normal deploy

N/A


packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/types.ts
```diff
@@ -0,0 +1,15 @@
+import type { ResponseError } from '@/shared/types/global.type';
+
+export interface RegisterAddressCallbacks {
```
mpblocky, line 3: same as above

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-mutation.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    const messageToSign = JSON.stringify(dataToSign);
+    const signature = await signMessage(messageToSign);
+
+    if (!signature) {
+      throw new Error(t('errors.unknown'));
+    }
+
+    await registerAddressRequest({ address, signature });
+    await refreshAccessTokenAsync({ authType: 'web2' });
```
mgacek-blockydevs, line 73: can we do both actions asynchronous using Promise.all()?

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-validation.ts
```diff
@@ -0,0 +1,14 @@
+import { useWalletConnect } from '@/shared/hooks/use-wallet-connect';
+
+export function useRegisterAddressValidation() {
```
mgacek-blockydevs, line 3: maybe this functionality should be implemented in `useWalletConnect()` hook?

mgacek-blockydevs: better place -> Provider

adrian-oleskiewicz: I will create separate ticket for this because I don't want to create conflicts in shared module 
https://silksoftwarehouse.atlassian.net/browse/HUM-168

adrian-oleskiewicz:
@dnechay ready to review

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/index.ts
```diff
@@ -0,0 +1,3 @@
+export { useRegisterAddress } from './use-register-address';
+export { useRegisterAddressMutation } from './use-register-address-mutation';
```
dnechay, line 2: `useRegisterAddressMutation` and `useRegisterAddressValidation` aren't used anywhere except `useRegisterAddress` and I'm not sure they will be.
Do you see a potential use case? If no - maybe we remove these exports, so when you import from `use-register-address` your IDE suggest only `useRegisterAddress`, not mutation or validation, wdyt?

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-validation.ts
```diff
@@ -0,0 +1,14 @@
...
+  const { address } = useWalletConnect();
+
+  const errors = {
```
dnechay, line 6: Every time `useRegisterAddressValidation` is called - it will return new `errors` object reference even if input and validation result do not change, hence anything that relies on `errors` reference (e.g. if `errors` is listed as dependencies in `useEffect` hook) won't be able to properly check if it's changed.

Probably `useMemo` should make the deal

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-mutation.ts
```diff
@@ -0,0 +1,98 @@
...
+export function useRegisterAddressMutation(
+  callbacks?: RegisterAddressCallbacks
+) {
+  const queryClient = useQueryClient();
+  const { user } = useAuthenticatedUser();
+  const { refreshAccessTokenAsync } = useAccessTokenRefresh();
+  const { address, signMessage } = useWalletConnect();
+
+  const mutationFn = async () => {
+    if (!address) {
```
dnechay, line 56: That's basically the same logic that is done in `useRegisterAddressValidation` and if we assume that `useRegisterAddressMutation` shouldn't be used outside of `useRegisterAddress` - no need to check if you have address here. Instead this check should be done in validation hook and validation hook should return error with i18n

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-mutation.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    const messageToSign = JSON.stringify(dataToSign);
+    const signature = await signMessage(messageToSign);
+
+    if (!signature) {
+      throw new Error(t('errors.unknown'));
+    }
+
+    await Promise.all([
```
dnechay, line 72: Please correct me if I'm wrong, but these two operations must be sequential, not parallel: `wallet_address` is part of JWT payload and is added there only after registered, so by making these parallel you add a race-condition

Might worth dropping a comment about it so developer from the future doesn't make the same

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-validation.ts
```diff
@@ -0,0 +1,14 @@
...
+    noAddress: !address,
```
dnechay, line 7: Why making `errors` object with specific properties instead of e.g. array with error messages?

dnechay:
@mpblocky ^

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/index.ts
mpblocky: Definitely!

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-mutation.ts
mpblocky: Good point, these also don't look like a heavy calculations, so they don't need this kind of optimization

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address/use-register-address-mutation.ts
mpblocky: The way types are built around `useWalletConnect` we cannot really remove that, because it returns: `address: string | undefined`

dnechay: @mpblocky does it make sense to move that validation to a separate hook then? Even if `useRegisterAddressValidation` is called before `useRegisterAddressMutation` - you still have to do implement its own validation.

If we want to simplify the code, it would be better to implement validation as a regular function and use it in `useRegisterAddressMutation`.

As a follow up to this, there will be no need to split `useRegisterAddress` to multiple files, just one file with the wrapped mutation hook

mpblocky: Yes, this is way better

dnechay: > can we get rid of the validation hook altogether?

Yes! You read between the lines ;)

mpblocky:
@dnechay ready!

packages/apps/human-app/frontend/src/modules/worker/hooks/use-register-address.ts
```diff
@@ -1,73 +1,107 @@
...
+  onError?: (error: ResponseError) => void | Promise<void>;
+}
+
+interface RegisterAddressParams {
+  address: string;
+  signature: string;
+}
+
+interface RegisterAddressResponse {
+  success: boolean;
```
dnechay, line 26: @mpblocky  That schema doesn't correspond to what we have on BE:
```
{
  "key": "string",
  "value": "string"
}
```
so because of that UI displays schema validation error after successful operation.

If we don't use the response somehow later it might worth to keep it `z.unknown()` or don't validate at all

dnechay (commented):
One comment about response validation, other than that - works and lgtm

mpblocky:
@dnechay all ready!

dnechay (approved):

PR #3009, opened 2025-01-14
HUM-93 - refactor email verification pages

adrian-oleskiewicz:
## Issue tracking

HUM-93
## Context behind the change

1. Refactor hook to align with SRP
2. Code split into hooks, components, types
3. Files renamed

## How has this been tested?

registered new user and checked both pages displaying correctly
## Release plan

normal deploy

N/A


packages/apps/human-app/frontend/src/modules/worker/types/email-verification.types.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from 'zod';
+
+export const tokenSchema = z.string().transform((value, ctx) => {
```
mpblocky, line 3: This file should be named schemas, not types

packages/apps/human-app/frontend/src/modules/worker/hooks/use-router-state.ts
```diff
@@ -0,0 +1,11 @@
+import { useLocationState } from '@/modules/worker/hooks/use-location-state';
+import { routerStateSchema } from '@/modules/worker/types/email-verification.types';
+
+export function useRouterState() {
```
mpblocky, line 4: Is this used only here? It looks like a shared hook

packages/apps/human-app/frontend/src/modules/worker/components/email-verification/email-verification-form-container.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <PageCard
+      alert={alertComponent}
+      cancelRouterPathOrCallback={handleCancel}
+      title="Verify Email"
+    >
+      <ResendVerificationEmailForm
+        methods={methods}
+        handleResend={handleResend}
+        email={routerState.email}
+        isAuthenticated={isAuthenticated}
```
mgacek-blockydevs, line 45: `isAuthenticated={!!user}`

adxable: eslint doesn't like !! and prefers Boolean()

adrian-oleskiewicz: eslint doesn't like !! and changes it to Boolean()

adrian-oleskiewicz:
@dnechay ready to review

dnechay (commented):
Please re-organize things according to their business logic purpose, not by their nature

dnechay:
@mpblocky ^

mpblocky:
@dnechay this one is ready

packages/apps/human-app/frontend/src/modules/worker/email-verification/components/email-verification-process.tsx
```diff
@@ -0,0 +1,28 @@
...
+  token: string;
+}
+
+export function EmailVerificationProcess({
+  token,
+}: Readonly<EmailVerificationProcessProps>) {
+  const { errorMsg, isEmailVerificationError, isEmailVerificationPending } =
+    useEmailVerification(token);
+
+  if (isEmailVerificationError && errorMsg) {
```
dnechay, line 19: What if it's errored, but there is no `errorMsg`? Probably in this case we still should show error

Also you use `getErrorMessageForError` in hook and here, makes sense to keep it only in JSX since `getErrorMessageForError` is purely for user-facing parts (which are components, not hooks)

packages/apps/human-app/frontend/src/modules/worker/email-verification/components/email-verification-process.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface EmailVerificationProcessProps {
+  token: string;
+}
+
+export function EmailVerificationProcess({
+  token,
+}: Readonly<EmailVerificationProcessProps>) {
+  const { errorMsg, isEmailVerificationError, isEmailVerificationPending } =
+    useEmailVerification(token);
```
dnechay, line 17: `useEmailVerification` logic can be part of this component because nothing specific happens there

packages/apps/human-app/frontend/src/modules/worker/email-verification/hooks/use-email-verification-token.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export function useEmailVerificationToken() {
+  const { field: token } = useLocationState({
+    schema: tokenSchema,
+    locationStorage: 'search',
+  });
+
+  return {
+    token,
+    isLoading: token === undefined,
```
dnechay, line 24: `useLocationState` is not something async, so adding `isLoading` here is not necessary and confusing.

Email verification logic process should look like:
1) page loads
2) we grab the token from query params
3) we run the verification - BE query to verify the token; since this is async action - it can trigger "loading" state
4) we show the verification result

So we can keep `useEmailVerificationToken` in separate file with its schema to ease reading of `EmailVerificationProcess` component, but not need to pass `isLoading`

packages/apps/human-app/frontend/src/modules/worker/email-verification/components/email-verification-form-container.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { methods, handleResend } = useResendEmail(routerState?.email ?? '');
+
+  const isAuthenticated = Boolean(user);
+
+  const handleCancel = () => {
+    signOut();
+    navigate(routerPaths.homePage);
+  };
+
+  if (!routerState?.email) {
```
dnechay, line 23: Could you please shed a light on this: why we show `page loader` in case there is no email in router state? Why not just redirect to home page or smth?

packages/apps/human-app/frontend/src/modules/worker/email-verification/pages/worker-email-verification-process.page.tsx
```diff
@@ -0,0 +1,13 @@
+import { PageCardLoader } from '@/shared/components/ui/page-card';
+import { EmailVerificationProcess } from '../components';
+import { useEmailVerificationToken } from '../hooks';
+
+export function WorkerEmailVerificationProcessPage() {
+  const { token, isLoading } = useEmailVerificationToken();
+
+  if (isLoading || !token) {
```
dnechay, line 8: Why showing `PageCardLoader` if no token? Won't it redirect to home page automatically?

dnechay (commented):
1. We can get rid of `WorkerEmailVerificationProcessPage` component: there is no need in `isLoading` state in `useEmailVerificationToken` hook, so this hook can be moved directly to `EmailVerificationProcess` component and that component becomes `EmailVerificationProcessPage`
2. No need to have `WorkerVerifyEmailPage` because `EmailVerificationFormContainer` acts exactly like the "page"

packages/apps/human-app/frontend/src/modules/worker/email-verification/pages/worker-email-verification-process.page.tsx
mpblocky: Yep, totally. There was a handover misunderstanding of what I had to do, sorry 😓

packages/apps/human-app/frontend/src/modules/worker/email-verification/components/email-verification-form-container.tsx
mpblocky: As mentioned [in here](https://github.com/humanprotocol/human-protocol/pull/3009#discussion_r1967473281)

packages/apps/human-app/frontend/src/modules/worker/email-verification/hooks/use-email-verification-token.ts
mpblocky: @dnechay this will make the code a bit more messy IMO as another hook depends on that token (which then calls the backend with that token). My proposition is to leave it as is and remove `isLoading` and `Loader` component in `worker-email-verification-process.page.tsx` and replace it with `PageCardError`. WDYT?

mpblocky:
@dnechay bumping [this comment](https://github.com/humanprotocol/human-protocol/pull/3009#discussion_r1967510767) as it might have been lost in the notifications 😅

packages/apps/human-app/frontend/src/modules/worker/email-verification/hooks/use-email-verification-token.ts
mpblocky: @dnechay Exactly, `WorkerEmailVerificationProcessPage` would check if token exists, display error if not and return `EmailVerificationProcess` if it exists. That way we do not have to make workarounds for required token in `useEmailVerification`

dnechay: @mpblocky sounds good!

mpblocky: @dnechay actually I found something. `useEmailVerificationToken` uses `useLocationState` which has default navigation path in case validation of the param failed, so we never reach the `pageCardError` but at the same time I can't do `token!` because lint does not let me. Should I leave it as is or disable the rule for that one line?

dnechay: Hm, that is an interesting case: actually `token` can be undefined because fields from location are loaded by `useEffect` hook that runs after initial render of the component, so:
1) initial render phase and `token` is undefined (that can cause some edge cases and expections)
2) `useEffect` in `useLocationState` called and it grabs `token` from location
3) component re-render with `token` as string

So we actually reach `pageCardError`

I would say that we will need to refactor `useLocationState` and its usages to avoid this implicit buggy behavior, also because sometimes you might want to use values from `useLocation`, but w/o doing a `navigate` and simply showing an error instead

packages/apps/human-app/frontend/src/modules/worker/email-verification/components/email-verification-process.tsx
```diff
@@ -0,0 +1,37 @@
...
+} from '@/shared/components/ui/page-card';
+import { getErrorMessageForError } from '@/shared/errors';
+import { useVerifyEmailQuery } from '../hooks';
+import { EmailVerificationSuccessMessage } from './email-verification-success-message';
+
+interface EmailVerificationProcessProps {
+  token: string;
+}
+
+function useEmailVerification(token: string) {
```
dnechay, line 13: IMO you don't need to go that granular and this code can be part of `EmailVerificationProcess`

packages/apps/human-app/frontend/src/modules/worker/email-verification/components/email-verification-form-container.tsx
```diff
@@ -0,0 +1,47 @@
...
+    return <Navigate to={routerPaths.homePage} />;
+  }
+
+  const alertComponent = methods.formState.isSubmitSuccessful ? (
```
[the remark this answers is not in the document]
mpblocky: Weirdly enough no, it works as intended with current state 👀 Will check it out

mpblocky:
@dnechay I fixed the email resend and redirect. I also improved notifications using the new system and fixed missing `p`

dnechay (approved):

PR #3010, opened 2025-01-14
HUM-95 - trnslation handling added

adrian-oleskiewicz:
## Issue tracking

HUM-95
## Context behind the change

 Implement internationalization (i18n)
## How has this been tested?

checked if page was deleted correctly
## Release plan

normal deploy

N/A


adrian-oleskiewicz:
@dnechay ready for review

dnechay (commented):
Could you please explain why we need this `ProtectedPage` component?
I'm not sure I understand its purpose, probably we can get rid of it at all

mpblocky:
@dnechay I removed this page completely as it was not used beside just displaying `protected page`

packages/apps/human-app/frontend/src/shared/i18n/en.json
```diff
@@ -13,6 +13,9 @@
     "jobs": "Tasks",
     "registrationInExchangeOracle": "Registration in Exchange Oracle"
   },
+  "protectedPage": {
```
dnechay, line 16: Should we remove this? Seems like `protectedPagesHeaders` is used instead

mpblocky: @dnechay Yes! I missed that one, thanks.

dnechay (approved):

PR #3011, opened 2025-01-14
[Subgraph] Add `name` and `category` to leader entity

### LINKED ISSUE #2989: [SDK] Add `name` and `category` to leaders
**Description**
Add 2 new fields to the Leaders entity in the subgraph:
- name: The value of this field should be fetched from the `name` key in the KVStore.
- category: The value of this field should be fetched from the `category` key in the KVStore. 
Additionally, update the SDK to include these fields when fetching leaders.

**Motivation**
Currently, the Leaders entity lacks a name field, which could be useful for better identification in the dashboard. Including the name field improves usability and clarity for users.
Category will help to identify what category of tasks can process the operator.

**Implementation details** 
1. Subgraph Updates

- Modify the Leaders entity schema to include a name and category fields.
- Update the indexing logic to fetch the value of the name key from the KVStore and populate this field in the Leaders entity.
- Update the indexing logic to fetch the value of the category key from the KVStore and populate this field in the Leaders entity.

2. SDK Updates

- Update the SDK's functions that fetch leaders to include the name and category fields in the response.
- Ensure compatibility with existing functionality to avoid breaking changes.

3. Define 2 categories in SDK as enum

- Machine learning
- Market making

leric7:
## Issue tracking

Closes #2989 

## Context behind the change

We need to keep track of `name`, and `category` information for the leaders stored on KVStore from the Subgraph and SDK, which will be used from the dashboard later on.

## How has this been tested?

- Unit tests are updated for subgraph & SDK changes.

## Release plan

N/A

N/A


packages/sdk/typescript/human-protocol-sdk/src/enums.ts
```diff
@@ -24,3 +24,8 @@ export enum OrderDirection {
   ASC = 'asc',
   DESC = 'desc',
 }
+
+export enum LeaderCategory {
```
portuu3, line 28: Lets use a constat for it like in https://github.com/humanprotocol/human-protocol/blob/b95124839b6f0153784afef5a3399a0c8dc6c157/packages/sdk/typescript/human-protocol-sdk/src/constants.ts#L303-L304 

Also use `machine_learning` and `market_making` as values

portuu3 (changes requested):
Define category types for python sdk

portuu3 (approved):

portuu3 (approved):

PR #3014, opened 2025-01-16
Hum 96 registration page refactor

adrian-oleskiewicz:
## Issue tracking

HUM-96
## Context behind the change

Code split into hook, component and page view to separate logic 

## How has this been tested?

manually
## Release plan

normal deploy

N/A

### DIFF
--- a/packages/apps/human-app/frontend/src/modules/worker/components/registration/registration-form.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/components/registration/registration-form.tsx
@@ -24,6 +24,10 @@ export function RegistrationForm({
   const { t } = useTranslation();
   const methods = useRegistrationForm();
 
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    void methods.handleSubmit(onSubmit)(event);
+  };
+
   return (
     <>
       <Button onClick={onInstructionsClick} fullWidth variant="contained">
@@ -31,7 +35,7 @@ export function RegistrationForm({
       </Button>
       <Box>{t('worker.registrationInExchangeOracle.completeMessage')}</Box>
       <FormProvider {...methods}>
-        <form onSubmit={(event) => void methods.handleSubmit(onSubmit)(event)}>
+        <form onSubmit={handleSubmit}>
           <Stack alignItems="center" spacing={2}>
             <FormCaptcha error={error} name="h_captcha_token" />
             <Button
--- a/packages/apps/human-app/frontend/src/modules/worker/components/registration/registration-form.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/components/registration/registration-form.tsx
@@ -1,0 +1,51 @@
+import { Box, Stack } from '@mui/material';
+import { FormProvider } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
+import { FormCaptcha } from '@/shared/components/h-captcha';
+import { Button } from '@/shared/components/ui/button';
+import { type RegistrationInExchangeOracleDto } from '@/modules/worker/services/registration-in-exchange-oracles';
+import { useRegistrationForm } from '@/modules/worker/hooks/use-registration-form';
+
+interface RegistrationFormProps {
+  hasClickedRegistrationLink: boolean;
+  isLoading: boolean;
+  error: unknown;
+  onInstructionsClick: () => void;
+  onSubmit: (data: RegistrationInExchangeOracleDto) => void;
+}
+
+export function RegistrationForm({
+  hasClickedRegistrationLink,
+  isLoading,
+  error,
+  onInstructionsClick,
+  onSubmit,
+}: RegistrationFormProps) {
+  const { t } = useTranslation();
+  const methods = useRegistrationForm();
+
+  return (
+    <>
+      <Button onClick={onInstructionsClick} fullWidth variant="contained">
+        {t('worker.registrationInExchangeOracle.instructionsButton')}
+      </Button>
+      <Box>{t('worker.registrationInExchangeOracle.completeMessage')}</Box>
+      <FormProvider {...methods}>
+        <form onSubmit={(event) => void methods.handleSubmit(onSubmit)(event)}>
+          <Stack alignItems="center" spacing={2}>
+            <FormCaptcha error={error} name="h_captcha_token" />
+            <Button
+              disabled={!hasClickedRegistrationLink}
+              fullWidth
+              loading={isLoading}
+              type="submit"
+              variant="contained"
+            >
+              {t('worker.registrationInExchangeOracle.completeButton')}
+            </Button>
+          </Stack>
+        </form>
+      </FormProvider>
+    </>
+  );
+}
--- a/packages/apps/human-app/frontend/src/modules/worker/hooks/use-registration-form.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/hooks/use-registration-form.tsx
@@ -1,0 +1,20 @@
+/* eslint-disable camelcase -- ... */
+import { useForm } from 'react-hook-form';
+import { useParams } from 'react-router-dom';
+import { zodResolver } from '@hookform/resolvers/zod';
+import {
+  type RegistrationInExchangeOracleDto,
+  registrationInExchangeOracleDtoSchema,
+} from '@/modules/worker/services/registration-in-exchange-oracles';
+
+export function useRegistrationForm() {
+  const { address: oracleAddress } = useParams<{ address: string }>();
+
+  return useForm<RegistrationInExchangeOracleDto>({
+    defaultValues: {
+      oracle_address: oracleAddress,
+      h_captcha_token: '',
+    },
+    resolver: zodResolver(registrationInExchangeOracleDtoSchema),
+  });
+}
--- a/packages/apps/human-app/frontend/src/modules/worker/hooks/use-registration-page-logic.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/hooks/use-registration-page-logic.tsx
@@ -1,0 +1,76 @@
+import { useState, useEffect, useCallback } from 'react';
+import { useNavigate, useParams } from 'react-router-dom';
+import {
+  type RegistrationInExchangeOracleDto,
+  useExchangeOracleRegistrationMutation,
+} from '@/modules/worker/services/registration-in-exchange-oracles';
+import { useRegisteredOracles } from '@/shared/contexts/registered-oracles';
+import { useGetOracles } from '@/modules/worker/services/oracles';
+import { routerPaths } from '@/router/router-paths';
+
+export function useRegistrationPageLogic() {
+  const navigate = useNavigate();
+  const { address: oracleAddress } = useParams<{ address: string }>();
+  const [hasClickedRegistrationLink, setHasClickedRegistrationLink] =
+    useState(false);
+
+  const { data: oraclesData } = useGetOracles();
+  const oracleData = oraclesData?.find(
+    (oracle) => oracle.address === oracleAddress
+  );
+
+  const { registeredOracles, setRegisteredOracles } = useRegisteredOracles();
+
+  const {
+    mutate: userRegistrationMutate,
+    isPending: isRegistrationInExchangeOraclePending,
+    error: registrationInExchangeOracleError,
+  } = useExchangeOracleRegistrationMutation();
+
+  const handleInstructionsLinkClick = () => {
+    window.open(oracleData?.registrationInstructions ?? '', '_blank');
+    setHasClickedRegistrationLink(true);
+  };
+
+  const handleRegistrationComplete = (
+    data: RegistrationInExchangeOracleDto
+  ) => {
+    userRegistrationMutate(data, {
+      onSuccess() {
+        if (oracleAddress) {
+          setRegisteredOracles((prev) =>
+            prev ? [...prev, oracleAddress] : [oracleAddress]
+          );
+        }
+        navigateToJobs();
+      },
+    });
+  };
+
+  const navigateToJobs = useCallback(() => {
+    navigate(`${routerPaths.worker.jobs}/${oracleAddress ?? ''}`, {
+      state: { oracleAddress },
+    });
+  }, [navigate, oracleAddress]);
+
+  useEffect(() => {
+    if (oracleData === undefined) {
+      navigate(routerPaths.worker.jobsDiscovery);
+    }
+  }, [oracleData, navigate]);
+
+  useEffect(() => {
+    if (registeredOracles?.find((a) => a === oracleAddress)) {
+      navigateToJobs();
+    }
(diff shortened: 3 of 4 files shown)


adrian-oleskiewicz:
@dnechay ready to review

dnechay:
@mpblocky ^

mpblocky:
@dnechay ready!

packages/apps/human-app/frontend/src/modules/worker/registration/registration.page.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export function RegistrationPage() {
+  const { t } = useTranslation();
+  const {
+    hasViewedInstructions,
+    handleInstructionsView,
+    handleRegistration,
+    isRegistrationPending,
+    registrationError,
+  } = useOracleRegistrationFlow();
```
dnechay, line 14: All items from `useOracleRegistrationFlow` are being used only by `RegistrationForm` component. I'm not 100% sure, but maybe it makes sense to just use this hook in `RegistrationForm` component then, wdyt?

packages/apps/human-app/frontend/src/modules/worker/registration/index.ts
```diff
@@ -0,0 +1 @@
+export * from './registration.page';
```
dnechay, line 1: Maybe let's change the module name to `oracle-registration`? Right now it looks a bit confusing, like it's related to sign-up process

packages/apps/human-app/frontend/src/modules/worker/registration/hooks/use-oracle-registration-flow.tsx
```diff
@@ -0,0 +1,50 @@
...
+      navigateToJobs();
+    }
+  }, [isAlreadyRegistered, navigateToJobs]);
+
+  const handleInstructionsViewWithData = () => {
+    handleInstructionsView(oracleData?.registrationInstructions ?? '');
+  };
+
+  return {
+    oracleData,
```
dnechay, line 43: `oracleData` is not used after export, so might be removed

packages/apps/human-app/frontend/src/modules/worker/registration/hooks/use-registration-page-logic.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useNavigate, useParams } from 'react-router-dom';
+import {
+  type RegistrationInExchangeOracleDto,
+  useExchangeOracleRegistrationMutation,
+} from '@/modules/worker/services/registration-in-exchange-oracles';
+import { useRegisteredOracles } from '@/shared/contexts/registered-oracles';
+import { useGetOracles } from '@/modules/worker/services/oracles';
+import { routerPaths } from '@/router/router-paths';
+
+export function useRegistrationPageLogic() {
```
dnechay, line 11: This hooks is not used anywhere, probably leftover from some intermediate steps

packages/apps/human-app/frontend/src/modules/worker/registration/hooks/use-oracle-registration-flow.tsx
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }, [oracleData, navigateToDiscovery]);
+
+  useEffect(() => {
+    if (isAlreadyRegistered) {
+      navigateToJobs();
+    }
+  }, [isAlreadyRegistered, navigateToJobs]);
+
+  const handleInstructionsViewWithData = () => {
```
dnechay, line 38: Maybe wrap it to `useCallback` as well?

dnechay (commented):
Seems working
Few cleanup comments

packages/apps/human-app/frontend/src/modules/worker/registration/index.ts
mpblocky: Yes, this sounds better, I was confused too

packages/apps/human-app/frontend/src/modules/worker/registration/registration.page.tsx
mpblocky: Feels like we are making the component less dumb, but looking from another perspective, this place will not change often, so I agree :)

dnechay: My rationale is:
- if you just map values from another hook - don't split it, use it directly
- if you have some custom logic over using another hook - estimate it's complexity; if simple - maybe keep as is, if difficult - move it out of component
- if you have exactly the same flow somewhere else - DRY, move it out

mpblocky: Yes, sounds reasonable!

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/hooks/use-oracle-registration-flow.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const {
+    handleRegistration,
+    isRegistrationPending,
+    registrationError,
+    isAlreadyRegistered,
+  } = useOracleRegistration(oracleAddress);
+  const { hasViewedInstructions, handleInstructionsView } =
+    useOracleInstructions();
+
+  useEffect(() => {
```
dnechay, line 26: Let's move these parts with navigation back to the `page` component.

Right now when you check `registration.page.tsx` code - you see only registration form there and it creates wrong perception that registration form and logic are always there

Would be nice to have it explicit there that if no oracle data or already registered - we navigate away (`isAlreadyRegistered` can be moved out of `useOracleRegistration` hook). This way `useOracleRegistrationFlow` hook will keep only logic related to the actual registration steps

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/hooks/use-oracle-registration-flow.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { navigateToJobs, navigateToDiscovery } =
+    useOracleNavigation(oracleAddress);
+  const {
+    handleRegistration,
+    isRegistrationPending,
+    registrationError,
+    isAlreadyRegistered,
+  } = useOracleRegistration(oracleAddress);
+  const { hasViewedInstructions, handleInstructionsView } =
+    useOracleInstructions();
```
dnechay, line 24: Maybe we pass `oracleData?.registrationInstructions ?? ''` here instead of `handleInstructionsView`? This way it will be consistent with `useOracleNavigation` style and also handler is tied to oracle data

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/registration-form.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { Button } from '@/shared/components/ui/button';
+import { HCaptchaForm } from '@/shared/components/hcaptcha/h-captcha-form';
+import { useOracleRegistrationFlow, useRegistrationForm } from './hooks';
+
+export function RegistrationForm() {
+  const { t } = useTranslation();
+  const methods = useRegistrationForm();
+  const {
+    hasViewedInstructions,
+    handleInstructionsView: onInstructionsClick,
```
dnechay, line 13: Generally, naming convention in react is:
- `onEventName` - for prop names of component, that access some event handler
- `handleEventName` - for handlers of some events that you pass to `onEventName` prop

So it's fine to keep `handleInstructionsView` (and others) name as is and simply use it below

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/hooks/use-registration-form.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { useForm } from 'react-hook-form';
+import { useParams } from 'react-router-dom';
+import { zodResolver } from '@hookform/resolvers/zod';
+import {
+  type RegistrationInExchangeOracleDto,
+  registrationInExchangeOracleDtoSchema,
+} from '@/modules/worker/services/registration-in-exchange-oracles';
+
+export function useRegistrationForm() {
+  const { address: oracleAddress } = useParams<{ address: string }>();
```
dnechay, line 11: We don't have some complex logic in this hook and simply wrap `useForm`, so I would suggest moving it to the `RegistrationForm` component in order to easily read it and see what are default values, how fields are named (and if they are consistent with what specified in the component)

dnechay (commented):
Left some suggestions

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/hooks/use-oracle-registration-flow.tsx
mpblocky: I will have to move the `oracleData` var into it as the callback relies on it. Will do that in separate commit to let You check that one easily!

mpblocky:
@dnechay ready for review

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/registration.page.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const { t } = useTranslation();
+
+  const { data } = useGetOracles();
+  const { address } = useParams<{ address: string }>();
+  const isAlreadyRegistered = useIsAlreadyRegistered(address);
+
+  const { navigateToJobs, navigateToDiscovery } = useOracleNavigation(address);
+
+  useEffect(() => {
+    if (data === undefined) {
```
dnechay, line 20: It should check if there is oracle data for specific oracle, not if we load all of them (see original page component)

Taking into account that `useGetOracles` is also a hook with data loading, should we consider `isLoading` state somehow? I suppose that `data` might be `undefined` while we loading data (unless it uses it from cache)

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/hooks/use-oracle-instructions.tsx
```diff
@@ -0,0 +1,17 @@
+import { useState, useCallback } from 'react';
+import { useGetOracles } from '../../services/oracles';
+
+export function useOracleInstructions(oracleAddress?: string) {
```
dnechay, line 4: It can accept just `registrationInstructionsUrl` as a param, which is passed from where `useOracleInstructions` is used - in `useOracleRegistrationFlow`

dnechay (commented):
Looks better!

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/registration-form.tsx
```diff
@@ -0,0 +1,67 @@
...
+      // eslint-disable-next-line camelcase
+      oracle_address: oracleAddress,
+      // eslint-disable-next-line camelcase
+      h_captcha_token: '',
+    },
+    resolver: zodResolver(registrationInExchangeOracleDtoSchema),
+  });
+}
+
+export function RegistrationForm() {
```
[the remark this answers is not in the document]
dnechay: So the idea for this and above comments is next:
1) `RegistrationPage` components gets address from url and uses `useGetOracles`
2) If no data for oracle found - page is in favor of navigation
3) if oracle data is found - page juggles registration
3a) if already registered - skip
3b) if not registered - render registration form that should just use the same (already loaded) data

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/registration.page.tsx
mpblocky: Yes, I think this is a good way to handle that

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/registration-form.tsx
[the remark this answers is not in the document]
mpblocky: Yep, sounds good and prevents from accidental usage :)

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/registration-form.tsx
```diff
@@ -0,0 +1,71 @@
...
+    resolver: zodResolver(registrationInExchangeOracleDtoSchema),
+  });
+}
+
+export function RegistrationForm({
+  address,
+  oracleInstructions,
+}: Readonly<{
+  address: string | undefined;
+  oracleInstructions: string | URL | null | undefined;
```
mpblocky, line 30: @dnechay is using `undefined` for address good here? I would do just string ( with checks for `undefined` and `''` in the page component), but the linter forbids from using non-null assertion `address!` and I don't know if I can override the rule here.

dnechay: @mpblocky great suggestion on leaving it as just string!

In order to remove `undefined` from possible types in the parent components and not using `non-null assertion` you can go one of two ways:
1) use type assertion: `address as string` in case you are sure it's not `undefined`
2) use type predicates as type guards ([ref](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)) or something similar

Let's try to make this suggestion live!

mpblocky: Sure, I wanted to make sure that is desired :)

mpblocky:
@dnechay ready to check

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/registration.page.tsx
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }, [oracleData, address, navigateToDiscovery]);
+
+  useEffect(() => {
+    if (isAlreadyRegistered) {
+      navigateToJobs();
+    }
+  }, [isAlreadyRegistered, navigateToJobs]);
+
+  if (!isAddress(address)) {
```
dnechay, line 36: We kinda have the same logic twice: here and in the hook with `navigateToDiscovery` above.

What if we change this to use `<Navigate repalce ...`? Then we will have it in one place and clear and not losing type guards for further usage

Same approach can be applied for `navigateToJobs`

In general, it makes more sense to use `Navigate` instead of hooks when you don't have to render your component and just redirect, and `useNavigate` when you need programmatic control, so I would suggest to stick to that approach in the future PRs

dnechay (commented):
LGTM, one suggestion on navigation

mpblocky:
@dnechay ready

dnechay (approved):
LGTM

However the "redirect" behavior is a bit weird both for this PR and previous implementation on staging, e.g. if I go to `worker/registration-in-exchange-oracle/0x54b681e95ba0c02cf6ac2adaad9f99fd33a3a092` (Fortune oracle) when I already registered in it - it redirects me to `job-discovery` page instead of `worker/jobs`. When I try `worker/registration-in-exchange-oracle` - it redirects me to `profile` page.

It's not related to this PR and might require some effort on changing the way how we work we react router and routes that load data, so please create a separate Jira for future refactoring about this

PR #3016, opened 2025-01-17
Fix/locking

ivhus:
## Issue tracking

https://github.com/orgs/humanprotocol/projects/28/views/1?pane=issue&itemId=83071490&issue=humanprotocol%7CDevOps-Design%7C195 and Slack messaging with @Dzeranov 

## Context behind the change

Mitigating random failing deploys when different docker-compose services being deployed more or less at the same time

## How has this been tested?

Manual GHA triggering

## Release plan

Not much, just develop → main should be enough

No significant risk, worst case revert the commit :D

### DIFF
--- a/.github/workflows/cd-cvat-exchange-oracle.yaml
+++ b/.github/workflows/cd-cvat-exchange-oracle.yaml
@@ -18,6 +18,9 @@ permissions:
 jobs:
   pull-request-check:
     runs-on: ubuntu-latest
+    concurrency:
+      group: docker-compose-operations
+      cancel-in-progress: false  # Ensure the job waits instead of canceling others
     if: ${{ github.event_name == 'pull_request' }}
     steps:
       - name: Checkout
--- a/.github/workflows/cd-cvat-exchange-oracle.yaml
+++ b/.github/workflows/cd-cvat-exchange-oracle.yaml
@@ -14,13 +14,13 @@ on:
 permissions:
   id-token: write
   contents: read
+concurrency:
+  group: docker-compose-operations
+  cancel-in-progress: false  # Ensure the workflow waits instead of canceling others
 
 jobs:
   pull-request-check:
     runs-on: ubuntu-latest
-    concurrency:
-      group: docker-compose-operations
-      cancel-in-progress: false  # Ensure the job waits instead of canceling others
     if: ${{ github.event_name == 'pull_request' }}
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


dnechay (commented):
Might we need concurrency check for the whole workflow? Even though `trigger-build-and-deploy` relies on `pull-request-check`, the latter won't run in case of `workflow_dispatch` trigger, will be skipped and there will be multiple runs of `trigger-build-and-deploy`

ivhus:
> Might we need concurrency check for the whole workflow? Even though `trigger-build-and-deploy` relies on `pull-request-check`, the latter won't run in case of `workflow_dispatch` trigger, will be skipped and there will be multiple runs of `trigger-build-and-deploy`

@dnechay Yeah valid point, made it workflow-level

dnechay (commented):
IIUC, group key is shared across all workflows in all owner's account repos ([ref](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#overview)).

So taking into account that for both oracles we have `docker-compose-operations`, it means that they will be deployed consequently, only one worker at a time. Not sure if that was intended behavior or there was a need to avoid parallel runs of the same workflow, so please clarify that.

Ideally, I wouldn't use such a general key and rename it to something like:
- `${{ github.workflow }}` in case we want to have only one run of specific workflow across all branches
- `${{ github.workflow }}-${{ github.ref }}` if unique per branch
- `cvat-oracles-ci` if only one active oracle deployment
- and etc.

Other than that - lgtm. @Dzeranov, @ivhus what is the expected behavior of CI here?

ivhus:
> IIUC, group key is shared across all workflows in all owner's account repos ([ref](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#overview)).
> 
> So taking into account that for both oracles we have `docker-compose-operations`, it means that they will be deployed consequently, only one worker at a time. Not sure if that was intended behavior or there was a need to avoid parallel runs of the same workflow, so please clarify that.
> 
> Ideally, I wouldn't use such a general key and rename it to something like:
> - `${{ github.workflow }}` in case we want to have only one run of specific workflow across all branches
> - `${{ github.workflow }}-${{ github.ref }}` if unique per branch
> - `cvat-oracles-ci` if only one active oracle deployment
> - and etc.
> 
> Other than that - lgtm. @Dzeranov, @ivhus what is the expected behavior of CI here?

exchange and recording oracles share same docker-compose and machine here

so the intention for the pr is that workflows for exchange and for recording oracles queue with each other

but the branch yes, better have it distinguished

How about
docker-compose-oracles-${{ github.ref }}
?

dnechay (approved):
LGTM

If we need branch in name or not - you know better :)

PR #3023, opened 2025-01-17
HUM-97: refactor sign in page

adrian-oleskiewicz:
## Issue tracking

HUM-97
## Context behind the change

Split page component into components, hooks, utils

## How has this been tested?

manually
## Release plan

normal deploy

N/A

### DIFF
--- a/packages/apps/human-app/frontend/src/modules/worker/components/sign-in/sign-in-form.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/components/sign-in/sign-in-form.tsx
@@ -1,0 +1,87 @@
+/* eslint-disable camelcase -- ...*/
+import { FormProvider, useForm } from 'react-hook-form';
+import { Box, Grid, Typography } from '@mui/material';
+import { zodResolver } from '@hookform/resolvers/zod';
+import { useTranslation } from 'react-i18next';
+import { Link } from 'react-router-dom';
+import { Input } from '@/shared/components/data-entry/input';
+import { Button } from '@/shared/components/ui/button';
+import { Password } from '@/shared/components/data-entry/password/password';
+import { routerPaths } from '@/router/router-paths';
+import { FormCaptcha } from '@/shared/components/h-captcha';
+import { type SignInDto } from '@/modules/worker/services/sign-in/types';
+import { signInDtoSchema } from '@/modules/worker/services/sign-in/schema';
+import { useResetMutationErrors } from '@/shared/hooks/use-reset-mutation-errors';
+
+interface SignInFormProps {
+  onSubmit: (data: SignInDto) => void;
+  error?: unknown;
+  isLoading?: boolean;
+  resetMutation: () => void;
+}
+
+export function SignInForm({
+  onSubmit,
+  error,
+  isLoading,
+  resetMutation,
+}: SignInFormProps) {
+  const { t } = useTranslation();
+
+  const methods = useForm<SignInDto>({
+    defaultValues: {
+      email: '',
+      password: '',
+      h_captcha_token: '',
+    },
+    resolver: zodResolver(signInDtoSchema),
+  });
+
+  useResetMutationErrors(methods.watch, resetMutation);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+    void methods.handleSubmit(onSubmit)(event);
+  };
+
+  return (
+    <FormProvider {...methods}>
+      <form onSubmit={handleSubmit}>
+        <Grid container gap="1.5rem">
+          <Input
+            fullWidth
+            label={t('worker.signInForm.fields.email')}
+            name="email"
+          />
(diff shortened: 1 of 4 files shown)


adrian-oleskiewicz:
@dnechay ready to review

packages/apps/human-app/frontend/src/modules/worker/components/sign-in/sign-in-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { useResetMutationErrors } from '@/shared/hooks/use-reset-mutation-errors';
+
+interface SignInFormProps {
+  onSubmit: (data: SignInDto) => void;
+  error?: unknown;
+  isLoading?: boolean;
+  resetMutation: () => void;
+}
+
+export function SignInForm({
```
dnechay, line 23: Is this component reusable in `worker` module or is it only for `SignIn` page?
If it's only for sing in - then probably we shouldn't have it in `worker/components`

packages/apps/human-app/frontend/src/modules/worker/hooks/use-sign-in.tsx
```diff
@@ -0,0 +1,30 @@
+import { useEffect } from 'react';
+import { useSignInMutation } from '@/modules/worker/services/sign-in/sign-in';
+import { browserAuthProvider } from '@/shared/contexts/browser-auth-provider';
+import { type SignInDto } from '@/modules/worker/services/sign-in/types';
+
+export function useSignIn() {
```
dnechay, line 6: Is this hook reusable across `worker` module or only for `sign-in` view?

packages/apps/human-app/frontend/src/modules/worker/utils/formatted-sign-in-error-message.ts
```diff
@@ -0,0 +1,10 @@
+import { t } from 'i18next';
+import { FetchError } from '@/api/fetcher';
+
+export function formattedSignInErrorMessage(
```
dnechay, line 4: Similar to [this comment](https://github.com/humanprotocol/human-protocol/pull/3029#discussion_r1926944226)

packages/apps/human-app/frontend/src/modules/worker/components/sign-in/sign-in-form.tsx
```diff
@@ -0,0 +1,87 @@
...
+      email: '',
+      password: '',
+      h_captcha_token: '',
+    },
+    resolver: zodResolver(signInDtoSchema),
+  });
+
+  useResetMutationErrors(methods.watch, resetMutation);
+
+  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
```
dnechay, line 42: You probably need to create submit handler in different way:
```suggestion
  const handleSubmit = methods.handleSubmit(onSubmit);
```
Or for simplicity you can pass  `methods.handleSubmit(onSubmit)` to `onSubmit` directly

dnechay (commented):
In "How has this been tested?" please don't just write "manually", write about some step you took to test the code or test-cases you accomplished, e.g.:
- [x] attempted to sign in with invalid password - error is shown
or
- [x] success sign-in case

dnechay:
@mpblocky ^

packages/apps/human-app/frontend/src/modules/worker/components/sign-in/sign-in-form.tsx
mpblocky: Hey, I tried to do this (and similar things) but types here are not compatibile: `Promise-returning function provided to attribute where a void return was expected.`. 

`methods.handleSubmit` returns `Promise<void>` while form's `onSubmit` expects `FormEventHandler<T> | undefined;`


I can do something like:
```
  const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
    event.preventDefault();
    void methods.handleSubmit(onSubmit)(event);
  };

  return (
    <FormProvider {...methods}>
      <form onSubmit={handleSubmit}>
``` 

but it's not a big change, WDYT?

dnechay: @mpblocky weird peculiarity, thanks for the info

Then it's fine to keep it as is, no need to add `event.preventDefault()` because it's already handled by react-hook-form ([ref](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L1206); otherwise we would get page reloaded as a default form submit action)

mpblocky:
@dnechay resolved conflicts :)

dnechay (commented):
@mpblocky 
By analogy with `sign-up` module, let's move `sing-in.page` to `signin/worker` module, so everything is there?

mpblocky:
@dnechay rearranged files

dnechay (approved):

PR #3029, opened 2025-01-20
Hum 98 sign up refactor

adrian-oleskiewicz:
## Issue tracking

HUM-98
## Context behind the change

Sign up page refactor
split code into smaller hooks, utils (SRP)

## How has this been tested?

- Tried to create an worker account with existing signatures - an error is shown as expected
- New worker account creation - verify email view is shown as expected
- Password and confirm password validation works correctly.
## Release plan

normal deploy

N/A


packages/apps/human-app/frontend/src/modules/worker/utils/formatted-sign-up-error-message.tsx
```diff
@@ -0,0 +1,10 @@
+import { t } from 'i18next';
+import { FetchError } from '@/api/fetcher';
+
+export function formattedSignUpErrorMessage(
+  unknownError: unknown
+): string | undefined {
+  if (unknownError instanceof FetchError && unknownError.status === 409) {
+    return t('worker.signUpForm.errors.emailTaken');
```
mgacek-blockydevs, line 8: I think that error message could pass as prop or `t` function should be passed by prop, because this code is in isolation or part of a library

adrian-oleskiewicz:
@dnechay ready to review

packages/apps/human-app/frontend/src/modules/worker/views/sign/sign-up.page.tsx
```diff
@@ -46,29 +32,18 @@ export function SignUpWorkerPage() {
...
-    reset: signUpWorkerMutationReset,
-  } = useSignUpMutation();
-
-  useResetMutationErrors(methods.watch, signUpWorkerMutationReset);
+  useResetMutationErrors(methods.watch, reset);
 
-  const handleWorkerSignUp = (data: SignUpDto) => {
-    signUpWorkerMutate(omit(data, ['confirmPassword']));
+  const handleSignupError = (unknownError: unknown) => {
+    return formattedSignUpErrorMessage(unknownError, t);
```
dnechay, line 38: You can move implementation of `formattedSignUpErrorMessage` into this handler, no need to have a separate utility function for that because such "formatting" logic is purely for sign-up ui and highly-likely won't be used anywhere else.

Also a side-note: passing `t` as a param is over-engineering here because only purpose of `formattedSignUpErrorMessage` is to prepare i18n message for the user and you probably won't use different tools for that you might want to abstract as param

packages/apps/human-app/frontend/src/modules/worker/hooks/use-sign-up.tsx
```diff
@@ -0,0 +1,31 @@
+import { useEffect } from 'react';
+import omit from 'lodash/omit';
+import { browserAuthProvider } from '@/shared/contexts/browser-auth-provider';
+import type { SignUpDto } from '@/modules/worker/services/sign-up';
+import { useSignUpMutation } from '@/modules/worker/services/sign-up';
+
+export function useSignUp() {
```
dnechay, line 7: You don't have to move it to it's own file: the only place where it's used is `sign up` page and it's fine to have it declared there:
- it won't be used anywhere else, so by moving it to `hooks` you only inflate this folder; `hooks` should contain only something that is shared for `worker` module
- in case I need to debug code or make some change to signup, it's easier to have only one file opened in editor, navigate and gain context from there

SRP is not about splitting components/classes/whatever to small pieces of code into different files, it's about separating things with different purpose ([this article worth reading](https://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html))

packages/apps/human-app/frontend/src/modules/worker/views/sign/sign-up.page.tsx
```diff
@@ -77,7 +52,7 @@ export function SignUpWorkerPage() {
       <FormProvider {...methods}>
         <form
           onSubmit={(event) => {
```
dnechay, line 54:
```suggestion
          onSubmit={methods.hadnleSubmit(signUp)}
```

packages/apps/human-app/frontend/src/modules/worker/hooks/use-sign-up.tsx
adrian-oleskiewicz: I think separating the logic into a custom hook aligns better with React best practices and has some solid long-term advantages. By extracting the logic, the component stays cleaner and more focused on the UI, which makes it easier to maintain. So I would prefer to keep both files under same folder even if it is not reused by any other component atm.

packages/apps/human-app/frontend/src/modules/worker/views/sign/sign-up.page.tsx
adrian-oleskiewicz: It causing a lint error
Zrzut ekranu 2025-01-27 o 16 19 07

I implemented this way:
```
const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
   void methods.handleSubmit(signUp)(event);
};
  ```

File: packages/apps/human-app/frontend/src/modules/worker/views/sign/sign-up.page.tsx (hunk 2 above)
Comment (adrian-oleskiewicz) [reply packages/apps/human-app/frontend/src/modules/worker/views/sign/sign-up.page.tsx]: Good point, I removed this unnecessary file.

File: packages/apps/human-app/frontend/src/modules/worker/hooks/use-sign-up.tsx (hunk 3 above)
Comment (adrian-oleskiewicz) [reply packages/apps/human-app/frontend/src/modules/worker/hooks/use-sign-up.tsx]: @dnechay Please take a look again and review :)

Comment (dnechay) [reply packages/apps/human-app/frontend/src/modules/worker/hooks/use-sign-up.tsx]: @adrian-oleskiewicz 

> I think separating the logic into a custom hook aligns better with React best practices and has some solid long-term advantages

That's for sure! I mean't that you don't have to move this custom hook to a separate file because it's used only in one component :)

But as long as they live together in the same `signup/worker` module - I'm fine with it :)

File: packages/apps/human-app/frontend/src/router/routes.tsx [hunk 5]
```diff
@@ -2,7 +2,7 @@ import type { RouteProps } from 'react-router-dom';
 import { t } from 'i18next';
 import { ProtectedPage } from '@/modules/worker/views/protected/protected.page';
 import { SignInWorkerPage } from '@/modules/worker/views/sign/sign-in.page';
-import { SignUpWorkerPage } from '@/modules/worker/views/sign/sign-up.page';
+import { SignUpWorkerPage } from '@/modules/signup/worker/views/sign-up-worker.page';
```
dnechay, line 5: @adrian-oleskiewicz 

Will it make sense to have `modules/signup/worker/index.ts` file, where you re-export what you need from this module?
This way importing should be simpler and clear + can refactor `signup/worker` module internally w/o changing any imports that depend on it, wdyt?

dnechay (commented):
LGTM overall, one question about `index.ts`

packages/apps/human-app/frontend/src/router/routes.tsx
dnechay: @mpblocky ^

mpblocky: Yes, sounds good to me, additionally it gives opportunity for later on: imports can be even restricted to imports only from current dir or `index.ts` files, which will prevent any unwanted imports while still allowing to safely break up code in smaller pieces. Will do!

mpblocky: FYI we must be wary of growing index files however, as [this issue](https://vite.dev/guide/performance.html#avoid-barrel-files) is still present in vite

mpblocky:
@dnechay ready

packages/apps/human-app/frontend/src/router/routes.tsx
dnechay: @mpblocky thanks for the info!

That issue in vite is interesting, however I'm inclined to think that it mostly related to things like `utils` (or libs like `lodash`) and in this case we should be smart when grouping them, so indeed can import specific utils (e.g. `shared/utils/date` instead of `shared/utils`) and get benefits of tree-shaking. In case we are thinking about real components, even if you import only, let's say, `SignUp page`, your browser has to load everything that is in `sign-up` folder because it's used for sign-up page.

So it's fine to have `index.ts` files in case you group your things properly (i.e. they are used/loaded together)

packages/apps/human-app/frontend/src/modules/signup/worker/index.ts
```diff
@@ -0,0 +1 @@
+export * from './views/sign-up-worker.page';
```
dnechay, line 1: For further refactoring would be nice to export from `index.ts` only the things that should be accessible outside, not everything

dnechay (approved):

packages/apps/human-app/frontend/src/modules/signup/worker/index.ts
dnechay: @mpblocky

PR #3042, opened 2025-01-24
[Reputation Oracle] Refactor `User` module error handling

Dzeranov:
## Issue tracking

Subtask of a #2928 

## Context behind the change

1. Changed error handling in `User` module.
2. Removed unused code.
3. Refactored tests (removed some mocks, etc.)
4. Partially removed `Credentials` functionality. Will be deleted as a part of #3045 

### _**DO NOT MERGE TILL THIS IS CHECKED**_
- [ ] Deploy this branch to staging and test after the initial review.

## Release plan

N/A

N/A


packages/apps/reputation-oracle/server/src/modules/user/user.controller.ts
```diff
@@ -170,7 +178,16 @@ export class UserController {
   public async prepareSignature(
     @Body() data: PrepareSignatureDto,
   ): Promise<SignatureBodyDto> {
-    return await this.userService.prepareSignatureBody(data.type, data.address);
+    let nonce;
+    if (data.type === SignatureType.SIGNIN) {
+      nonce = (await this.userRepository.findOneByAddress(data.address))?.nonce;
```
dnechay, line 183: @portuu3 do you see any reason why we store this in DB and not just generate on UI?

IMU we can get rid of this `prepare-signature` completely on RepO, UI/user needs to generate some "nonce", sign it and send together with signature, then RepO is able to verify it

packages/apps/reputation-oracle/server/src/modules/user/user.error.filter.ts
```diff
@@ -0,0 +1,44 @@
...
+  catch(exception: UserControllerError, host: ArgumentsHost) {
+    const ctx = host.switchToHttp();
+    const response = ctx.getResponse<Response>();
+    const request = ctx.getRequest<Request>();
+    let status = HttpStatus.BAD_REQUEST;
+
+    if (exception instanceof DuplicatedWalletAddressError) {
+      status = HttpStatus.CONFLICT;
+    } else if (exception instanceof InvalidWeb3SignatureError) {
+      status = HttpStatus.CONFLICT;
```
dnechay, line 33: Maybe we change it to `400`? I checked the UI and don't see us relying on 409 for signature flow

packages/apps/reputation-oracle/server/src/modules/user/user.service.ts
```diff
@@ -189,33 +151,31 @@ export class UserService {
     data.address = data.address.toLowerCase();
```
dnechay, line 151: Would be better to not change the input, but have a separate variable and use it:
```
const addressToRegister = data.address.toLowerCase();
```

packages/apps/reputation-oracle/server/src/common/guards/signature.auth.spec.ts
```diff
@@ -79,18 +78,34 @@ describe('SignatureAuthGuard', () => {
     });
 
     it('should throw unauthorized exception if signature is not verified', async () => {
-      (verifySignature as jest.Mock).mockReturnValue(false);
+      (verifySignature as jest.Mock).mockReturnValueOnce(false);
```
dnechay, line 81: Previous implementation of `verifySignature` couldn't return `false`, it could only throw if not verified, so probably we can remove this mock

packages/apps/reputation-oracle/server/src/common/guards/signature.auth.spec.ts
```diff
@@ -79,18 +78,34 @@ describe('SignatureAuthGuard', () => {
...
 
     it('should throw unauthorized exception for unrecognized oracle type', async () => {
       mockRequest.originalUrl = '/some/random/path';
-      await expect(guard.canActivate(context as any)).rejects.toThrow(
-        new ControlledError('Unauthorized', HttpStatus.UNAUTHORIZED),
-      );
+      try {
+        await guard.canActivate(context as any);
+      } catch (error) {
+        expect(error).toBeInstanceOf(HttpException);
```
dnechay, line 101: If you do testing using `try..catch` block, you should do something like:
```
let thrownError: any;
try {
  await someService.logicToTest();
} catch (error) {
  thrownError = error;
}

expect(thrownError).toBeDefined();
expect(thrownError.message).toBe('expected message');
```

Otherwise there is a risk that the method you test executes successfully and test case silently passes

packages/apps/reputation-oracle/server/src/modules/auth/auth.error.filter.ts
```diff
@@ -9,16 +9,16 @@ import { Request, Response } from 'express';
...
+} from './auth.error';
 
 type AuthControllerError =
   | AuthError
-  | DuplicatedUserError
+  | DuplicatedUserEmailError
   | InvalidOperatorSignupDataError;
 
-@Catch(AuthError, DuplicatedUserError, InvalidOperatorSignupDataError)
+@Catch(AuthError, DuplicatedUserEmailError, InvalidOperatorSignupDataError)
```
dnechay, line 21: Auth service now also throws `DuplicatedUserAddressError`, so we should catch it here as well

packages/apps/reputation-oracle/server/src/common/guards/signature.auth.spec.ts
Dzeranov: Agree on removing this mock but for a different reason :)

dnechay (approved):

dnechay (approved):
Tested ✔️

PR #3043, opened 2025-01-24
[Subgraph] Clean KVStore fee values

flopez7:
## Issue tracking
None

## Context behind the change
An error was thrown in subgraph when 

## How has this been tested?
modified and run tests for checking new conditional

## Release plan
Deploy new Subgraph version

## Potential risks; What to monitor; Check if '' as fee value is handled properly

### DIFF
--- a/packages/sdk/typescript/subgraph/src/mapping/KVStore.ts
+++ b/packages/sdk/typescript/subgraph/src/mapping/KVStore.ts
@@ -79,7 +79,11 @@ export function handleDataSaved(event: DataSaved): void {
   if (key == 'role') {
     leader.role = event.params.value;
   } else if (key == 'fee') {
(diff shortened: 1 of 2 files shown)


packages/sdk/typescript/subgraph/src/mapping/KVStore.ts
```diff
@@ -79,7 +79,11 @@ export function handleDataSaved(event: DataSaved): void {
   if (key == 'role') {
     leader.role = event.params.value;
   } else if (key == 'fee') {
-    leader.fee = BigInt.fromString(event.params.value);
+    if (event.params.value !== '') {
```
portuu3, line 82: Let's set null, also the same for the rest of keys.

portuu3 (changes requested):

portuu3 (approved):

Issue #2912, opened 2024-12-13
[Typescript SDK] Invalid date format when querying GraphQL


dnechay:
**Describe the bug**
In some places (e.g. in methods of `StatisticsClient`), when working with dates we convert them to seconds like `filter.from.getTime() / 1000`. If the date value passed to method is not prepared to be rounded to seconds already (i.e. have `...000` at the end), then SDK gets decimal value and makes query with it, leading to an error:
```
An error occurred while interacting with Ethereum: Invalid value provided for argument `from`: Float(1731492900.038)
```

**To Reproduce**
1. Go to SDK's `examples/statistics`
2. Adjust `getEscrowStatistics` to be called with full date, like `new Date()`
3. Run the example

**Expected behavior**
Consumer of SDK's API shouldn't care about date format if they pass date to SDK methods. Under the hood SDK should prepare the date by rounding it. Rounding must be consistent for all dates.

> [!NOTE]
As part of this issue, it's expected to revisit and fix all SDK's methods that behave in the same way and their consumers. Not just `StatisticsClient`

PR #3044, opened 2025-01-24
[SDK] fix: date rounding issues

flopez7:
## Issue tracking
#2912

## Context behind the change
1. Updated all instances of getTime() / 1000 to use Math.floor to ensure dates are rounded to the nearest second.
2. Updated tests to round also expected values

## How has this been tested?
Ran examples and tests with different dates.

## Release plan
Deploy new SDK version

## Potential risks; What to monitor; None


dnechay (approved):
LGTM

Does it make sense to have some internal util that is used in all these repeated places (but not in unit tests)?

portuu3 (dismissed):

dnechay (approved):
Nice!

PR #3057, opened 2025-01-29
[Reputation Oracle] refactor: get rid of controlled error in escrow completion

dnechay:
## Issue tracking
Part of #2928 

## Context behind the change
Removing `ControlledError` from escrow-completion module as per refactoring plan.
WIth this module it was decided to not create custom error classes because there is no need to control execution flow based on specific error type, it's enough to just throw and log details in place.

## How has this been tested?
- [x] just replacing existing errors, so lint + build
- [x] existing unit tests pass 

## Release plan
Simply merge.

Can't think of any

### DIFF
--- a/packages/apps/reputation-oracle/server/src/common/constants/errors.ts
+++ b/packages/apps/reputation-oracle/server/src/common/constants/errors.ts
@@ -1,15 +1,4 @@
 /**
- * Represents error messages related to escrow completion.
- */
-export enum ErrorEscrowCompletion {
-  NotFound = 'Escrow completion not found',
-  NotCreated = 'Escrow completion has not been created',
-  PendingProcessingFailed = 'Failed to process pending escrow completion',
-  PaidProcessingFailed = 'Failed to process paid escrow completion',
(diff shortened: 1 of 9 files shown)


packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.error.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseError } from '../../common/errors/base';
+
+type ErrorMeta = Record<string, unknown>;
+
+export class EscrowCompletionError extends BaseError {
+  public readonly meta: ErrorMeta;
+
+  constructor(message: string, meta: ErrorMeta = {}) {
+    super(message);
+    this.meta = meta;
```
dnechay, line 10: Later when we use structured logging we will be able to just:
```
logger.error('Failed to process escrow completion, {
  errorDetail: error.message,
  escrowCompletionId,
  ...error.meta
})
```

Dzeranov (approved):
lgtm

Issue #3039, opened 2025-01-23
[Job Launcher] Failing to transform case of request body properties for "failed" webhooks


dnechay:
**Describe the bug**
Job Launcher (probably other services that use similar transform logic) fails to transform properties in request body of "failed" webhook. Assumed that it might be related not only to failed webhooks, but to any request with similar strufcture.

**To Reproduce**
1. Go to JL swagger, find `webhook` endpoint
2. Use next data as example:
```
human-signature: 0xbddf449ec6acf6f8350573256dfd7994f300f9a3920dab27b4d60082d8b11309558a519e32a0646629e6cac981c56708a51a52db0fd7caa93de92e925d50c8ff1c
// request body
{
  "chain_id": 80002,
  "escrow_address": "0x621b57Fb82896a9e733AD6CCCb4EBE465aE59745",
  "event_type": "job_creation_failed",
  "event_data": {
    "reason": "Test failure reason"
  }
}
```
3. Send request
4. See 500 error

**Expected behavior**
Requests processed w/o error related to request body transformations

**Screenshots**
<img width="1131" alt="Image" src="https://github.com/user-attachments/assets/89046b51-414c-4ff1-8a7f-f50d8c7668fc" />

PR #3060, opened 2025-01-30
[Job Launcher] Fixed case transformation issue for request body properties in "Failed" webhooks

eugenvoronov:
## Issue tracking
The issue was tracked on GitHub under issue [Job Launcher] Failing to transform case of request body properties for "failed" webhooks
[#3039](https://github.com/humanprotocol/human-protocol/issues/3039)

## Context behind the change
Job Launcher (and other services with similar transformation logic) was failing to processing nested objects.
The issue occurs because `instance[property]` might be undefined when `bodyOrQuery[property]` is an object, causing an error when trying to access `instance[property][nestedKey]`.
- Next services were updated: Fortune, Exchange Oracle, Recording Oracle, Human App, Job Launcher, Reputation Oracle

## Release plan
This change can be deployed as part of the regular deployment cycle.


dnechay (commented):
Fix lgtm. Is it possible to add some simple unit test to handle this case?

eugenvoronov:
> Fix lgtm. Is it possible to add some simple unit test to handle this case?
> 
> @eugenvoronov

I think I can add some unit test for this case, I will check

packages/apps/fortune/exchange-oracle/server/src/modules/job/job.controller.ts
```diff
@@ -95,7 +94,6 @@ export class JobController {
   @AllowedRoles([AuthSignatureRole.Worker])
   async solveJob(
     @Body() solveJobDto: SolveJobDto,
-    @Headers(HEADER_SIGNATURE_KEY) signature: string,
```
dnechay, line 98: Test case for it failing because `solveJob` called with mocked signature

packages/apps/fortune/exchange-oracle/server/src/modules/webhook/webhook.controller.ts
```diff
@@ -58,7 +58,6 @@ export class WebhookController {
   })
   public async processWebhook(
     @Body() body: WebhookDto,
-    @Headers(HEADER_SIGNATURE_KEY) _: string,
```
dnechay, line 61: Test case for it failing because `processWebhook` called with mocked signature

packages/apps/job-launcher/server/src/common/guards/signature.auth.ts
```diff
@@ -18,6 +18,7 @@ export class SignatureAuthGuard implements CanActivate {
     const request = context.switchToHttp().getRequest();
 
     const data = request.body;
+    console.log(request.headers);
```
dnechay, line 21:
```suggestion
```

packages/apps/fortune/exchange-oracle/server/src/common/interceptors/transform-enum.interceptor.spec.ts
```diff
@@ -161,4 +161,38 @@ describe('TransformEnumInterceptor', () => {
     });
     expect(callHandler.handle).toHaveBeenCalled();
   });
+
+  it('should return bodyOrQuery if instance is not an object', () => {
+    // Test with `null` as the instance
+    let result = interceptor['lowercaseEnumProperties'](
```
dnechay, line 167: Why not `await interceptor.intercept(executionContext, callHandler).toPromise();` as in other tests and comparing that body is transformed properly for objects w/o nested values?

eugenvoronov: Because this test is directly calling the private method `lowercaseEnumProperties`, rather than executing the full interceptor flow.

eugenvoronov: @dnechay

dnechay (approved):
Nice

Issue #2982, opened 2025-01-09
[Job Launcher] Payments refactor


portuu3:
**Description**
Currently, the system only supports funding an escrow using HMT tokens. To enhance flexibility, we need to allow users to fund escrows using other token balances (e.g., USD) when the HMT balance is insufficient. This involves updating the logic to support cross-token funding and reflecting this correctly in the database, specifically in the payments table.

**Implementation details**
Add the token fund to escrow table to reflect in the database the token used for funding the escrow when we enable the ability to use other tokens than HMT.

Add the ability to use other tokens balance to fund an escrow, for example imagine a user has these balances:
- 100 HMT (rate 0.05 USD)
- 50 USD 

The user wants to launch an escrow and fund it with 200 HMT. In payments table we should reflect it as:

| Amount | Token |  USD_rate  | Job Id | 
|---|---|---|---|
| -100 | HMT | 0.05 | 1 |
| -5 | USD  | 1 | 1 |

PR #3069, opened 2025-02-04
Balance refactor

portuu3:
## Issue tracking
Closes https://github.com/humanprotocol/human-protocol/issues/2982

## Context behind the change
- Fixes possible negative balances
- Allows to choose the token for funding the escrow.
- Use utils for handling operations like mul, div, etc

## How has this been tested?
- [x] Created a job paid with fiat USD and funded with HMT
- [x] Created a job paid with HMT and funded with HMT
- [x] Funded the account with crypto
- [x] Funded the account with fiat
- [x] Refund after cancellation
- [x] Created a job using balance

Unit tests partially updated. Job service needs a refactor so tests for this service will be implemented from scratch.

## Release plan
Includes a migration to add `token` to jobs table

Monitor payments table
Rollback: undo last migration


packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -413,21 +410,40 @@ export class PaymentService {
...
+          : await this.rateService.getRate(token, FiatCurrency.USD);
+      totalBalance += mul(tokenAmountSum, rate);
     }
 
     return totalBalance;
   }
 
-  public async createRefundPayment(dto: PaymentRefundCreateDto) {
-    const rate = await this.rateService.getRate(TokenId.HMT, Currency.USD);
+  public async getUserBalanceByCurrency(
```
Dzeranov, line 425: I think that we're not going to support anything except for usd in the nearest future so we can remove this method and simplify a bit our business logic.

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
```diff
@@ -226,7 +223,7 @@ export class PaymentService {
     }
 
     // Record the payment details in the system.
-    const rate = await this.rateService.getRate(currency, Currency.USD);
+    const rate = await this.rateService.getRate(currency, FiatCurrency.USD);
```
Dzeranov, line 226: I think that we can remove it since we're using only usd for fiat.

packages/apps/job-launcher/server/src/modules/payment/payment.service.ts
Dzeranov: Also we'll be able to get of paymentCurrencyRate etc.

packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -947,18 +928,9 @@ export class JobService {
...
-      paymentEntity.currency = JobCurrency.USD;
-      paymentEntity.rate = 1;
-    } else {
-      paymentEntity.amount = -tokenTotalAmount;
-      paymentEntity.currency = TokenId.HMT;
-      paymentEntity.rate = div(1, rate);
-    }
+    paymentEntity.amount = -totalPaymentAmount; // In the currency used for the payment.
+    paymentEntity.currency = dto.paymentCurrency;
+    paymentEntity.rate = paymentCurrencyRate;
```
Dzeranov, line 933: I think it makes more sense to store an actual rate conversion (for example usd -> hmt (escrow fund token))

PR #3082, opened 2025-02-07
[Reputation Oracle] feat: custom json logger

### LINKED ISSUE #3065: [Reputation Oracle] Logging system
**Description**
Need to create a custom logger module with the following specs:
1. Log `info`, `error`, `debug` messages.(latter for local development and stg env).
2. Input parameters: `message: string` , `meta: Record<string, unknown>`. In case of an error, it should be joined type with `Error`.
3. Ability to gather and provide some general context (Service name, shortened commit SHA, timestamp, node env)
4. Add child creation support.

**Motivation**
To have a unified and structured logging that could be easily integrated with for example Datadog.
Also, as a next step (once this logger is implemented and tested), this module should be shared across all services in monorepo.

**Implementation details** 
Covered in Description section.

dnechay:
## Issue tracking
Closes #3065 

## Context behind the change
Nest logger is not able to log message in JSON format (tried different ways from their docs - no success) + we want to have specific structure of our logs and maybe some custom configurations in the future. So it was decided to write custom logger implementation for Reputation Oracle, that we later can move to shared package in monorepo and use across other services.

> [!NOTE]
> Revision of log messages and their meta is left for later.

Implementation rationale:
- we didn't got path of "Nest module" (aka injected module) in order to be easily able to log from some kind of "util" functions or some other places that don't work with decorators and/or Nest injection system, so we are framework agnostic
- we need to have a "wrapper" that implements `LoggerService` for Nest in order to use it as internal logger for the framework; as alternative - we can disable nest logging completely, but we might miss some internal errors
- `error` serialization is made to common format for error logging convenience
- even though logger interface is written in TypeScript, we still have some internal mechanism to "adjust" log record if `message` or `meta` of unexpected type (e.g. if somebody uses logger as `logger.error(error)` or `logger.error('debug', [1,2,3])`), so we are conscious about production log structure, but in case logger is misused it won't break the app

Samples of JSON log message:
<details>

<summary>Info log</summary>

```json
{
  "level": "info",
  "environment": "local",
  "service": "reputation-oracle",
  "context": "CronJobService",
  "message": "Pending incoming webhooks START",
  "timestamp": 1738937400037
}
```
Screenshot 2025-02-07 at 17 12 14

</details>

<details>

<summary>Error log</summary>

```json
{
  "level": "error",
  "environment": "local",
  "service": "reputation-oracle",
  "context": "AuthControllerErrorsFilter",
  "error": {
    "kind": "DuplicatedUserEmailError",
    "message": "The email you are trying to use already exists. Please check that the email is correct or use a different email.",
    "stack": "DuplicatedUserEmailError: The email you are trying to use already exists. Please check that the email is correct or use a different email.\n    at AuthService.signup (/usr/src/app/packages/apps/reputation-oracle/server/dist/src/modules/auth/auth.service.js:67:19)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at async AuthJwtController.signup (/usr/src/app/packages/apps/reputation-oracle/server/dist/src/modules/auth/auth.controller.js:34:9)",
    "email": "hp-annotator-0@mailinator.com"
  },
  "message": "Auth error",
  "timestamp": 1738937201388
}

```
Screenshot 2025-02-07 at 17 11 31

</details>

## How has this been tested?
Run application locally:
- [x] make sure nest logs its startup information as expected in pretty format
- [x] trigger some endpoints with invalid params, make sure errors are logged properly
- [x] use `logger.info` in some places, make sure it logs as expected
- [x] make `pretty: false` when creating logger instance; make sure that logs are written in JSON

## Release plan
1. Update `NODE_ENV` variable to correspond to environment
2. Simple deployment

Should be none. We simply log to stdout.


packages/apps/reputation-oracle/server/src/logger/index.ts
```diff
@@ -0,0 +1,25 @@
+import { createLogger } from './pino-logger';
+import NestLogger from './nest-logger';
+import { LogLevel } from './types';
+
+const environment = process.env.NODE_ENV || 'local';
```
Dzeranov, line 5: We're not actively using NODE_ENV but I think it might be good to have a enum with all possible values instead of hardcoded strings. This is not necessarily for this PR, just to keep in mind.

Dzeranov (commented):
Could you please add a usage example to the index file in a logger folder? 
Other than that looks good to me, great job.

@portuu3 please take a look.

Dzeranov (approved):

PR #3102, opened 2025-02-18
fix: broken dependency causing Docker builds to fail

dnechay:
## Issue tracking
N/A

## Context behind the change
https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1128
After recent deps update we got one dependency with the issue, so rolling it back.
Also refactored Docker file for CVAT tests bloackchain node to be build from the pulled code, not `develop`

## How has this been tested?
- [x] `docker-compose -f docker-compose.test.yml up blockchain-node --build` successful
- [x] running any service in `scripts/cvat` (local docker setup) successful 

## Release plan
Simply merge

No

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4260,12 +4260,12 @@
     undici "^6.11.1"
 
 "@openzeppelin/upgrades-core@^1.32.2", "@openzeppelin/upgrades-core@^1.41.0":
-  version "1.42.1"
-  resolved "https://registry.yarnpkg.com/@openzeppelin/upgrades-core/-/upgrades-core-1.42.1.tgz#a2784e8d9c09f4a79b7e5cbb11933062ad709835"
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -4260,12 +4260,12 @@
     undici "^6.11.1"
 
 "@openzeppelin/upgrades-core@^1.32.2", "@openzeppelin/upgrades-core@^1.41.0":
-  version "1.42.1"
-  resolved "https://registry.yarnpkg.com/@openzeppelin/upgrades-core/-/upgrades-core-1.42.1.tgz#a2784e8d9c09f4a79b7e5cbb11933062ad709835"
-  integrity sha512-8qnz2XfQrco8R8u9NjV+KiSLrVn7DnWFd+3BuhTUjhVy0bzCSu2SMKCVpZLtXbxf4f2dpz8jYPQYRa6s23PhLA==
+  version "1.41.0"
```
portuu3, line 4263: @dnechay I have checked to remove yarn.lock and regenerate it, and it gets back to version 1.42.1.
I think it can be solved if we use resolutions in the package.json of the root folder, adding `"@openzeppelin/upgrades-core": "1.41.0" `

Dzeranov (approved):

PR #3103, opened 2025-02-19
deps: bump node to 22 w/ dep updates

dnechay:
## Issue tracking
N/A

## Context behind the change
We got an [error from "Dependency Review" check](https://github.com/humanprotocol/human-protocol/actions/runs/13409932772) about `parse-duration` package vulnerability. In order to fix that we can't just use resolutions because fix is in major version, so we need to bump parent package - `@graphprotocol/graph-cli` - and that requires to use Node@20

Taking into account that Node@18 is not LTS anymore starting April 2025, decided to switch to current LTS - 22.

Also in scope of this PR fixed other resolutions to use proper versions of packages in lock file: previously we were skipping some major release for fixes, which is not correct and might have implicit issues.

## How has this been tested?
- [x] CI actions
- [x] changed `NODE_VERSION` for staging services on Render and rebuild + deploy, make sure they keep working fine
- [x] run `scripts/cvat` docker setup locally, make sure it's working e2e

## Release plan
- [x] Update `NODE_VERSION` on Render Staging
- [x] Update Node.js version in Vercel to 22
- [ ] Update `NODE_VERSION` on Render Production
- [ ] Update references on Node version in docs

That something is missing in release plan for specific env. We should notice it during the build phase since we have packages that require Node@20 now so `yarn install` will fail if invalid Node version provided. As a solution - provide the correct one


Dzeranov (approved):
lgtm
@portuu3 please take a look.

portuu3 (approved):

PR #3110, opened 2025-02-21
[HUM-170]: fix: error notification is displayed twice when should be once

mpblocky:
## Issue tracking

Closes: HUM-170

## Context behind the change

- Added wallet connect initialization check
- Relaxed `isReady` condition as the `mutation.data` could be empty

## How has this been tested?

- tested the expected behaviour on production build

## Release plan

normal

N/A

### DIFF
--- a/packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
@@ -14,7 +14,7 @@ import { useAuthenticatedUser } from '@/modules/auth/hooks/use-authenticated-use
 export function WorkerProfilePage() {
   const { user } = useAuthenticatedUser();
   const isMobile = useIsMobile();
-  const { isConnected } = useWalletConnect();
+  const { isConnected, initializing } = useWalletConnect();
   const { showNotification } = useNotification();
 
   const setNotifications = () => {
@@ -28,9 +28,9 @@ export function WorkerProfilePage() {
   };
 
   useEffect(() => {
-    setNotifications();
+    if (!initializing) setNotifications();
     // eslint-disable-next-line react-hooks/exhaustive-deps -- call this once
-  }, [isConnected]);
+  }, [isConnected, initializing]);
 
   return (
     <Paper
--- a/packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
@@ -18,17 +18,18 @@ export function WorkerProfilePage() {
   const { showNotification } = useNotification();
 
   const setNotifications = () => {
-    if (user.wallet_address) {
-      return;
+    if (initializing) return;
+
+    if (!isConnected || !user.wallet_address) {
+      showNotification({
+        type: TopNotificationType.WARNING,
+        message: t('worker.profile.topNotifications.completeSteps'),
+      });
     }
-    showNotification({
-      type: TopNotificationType.WARNING,
-      message: t('worker.profile.topNotifications.completeSteps'),
-    });
   };
 
   useEffect(() => {
-    if (!initializing) setNotifications();
+    setNotifications();
     // eslint-disable-next-line react-hooks/exhaustive-deps -- call this once
   }, [isConnected, initializing]);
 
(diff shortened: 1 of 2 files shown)


mpblocky:
@dnechay fix ready :)

packages/apps/human-app/frontend/src/shared/contexts/wallet-connect/wallet-connect.tsx
```diff
@@ -91,8 +91,7 @@ export function WalletConnectProvider({
     [web3ProviderMutation.data]
   );
 
-  const isReady =
```
dnechay, line 94: Can this change cause side-effects in other code parts that rely on `web3ProviderMutation.data`?

I'm not deep into this context, but looks like doing some regression tests of anything related won't hurt

packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
```diff
@@ -28,9 +28,9 @@ export function WorkerProfilePage() {
   };
 
   useEffect(() => {
-    setNotifications();
+    if (!initializing) setNotifications();
     // eslint-disable-next-line react-hooks/exhaustive-deps -- call this once
-  }, [isConnected]);
+  }, [isConnected, initializing]);
```
dnechay, line 33: Could you please refactor this effect to be something like:

```
if (initializing) return;

if (!isConnected || !user.wallet_address) {
  showNotification...
}
```

Right now it's a bit tricky to understand in what cases notification is shown

packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
```diff
@@ -14,23 +14,24 @@ import { useAuthenticatedUser } from '@/modules/auth/hooks/use-authenticated-use
 export function WorkerProfilePage() {
   const { user } = useAuthenticatedUser();
   const isMobile = useIsMobile();
-  const { isConnected } = useWalletConnect();
+  const { isConnected, initializing } = useWalletConnect();
   const { showNotification } = useNotification();
 
   const setNotifications = () => {
```
dnechay, line 20: Why not having the code of this function inline in the hook?

packages/apps/human-app/frontend/src/shared/contexts/wallet-connect/wallet-connect.tsx
dnechay: @mpblocky deployed this PR to preview, operator sign in and sign up don't work for this branch (works fine on staging)

packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
mpblocky: Oh, this looked like a preference :) Changed it!

mpblocky:
image

@dnechay I added this notification as well, becuase there was just a misleading "cannot sign message".

I'm looking into the operator problem

packages/apps/human-app/frontend/src/modules/worker/components/profile/profile.page.tsx
dnechay: Might be 😅

The perception of  `setNotifications` is that you might use it somewhere in click handler or business logic, but we actually use it only it hook and use data from other hooks, not passing some arguments, so it makes more sense to keep its code where it takes place

dnechay:
> I added this notification as well, becuase there was just a misleading "cannot sign message" when user had invalid network selected in the web3 provider.

@mpblocky great addition, thank you!

> I'm looking into the operator problem, where exactly?

On the main page, regular "Operator Sign In" button doesn't work for the account I have on STG. Same for "Sign Up". I suppose other functionality might be affected as well (if any)

packages/apps/human-app/frontend/src/shared/contexts/wallet-connect/wallet-connect.tsx
mpblocky: Yes, will investigate that - relaxing the condition fixed the issue with notifications, but might have broke something else. I will check that

mpblocky:
@dnechay thanks, I will investigate and check other places that could get affected

mpblocky:
@dnechay reverting the changes in wallet connect context were sufficient, so no regression tests are needed as changes in worker profile were enough to fix the notification problem :)

dnechay (approved):

PR #3114, opened 2025-02-24
fix: allow governance execution without dao spoke contract

leric7:
## Context behind the change

While Governance is designed to support cross-chain voting, we still want to support voting only on hub chain.
Current vote collection flow is only working when there are spoke contracts, and wormhole relayer is properly configured.
This changes allows us to finish collection phase only on hub chain after the vote ends, when there are no dao spoke contracts.

## How has this been tested?

I added a new test case for governance contract to execute the proposal, without dao spoke contracts deployed.


packages/core/contracts/governance/MetaHumanGovernor.sol
```diff
@@ -7,6 +7,7 @@ import '@openzeppelin/contracts/governance/extensions/GovernorVotes.sol';
 import '@openzeppelin/contracts/governance/extensions/GovernorVotesQuorumFraction.sol';
 import '@openzeppelin/contracts/governance/extensions/GovernorTimelockControl.sol';
 import '@openzeppelin/contracts/utils/Address.sol';
+import 'hardhat/console.sol';
```
portuu3, line 10: is this import actually necessary?

portuu3 (approved):

Issue #3105, opened 2025-02-19
[Reputation Oracle] NDA


portuu3:
**Description**
Workers should sign an NDA to protect requesters datasets

**Implementation details** 
- Add `nda_signed` field to `users` table. This field will contain the url of the last nda signed.
- Use an env variable to store the last nda url.
- During auth verify that `nda_signed` is equal to the env variable value and add `nda_signed` = true | false to jwt.
- Create nda endpoint to get last nda url (env value)
- Create sign nda endpoint (payload url)

PR #3132, opened 2025-02-27
[Reputation Oracle] NDA

flopez7:
## Issue tracking
#3105

## Context behind the change
add NDA module and related functionality

## How has this been tested?
Launched locally. 
Created one new user. 
Signed the NDA

## Release plan
Add `NDA_URL` in Reputation Oracle environment

None


Dzeranov:
@flopez7 please, rebase develop branch to resolve conflicts.

packages/apps/reputation-oracle/server/src/modules/nda/nda.module.ts
```diff
@@ -0,0 +1,12 @@
+import { Module } from '@nestjs/common';
+import { NDAController } from './nda.controller';
+import { NDAService } from './nda.service';
+import { UserModule } from '../user/user.module';
+import { UserRepository } from '../user/user.repository';
+
+@Module({
+  imports: [UserModule],
+  controllers: [NDAController],
+  providers: [NDAService, UserRepository],
```
dnechay, line 10:
```suggestion
  providers: [NDAService],
```
`UserRepository` is provider by `UserModule` import

packages/apps/reputation-oracle/server/src/modules/nda/nda.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  @ApiOperation({
+    summary: 'Sign NDA',
+    description:
+      'Signs the NDA with the provided URL. The URL must match the latest NDA URL.',
+  })
+  @ApiBody({ type: NDASignatureDto })
+  @ApiResponse({
+    status: 200,
```
dnechay, line 61: `@Post` defaults to `201`, so you need to add `@HttpCode(200)` decorator

packages/apps/reputation-oracle/server/src/modules/nda/nda.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+    summary: 'Retrieves latest NDA URL',
+    description:
+      'Retrieves the latest NDA URL that users must sign to join the oracle',
+  })
+  @ApiResponse({
+    status: 200,
+    description: 'URL retrieved successfully',
+    type: String,
+  })
+  @ApiResponse({
```
dnechay, line 45: Let's please remove this api response: we decided to not document `401` since it's common for api

packages/apps/reputation-oracle/server/src/modules/nda/nda.controller.ts
```diff
@@ -0,0 +1,78 @@
...
+    description:
+      'Signs the NDA with the provided URL. The URL must match the latest NDA URL.',
+  })
+  @ApiBody({ type: NDASignatureDto })
+  @ApiResponse({
+    status: 200,
+    description: 'NDA signed successfully',
+    type: String,
+  })
+  @ApiResponse({
```
dnechay, line 65: Let's remove this 401 as well

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('signNDA', () => {
+    const user: any = {
+      id: 1,
+      email: faker.internet.email(),
+      password: 'password',
```
dnechay, line 40: Let's remove `password` since it's not used.

Also please use faker to generate `id` and avoid using `any`, you can use `Pick<UserEntity, 'id' | 'email' | 'ndaSigned`

Ideally we just implement `generateUser` in `test/fixtures`, but you can leave this part for refactoring phase, where having `UserEntity` type usage will help us to find all places that need updates

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    service = module.get<NDAService>(NDAService);
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('signNDA', () => {
+    const user: any = {
```
dnechay, line 37: Maybe we generate new user before each test? This way we have independent test runs even if we forget to change something

dnechay (commented):
Looks good in general

Left some suggestions

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.ts
```diff
@@ -0,0 +1,28 @@
...
+    private readonly userRepository: UserRepository,
+    private readonly authConfigService: AuthConfigService,
+  ) {}
+
+  async signNDA(user: UserEntity, nda: NDASignatureDto) {
+    const ndaUrl = this.authConfigService.latestNdaUrl;
+    if (nda.url !== ndaUrl) {
+      throw new NDAError(NDAErrorMessage.INVALID_NDA, user.id);
+    }
+    if (user.ndaSigned === ndaUrl) {
```
Dzeranov, line 20: IMO we can just return OK instead of an error in this case.

packages/apps/reputation-oracle/server/src/modules/user/user.entity.ts
```diff
@@ -48,4 +48,7 @@ export class UserEntity extends BaseEntity implements IUser {
     (userQualification) => userQualification.user,
   )
   public userQualifications: UserQualificationEntity[];
+
+  @Column({ type: 'varchar', nullable: true })
+  public ndaSigned?: string;
```
Dzeranov, line 53: I'd rename `ndaSigned` to something like `ndaSignedUrl` or something similar because `ndaSigned` name makes me think that we store a boolean type.

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      providers: [
+        NDAService,
+        { provide: UserRepository, useValue: mockUserRepository },
+        { provide: AuthConfigService, useValue: mockAuthConfigService },
+      ],
+    }).compile();
+
+    service = module.get<NDAService>(NDAService);
+
+    user = {
```
portuu3, line 33: this could be out of beforeAll

packages/apps/reputation-oracle/server/test/constants.ts
```diff
@@ -96,6 +96,7 @@ export const MOCK_WEB3_RPC_URL = 'http://localhost:8545';
 export const MOCK_QUALIFICATION_MIN_VALIDITY = 100;
 export const MOCK_FE_URL = 'http://localhost:3001';
 export const MOCK_KYC_API_PRIVATE_KEY = 'api-private-key';
+export const MOCK_NDA_URL = 'https://staging.humanprotocol.org/nda';
```
portuu3, line 99: I think this is not used

portuu3 (changes requested):

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.spec.ts
flopez7: In this particular case I think it is better to leave it inside the signNDA module as it is not being modified and we can optimise the code

portuu3 (dismissed):

portuu3 (approved):

portuu3 (commented):
We need to verify that the nda is signed to allow accessing to oracles and jobs

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.spec.ts
dnechay: @flopez7 We can leave it inside of `signNDA` `describe` block and have its own inner `beforeEach` to generate new user.
It's not a big overhead to generate new simple small object, but we get independent test input for each run, which is not the case now because `user` is implicitly changed inside of `signNDA` method (it assigns `nda.url`) and you also change it in last test case  `should throw an error if the user has already signed the NDA`

Dzeranov:
@portuu3 I think that it should be done together with the frontend implementation. However, we can implement this with a feature flag on human app server.

packages/apps/reputation-oracle/server/src/modules/nda/nda.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+    service = module.get<NDAService>(NDAService);
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe('signNDA', () => {
+    let user: Pick<UserEntity, 'id' | 'email' | 'ndaSignedUrl'>;
+    beforeEach(async () => {
```
dnechay, line 39:
```suggestion
    beforeEach(() => {
```

dnechay (approved):
lgtm

Dzeranov (dismissed):
Lets do the frontend/server human app implementation in a separate PR.

portuu3 (dismissed):

portuu3 (approved):

Issue #3137, opened 2025-02-27
[Job Launcher] Move withdrawal creation logic to payments module


portuu3:
**Description**
Move the withdrawal creation logic to the payments module

**Motivation**
Enhance the modularity and testability of the code.

**Implementation details** 
To be defined by developer

PR #3147, opened 2025-03-03
[Job Launcher Server] Move withdrawal creation logic to payments module

flopez7:
## Issue tracking
#3137

## Context behind the change
Move the withdrawal creation logic to the payments module

## How has this been tested?
Deployed locally and launched one new escrow

## Release plan
None

Check payment creation in DB

### DIFF
--- a/packages/apps/job-launcher/server/src/modules/payment/payment.service.spec.ts
+++ b/packages/apps/job-launcher/server/src/modules/payment/payment.service.spec.ts
@@ -1529,4 +1529,61 @@ describe('PaymentService', () => {
       );
     });
   });
+
+  describe('createWithdrawalPayment', () => {
+    it('should create a withdrawal payment successfully', async () => {
+      const userId = 1;
+      const jobId = 2;
+      const amount = 100;
+      const currency = PaymentCurrency.USD;
+      const rate = 1;
+
+      jest
+        .spyOn(paymentRepository, 'createUnique')
+        .mockResolvedValueOnce(undefined as any);
+
+      await expect(
+        paymentService.createWithdrawalPayment(
+          userId,
+          jobId,
+          amount,
+          currency,
+          rate,
+        ),
+      ).resolves.not.toThrow();
+
+      expect(paymentRepository.createUnique).toHaveBeenCalledWith({
+        userId,
+        jobId,
+        source: PaymentSource.BALANCE,
+        type: PaymentType.WITHDRAWAL,
+        amount: -amount,
+        currency,
+        rate,
+        status: PaymentStatus.SUCCEEDED,
+      });
+    });
+
+    it('should throw an error if the payment creation fails', async () => {
+      const userId = 1;
+      const jobId = 2;
+      const amount = 100;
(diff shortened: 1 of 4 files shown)


portuu3 (commented):
I think withdrawal function should include the balance check so that is not possible to withdraw if the balance is not enough

flopez7:
> I think withdrawal function should include the balance check so that is not possible to withdraw if the balance is not enough
 There's a problem moving it directly to the method, as the check would be executed at the end, so if the user has no balance, the job entity would have already been created. 
There are many options here: 
- Keep it separate, that way we can do the check at the beginning without duplicating it.
- If you want to add it inside the method, you could run the creation of the payment at the beginning, making the check. 

IMO, we should keep them separately. This ensures that the balance is verified before any other operations are performed, preventing the creation of unnecessary entities and maintaining data consistency. Also, the `createWithdrawalPayment` method will remain focused on creating the withdrawal payment

cc: @portuu3 @Dzeranov

portuu3:
> > I think withdrawal function should include the balance check so that is not possible to withdraw if the balance is not enough
> 
> There's a problem moving it directly to the method, as the check would be executed at the end, so if the user has no balance, the job entity would have already been created. There are many options here:
> 
> * Keep it separate, that way we can do the check at the beginning without duplicating it.
> * If you want to add it inside the method, you could run the creation of the payment at the beginning, making the check.
> 
> IMO, we should keep them separately. This ensures that the balance is verified before any other operations are performed, preventing the creation of unnecessary entities and maintaining data consistency. Also, the `createWithdrawalPayment` method will remain focused on creating the withdrawal payment
> 
> cc: @portuu3 @Dzeranov

IMO, withdraw function should for sure include a balance check. 
About the order of payment entity and job entity creation, actually I think the payment should be created before the job entity, this way we avoid an edge case where we create the job and don't charge the user.

packages/apps/job-launcher/server/src/modules/payment/payment.service.spec.ts
```diff
@@ -1529,4 +1529,90 @@ describe('PaymentService', () => {
...
   });
+
+  describe('createWithdrawalPayment', () => {
+    it('should create a withdrawal payment successfully', async () => {
+      const userId = 1;
+      const amount = 100;
+      const currency = PaymentCurrency.USD;
+      const rate = 1;
+
+      jest
```
portuu3, line 1540: this should not be mocked, we should mock repository response

packages/apps/job-launcher/server/src/modules/payment/payment.service.spec.ts
```diff
@@ -1529,4 +1529,90 @@ describe('PaymentService', () => {
       );
     });
   });
+
+  describe('createWithdrawalPayment', () => {
+    it('should create a withdrawal payment successfully', async () => {
+      const userId = 1;
+      const amount = 100;
+      const currency = PaymentCurrency.USD;
+      const rate = 1;
```
portuu3, line 1538: these variables are used for every test case, we can define them globally inside this `describe`

Issue #3084, opened 2025-02-10
[Reputation Oracle] Revisit business logic


Dzeranov:
**Description**
Need to revisit business logic in reputation oracle.

**Motivation**
1. To make sure that everything works as intended.
2. To make sure that we're following best practices and our code is easy to understand and expandable if necessary.

**Implementation details** 
Revisit all the modules:

- [ ] auth
- [ ] cron-job
- [ ] escrow-completion
- [ ] health
- [ ] kyc
- [ ] payout 
- [ ] qualification
- [ ] reputation
- [ ] sendgrid
- [ ] storage
- [ ] user
- [ ] web3
- [ ] webhook
- [ ] hCaptcha
- [ ] common

PR #3174, opened 2025-03-11
[Reputation Oracle] refactor: user module

dnechay:
## Issue tracking
Part of #3084

## Context behind the change
ATM we have two different "domain entities" web2 user and operator (aka oracle) that stored within single DB entity - `User`. In this PR we are not going to split it because it requires some extra work on DB & migrations, so we only "separate" them using code and typings.

Goals of this PR are:
- cleanup entities & their relations
- change relations load: turn off eager loading by default and make the code to explicitly specify relations to load when they needed;
- introduce new test tools, e.g. user entity generators, that should make testing of other modules simpler
- rewrite unit tests for `UserService` to user common tooling and have slightly better structure/checks
- replace destructured import of utils to import all of them as a single namespace in order to improve readability
- atm `UserService` is highly-coupled in some places with `AuthService`, so some logic related to auth will be extracted/moved during `auth` refactoring

## How has this been tested?
- [x] unit tests
- [x] run `yarn migration:generate`, make sure no DB changes introduced other than migration added
- [x] locally: sign in as a user, make sure auth is success
- [x] locally: resend verification email & verify email address
- [x] locally: register address for user
- [x] locally: `/refresh` endpoint works
- [x] locally: `user/exchange-oracle-registration` should return sitekeys 

## Release plan
Simply merge

Some "eager" loading is off now and missing, which leads to exception. We should monitor logs after deployment, rollback is to run previous version.

No rollback needed for migration, it just aligns what we have in code & DB.


packages/apps/reputation-oracle/server/src/modules/user/types.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * ATM UserEntity is used to store two different domain objects
+ * that have intersection in properties:
+ * = Operator - aka "Oracle"; it should be authorized using web3 signature
+ * and always has evmAddres, but never has email, password and some other fields
+ * - Web2User - is "Worker", admin user and HUMAN App; they always
+ * have email & password, but might not have evmAddress
+ *
+ * Until we split the DB model - we differentiate them in code using differen types.
```
Dzeranov, line 12:
```suggestion
 * Until we split the DB model - we differentiate them in code using different types.
```

Dzeranov (dismissed):

PR #3179, opened 2025-03-11
[HUM-114]: fix: remove unused hook, move hook to module

mpblocky:
## Issue tracking

HUM-114

## Context behind the change

- moved hook to module where its used
- removed unused hook
- added types file 

## How has this been tested?

- tested build; no logic changes

## Release plan

normal

n/a


mpblocky:
@dnechay ready

dnechay (commented):
Schemas are not types, so let's rename `types` to `schemas` in this case

Also you can move some other schemas used in this module to that common file

mpblocky:
@dnechay should I separate types inferring from schemas to `types.ts`? The main reason to name it that way to not get too granular with files. 

Will do with the remaining schemas in that module!

dnechay:
> @dnechay should I separate types inferring from schemas to `types.ts`? The main reason to name it that way to not get too granular with files.
> 
> Will do with the remaining schemas in that module!

It's fine to keep schemas and inferred types in the same file

dnechay (approved):

PR #3187, opened 2025-03-12
[HUM-167]: fix: fix is mobile filter component logic

mpblocky:
## Issue tracking

HUM-167

## Context behind the change

- fixed filter prop

## How has this been tested?

- checked if filtering looks the same (mobile/desktop)
- checked if filtering works

## Release plan

normal

n/a


mpblocky:
@dnechay ready

dnechay (commented):
Could you please explain why we need to set `isMobile={isMobile}` for desktop filter components? Should we just omit this prop there?

IMU we have different sets of filter components for desktop and mobile that use `Filtering` component under the hood. For all desktop filters `isMobile` should be `false` according this PR, and for all mobile filters it should be `true`.

Am I missing something?

Also `isMobile` is indeed confusing and would be better to replace it with some sort of `showTitle` and `showClearButton` props

mpblocky:
@dnechay my mistake it will always be false in case of these components, so we can remove it completely from desktop components. I can do this or split this prop into two and name them separately (which was my first thought). Which do You prefer?

dnechay:
@mpblocky for me it's easier to understand when you have these props separately: one to display a title, another for "clean button"

dnechay (approved):

PR #3217, opened 2025-03-21
[Reputation Oracle] Abuse system

flopez7:
## Issue tracking
#3182

## Context behind the change
Report abuse endpoints: 
- Endpoint to report abuse. The abuse reports are stored in the database.

- Abuse results endpoint: Endpoint that receives the results of the abuse verification made by the support team. Based on results:
  -  Positive:
     - Slash in staking contract.
     - Notify the Job Launcher about the slash
  - Negative:
    - Decrease worker reputation
    - Notify Exchange Oracle
- Get abuse reports endpoint: Endpoint to get abuse reports open by the user.

Abuse webhooks: Reputation Oracle notifies the following entities:
- Human support team: Send a message to a private Slack channel.
- Exchange Oracle: Via webhook.

Slack workflow:

1. When a job (e.g., Job X) receives an abuse notification, we send a message to Slack containing key information such as the address, chain ID, and the URL of the manifest. This message also includes two buttons: Accept and Reject.
2. If support clicks Accept, Slack sends a message to the backend. The backend then responds with another Slack message that opens a modal, allowing support to enter the amount (in HMT) to be slashed.
3. Once the amount is submitted, Slack sends this data back to the backend, and we update the corresponding abuse entity in our database.
4. If support clicks Reject, we receive the message in the backend and mark the abuse entity as rejected.

## How has this been tested?
Deployed locally and generate some abuse reports. 
Ran unit tests

## Release plan
- ABUSE_SLACK_WEBHOOK_URL
- ABUSE_SLACK_OAUTH_TOKEN
- ABUSE_SLACK_SIGNING_SECRET

This shouldn't impact the current implementation. Please verify that the abuse logic functions correctly when submitting a new abuse report

### DIFF
--- a/packages/apps/reputation-oracle/server/src/integrations/slack-bot-app/slack-bot-app.spec.ts
+++ b/packages/apps/reputation-oracle/server/src/integrations/slack-bot-app/slack-bot-app.spec.ts
@@ -1,4 +1,3 @@
-import { Test } from '@nestjs/testing';
 import { HttpService } from '@nestjs/axios';
 import { SlackBotApp } from './slack-bot-app';
 import { faker } from '@faker-js/faker';
@@ -19,15 +18,6 @@ describe('SlackBotApp', () => {
   };
 
   beforeAll(async () => {
-    await Test.createTestingModule({
-      providers: [
-        {
-          provide: HttpService,
-          useValue: mockHttpService,
-        },
-      ],
-    }).compile();
-
     slackBotApp = new SlackBotApp(
       mockHttpService as unknown as HttpService,
       config,


packages/apps/reputation-oracle/server/src/modules/abuse/abuse.dto.ts
```diff
@@ -0,0 +1,28 @@
+import { ChainId } from '@human-protocol/sdk';
+import { ApiProperty } from '@nestjs/swagger';
+import { IsEnum, IsString } from 'class-validator';
+import { AbuseStatus } from '../../common/enums/abuse';
+
+export class ReportAbuseDto {
+  @ApiProperty({ name: 'chain_id' })
+  @IsEnum(ChainId)
+  public chainId: ChainId;
```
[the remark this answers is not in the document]
Dzeranov: Lets remove all the public modifiers, they're applied by default.

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+    enum: AbuseDecision,
+    nullable: true,
+  })
+  public decision?: AbuseDecision;
+
+  @Column({ type: 'decimal', precision: 30, scale: 18, nullable: true })
+  public amount?: number;
+
+  @JoinColumn()
+  @ManyToOne(() => UserEntity, (user) => user.abuse)
```
Dzeranov, line 35: Please check `user` entity on how we want to declare relations.

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.ts
```diff
@@ -0,0 +1,469 @@
...
+import { Web3Service } from '../web3/web3.service';
+import { WebhookOutgoingService } from '../webhook/webhook-outgoing.service';
+import { ReportAbuseDto, AbuseResponseDto } from './abuse.dto';
+import { AbuseEntity } from './abuse.entity';
+import { AbuseError, AbuseErrorMessage } from './abuse.error';
+import { AbuseRepository } from './abuse.repository';
+
+@Injectable()
+export class AbuseService {
+  [x: string]: any;
```
Dzeranov, line 27: ?

packages/apps/reputation-oracle/server/src/modules/user/user.entity.ts
```diff
@@ -67,6 +68,9 @@ export class UserEntity extends BaseEntity {
   @OneToOne('KycEntity', (kyc: KycEntity) => kyc.user)
   kyc?: KycEntity;
 
+  @OneToMany(() => AbuseEntity, (abuse) => abuse.user)
+  public abuse: AbuseEntity[];
```
Dzeranov, line 72: We can remove this declaration because we're not loading this relation anywhere.

Dzeranov (changes requested):
1. Lets create some sort of a `SlackService` responsible for generating modals, payloads etc (slack related). I think you can check sendgrid service for that matter.
2. Why do you need to build `EscrowClient` for get only operations? I think we can use utils for that.
3. Do we need this double checks on manifest_url, etc? I mean, if someone has reported an abuse, it means that escrow has been successfully processed by an exchange oracle which means that manifest_url, webhook_url etc exist.
4. Abuse service methods should accept parameters like chainId etc. instead of a raw slack body.

packages/apps/reputation-oracle/server/src/main.ts
```diff
@@ -8,16 +8,20 @@ import { AppModule } from './app.module';
 import { useContainer } from 'class-validator';
 import { ServerConfigService } from './config/server-config.service';
 import logger, { nestLoggerOverride } from './logger';
+import { getBodyParserOptions } from '@nestjs/platform-express/adapters/utils/get-body-parser-options.util';
```
dnechay, line 11: We should avoid importing anything by direct paths from libraries because it violates their encapsulation and can break at any moment, instead we can have our own implementation

packages/apps/reputation-oracle/server/src/modules/user/user.entity.ts
flopez7: I can't remove it because it's necessary for specifying the relation in abuse entity

dnechay: @flopez7 you actually can ([ref](https://typeorm.io/many-to-one-one-to-many-relations#:~:text=%40ManyToOne.-,If%20you%20want%20to%20use,-%40OneToMany%2C%20%40ManyToOne)), we already do this e.g. for `TokenEntity`

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -219,4 +221,58 @@ export class CronJobService {
...
+      return;
+    }
+
+    this.logger.info('Process Abuse START');
+    const cronJob = await this.startCronJob(CronJobType.ProcessRequestedAbuse);
+
+    try {
+      await this.abuseService.processAbuseRequests();
+    } catch (e) {
+      this.logger.error(e);
```
dnechay, line 245:
```suggestion
      this.logger.error('Error processing abuse requests', e);
```

packages/apps/reputation-oracle/server/src/modules/cron-job/cron-job.service.ts
```diff
@@ -219,4 +221,58 @@ export class CronJobService {
...
+      return;
+    }
+
+    this.logger.info('Process classified abuses START');
+    const cronJob = await this.startCronJob(CronJobType.ProcessClassifiedAbuse);
+
+    try {
+      await this.abuseService.processClassifiedAbuses();
+    } catch (e) {
+      this.logger.error(e);
```
dnechay, line 272:
```suggestion
       this.logger.error('Error processing classified abuse requests', e);
```

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
```diff
@@ -0,0 +1,46 @@
+import { Column, Entity, Index, JoinColumn, ManyToOne } from 'typeorm';
+
+import { DATABASE_SCHEMA_NAME } from '../../common/constants';
+import { BaseEntity } from '../../database/base.entity';
+import { UserEntity } from '../user/user.entity';
+import { ChainId } from '@human-protocol/sdk';
+import { AbuseDecision, AbuseStatus } from '../../common/enums/abuse';
+
+@Entity({ schema: DATABASE_SCHEMA_NAME, name: 'abuses' })
+@Index(['chainId', 'escrowAddress'], { unique: true })
```
dnechay, line 10: Should it be unique by these two params or by these two and user?

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+export class AbuseEntity extends BaseEntity {
+  @Column({ type: 'int' })
+  public chainId: ChainId;
```
dnechay, line 13: There is no need in `public` modifiers

dnechay (commented):
If my understanding of slack integration is correct, then config values and `SlackService` are able to interact with only one app - `HUMAN Abuse` that you created; so in case RepO might want to send some other messages/notifications - we will have to defined new set of env variables and write new slack service; this approach is not easy to maintain, so would be nice to change that; in case we want to have different apps for different features - then we might need to refactor `SlackService` to be injected per feature with specific config values instead of being injected globally; another approach is to have a single slack app for different purpose and provide channel id where to post for specific feature via config

What's related to the abuse logic itself - looks like described in PR context, so should be fine. The only question is about cases when the same escrow is reported by different users

packages/apps/reputation-oracle/server/src/common/enums/webhook.ts
```diff
@@ -1,6 +1,8 @@
 export enum EventType {
   JOB_COMPLETED = 'job_completed',
   ESCROW_COMPLETED = 'escrow_completed',
+  ABUSE_REPORTED = 'abuse_reported',
```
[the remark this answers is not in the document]
dnechay: Sorry, I got confused by the naming here and the mix of incoming/outgoing event types here, we 

These two new event types are actually for webhooks that RepO sends to other oracles, so they should be about some event that happened in Reputation Oracle, i.e. named with verb in past form.

When `abuse_reported` is clear enough for event name, `resume_reported_abuse` is read as a "command" name, not an "event", so I would like to suggest to change `RESUME_REPORTED_ABUSE` to something more meaningful, e.g. `abuse_dismissed` or `abuse_report_dismissed`.

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+    nullable: true,
+  })
+  public decision?: AbuseDecision;
+
+  @Column({ type: 'decimal', precision: 30, scale: 18, nullable: true })
+  public amount?: number;
+
+  @JoinColumn()
+  @ManyToOne('UserEntity')
+  public user: UserEntity;
```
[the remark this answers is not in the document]
flopez7: This can't be undefined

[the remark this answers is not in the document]
dnechay: It's undefined when you create new `AbuseEntity` or when you load it w/o loading a user relation. It's peculiarity of `typeorm` 🤷🏻

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.ts
```diff
@@ -0,0 +1,382 @@
...
+      BigInt(ethers.parseUnits(amount.toString(), 'ether')),
+    );
+  }
+
+  async receiveInteractions(data: any): Promise<string> {
+    const callback_id = data.callback_id
+      ? data.callback_id
+      : data.view.callback_id;
+    const callbackId = (callback_id as string).split('-');
+    const escrowAddress = callbackId[0];
```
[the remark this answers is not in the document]
flopez7: For now we will only accept one report per escrow so it should be fine. Will create the util as you said for it

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
flopez7: Since we're going to have only one request per escrow, this way is fine

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  @Column({
+    type: 'enum',
+    enum: AbuseDecision,
+    nullable: true,
+  })
+  decision: AbuseDecision | null;
+
+  @Column({ type: 'decimal', precision: 30, scale: 18, nullable: true })
+  amount?: number | null;
```
dnechay, line 32:
```suggestion
  amount: number | null;
```

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  @JoinColumn()
+  @ManyToOne('UserEntity', { nullable: false })
+  user?: UserEntity;
+
+  @Column({ type: 'int' })
+  userId: number;
+
+  @Column({ type: 'int' })
+  retriesCount = 0;
```
dnechay, line 42: Entity should not have default values to avoid reading/writing them when working with projected/partial data

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.entity.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  @Column({ type: 'timestamptz' })
+  waitUntil = new Date();
```
dnechay, line 45: Entity should not have default values to avoid reading/writing them when working with projected/partial data

packages/apps/reputation-oracle/server/src/modules/abuse/types.ts
```diff
@@ -0,0 +1,83 @@
+import { ChainId } from '@human-protocol/sdk';
+
+export type ReportAbuseInput = {
+  userId: number;
+  chainId: ChainId;
+  escrowAddress: string;
+};
+
+export type SlackInteractionBase = {
```
dnechay, line 9: There was no need to define everything in this and other slack types, probably only properties you use in business logic are necessary, but we can keep it, nice 👍🏻

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.ts
```diff
@@ -0,0 +1,293 @@
...
+    } else if (isViewSubmission(data)) {
+      const privateMetadata = JSON.parse(data.view.private_metadata);
+      const responseUrl = privateMetadata.responseUrl;
+      abuseEntity.decision = AbuseDecision.ACCEPTED;
+      abuseEntity.amount = data.view.state.values.quantity_input.quantity.value;
+      abuseEntity.retriesCount = 0;
+
+      await this.slackService.updateMessage(
+        responseUrl,
+        `Abuse ${(abuseEntity.decision as string).toLowerCase()}. Escrow: ${abuseEntity.escrowAddress}, ChainId: ${abuseEntity.chainId}, Slashed amount: ${abuseEntity.amount} HMT`,
```
dnechay, line 114:
```suggestion
        `Abuse ${abuseEntity.decision.toLowerCase()}. Escrow: ${abuseEntity.escrowAddress}, ChainId: ${abuseEntity.chainId}, Slashed amount: ${abuseEntity.amount} HMT`,
```

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.ts
```diff
@@ -0,0 +1,293 @@
...
+      );
+      await this.abuseRepository.updateOne(abuseEntity);
+      return '';
+    } else {
+      abuseEntity.decision = data.actions[0].value as AbuseDecision;
+      abuseEntity.retriesCount = 0;
+      await this.abuseRepository.updateOne(abuseEntity);
+      return `Abuse ${(abuseEntity.decision as string).toLowerCase()}. Escrow: ${abuseEntity.escrowAddress}, ChainId: ${abuseEntity.chainId}`;
```
dnechay, line 122:
```suggestion
      return `Abuse ${abuseEntity.decision.toLowerCase()}. Escrow: ${abuseEntity.escrowAddress}, ChainId: ${abuseEntity.chainId}`;
```

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.ts
```diff
@@ -0,0 +1,293 @@
...
+   */
+  private async handleAbuseError(abuseEntity: AbuseEntity): Promise<void> {
+    if (abuseEntity.retriesCount >= this.serverConfigService.maxRetryCount) {
+      abuseEntity.status = AbuseStatus.FAILED;
+    } else {
+      abuseEntity.waitUntil = new Date();
+      abuseEntity.retriesCount = abuseEntity.retriesCount + 1;
+    }
+
+    this.abuseRepository.updateOne(abuseEntity);
```
dnechay, line 141:
```suggestion
    await this.abuseRepository.updateOne(abuseEntity);
```

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.ts
```diff
@@ -0,0 +1,293 @@
...
+    for (const abuseEntity of abuseEntities) {
+      try {
+        const { chainId, escrowAddress } = abuseEntity;
+        const escrow = await EscrowUtils.getEscrow(
+          abuseEntity.chainId,
+          abuseEntity.escrowAddress,
+        );
+
+        if (abuseEntity.decision === AbuseDecision.ACCEPTED) {
+          this.slashAccount({
```
dnechay, line 212:
```suggestion
          await this.slashAccount({
```

packages/apps/reputation-oracle/server/src/modules/slack/slack.service.ts
```diff
@@ -0,0 +1,169 @@
...
+      throw error;
+    }
+  }
+
+  async sendAbuseNotification(data: {
+    abuseId: number;
+    chainId: ChainId;
+    escrowAddress: string;
+  }): Promise<void> {
+    const escrow = await EscrowUtils.getEscrow(
```
dnechay, line 71: This method should accept all necessary data for the message and not have any system-specific business logic, i.e. just take some input, format it to slack message/modal/whatever and send

packages/apps/reputation-oracle/server/src/modules/slack/slack.service.ts
```diff
@@ -0,0 +1,169 @@
...
+    chainId: ChainId;
+    escrowAddress: string;
+    triggerId: string;
+    responseUrl: string;
+  }) {
+    const escrow = await EscrowUtils.getEscrow(
+      data.chainId,
+      data.escrowAddress,
+    );
+    const maxAmount = (
```
dnechay, line 133: Same as for `sendAbuseNotification`: all inputs should be provided from the outside.
With current implementation when e.g. writing unit tests or playing around with this class you will have to mock `EscrowUtils` somehow, but if you just pass some inputs - you can easily test sending flow

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.ts
```diff
@@ -0,0 +1,293 @@
...
+  ) {}
+
+  async reportAbuse(data: ReportAbuseInput): Promise<void> {
+    const abuseEntity = new AbuseEntity();
+    abuseEntity.escrowAddress = data.escrowAddress;
+    abuseEntity.chainId = data.chainId;
+    abuseEntity.userId = data.userId;
+
+    await this.abuseRepository.createUnique(abuseEntity);
+    return;
```
dnechay, line 49:
```suggestion
```

dnechay (commented):
Nice work! Looks much cleaner and easier to digest now

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.repository.ts
```diff
@@ -0,0 +1,77 @@
...
+  ): Promise<AbuseEntity | null> {
+    return this.findOne({
+      where: {
+        chainId,
+        escrowAddress,
+      },
+    });
+  }
+
+  async findByStatus(status: AbuseStatus): Promise<AbuseEntity[]> {
```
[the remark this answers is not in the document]
flopez7: Totally agree, we can avoid some missunderstandings if clarifying that way

packages/apps/reputation-oracle/server/src/modules/abuse/types.ts
flopez7: I defined everything in case we need anything in the future, that way we know the properties the object will have and we can save time if we'll need to add any other data like user name or similar

packages/apps/reputation-oracle/server/src/common/enums/webhook.ts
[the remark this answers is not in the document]
flopez7: Would you mind if instead of abuse_reported and abuse_report_dismissed we use abuse_detected and abuse_dismissed. We already have abuse_detected in Job Launcher deployed

@Dzeranov @dnechay

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+            getSigner: jest.fn().mockReturnValue({
+              address: faker.finance.ethereumAddress(),
+              getNetwork: jest.fn().mockResolvedValue({ chainId }),
+            }),
+          },
+        },
+        { provide: AbuseRepository, useValue: createMock<AbuseRepository>() },
+        {
+          provide: ReputationService,
+          useValue: createMock<ReputationService>(),
```
dnechay, line 69: Please defined all mocks after imports, you don't have to get them from compiled module

It's a "style"-thing, but makes it easier to digest unit tests files when you looking at them

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+    reputationService = moduleRef.get<ReputationService>(ReputationService);
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('reportAbuse', () => {
+    it('should create a new abuse entity', async () => {
+      const userId = 1;
```
dnechay, line 93: Should use faker

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+        callback_id: abuseEntity.id,
+        chainId,
+        type: 'interactive_message',
+        actions: [{ value: AbuseDecision.ACCEPTED }],
+        trigger_id: faker.string.uuid(),
+        response_url: faker.internet.url(),
+      };
+
+      jest
+        .spyOn(abuseRepository, 'findOneById')
```
dnechay, line 126: You don't have to create new spy here, `abuseRepository` already mocked, so you should be able to mock resolved value w/o new spies

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+          state: {
+            values: {
+              quantity_input: { quantity: { value: 10 } },
+            },
+          },
+        },
+      };
+
+      jest
+        .spyOn(abuseRepository, 'findOneById')
```
dnechay, line 169: Same here, it's already a mock

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+        .mockResolvedValueOnce(abuseEntity);
+
+      await abuseService.processSlackInteraction(dto as any);
+
+      expect(abuseSlackBot.updateMessage).toHaveBeenCalledWith(
```
dnechay, line 174: Here and in other cases, would be also nice to check that some method is called exactly once

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.spec.ts
```diff
@@ -0,0 +1,495 @@
...
+
+      const dto = {
+        callback_id: abuseEntity.id,
+        chainId,
+        type: 'interactive_message',
+        actions: [{ value: AbuseDecision.REJECTED }],
+      };
+
+      jest
+        .spyOn(abuseRepository, 'findOneById')
```
dnechay, line 197: Same here & on other spies below

packages/apps/reputation-oracle/server/test/setup.ts
```diff
@@ -0,0 +1,82 @@
+import { KVStoreClient, KVStoreKeys, Role } from '@human-protocol/sdk';
+import { Wallet, ethers } from 'ethers';
+import * as Minio from 'minio';
+import * as dotenv from 'dotenv';
+dotenv.config({ path: '.env.local' });
+
+export async function setup(): Promise<void> {
```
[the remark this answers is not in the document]
flopez7: It was committed by error, sorry

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.module.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AbuseController } from './abuse.controller';
+import { AbuseSlackAuthGuard } from './abuse.slack-auth.guard';
+
+@Module({
+  imports: [HttpModule, Web3Module, ReputationModule, WebhookOutgoingModule],
+  providers: [
+    AbuseRepository,
+    AbuseService,
+    AbuseSlackBot,
+    AbuseSlackAuthGuard,
```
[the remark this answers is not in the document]
flopez7: I have read that it is necessary to properly inject dependencies

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.service.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+        webhookUrl: webhookUrl1,
+      } as IOperator);
+      mockWebhookOutgoingService.createOutgoingWebhook.mockResolvedValueOnce(
+        undefined,
+      );
+
+      await abuseService.processClassifiedAbuses();
+
+      expect(mockAbuseRepository.findClassified).toHaveBeenCalledTimes(1);
+      // expect(slashMock).toHaveBeenCalledWith(
```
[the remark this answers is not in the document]
flopez7: I left this code there as a reminder. That's why it's commented, because Reputation Oracle is not allowed for slashing.
We still have to discuss how are we going to handle slash

yarn.lock
```diff
@@ -2,6 +2,11 @@
 # yarn lockfile v1
 
 
+"@adraffy/ens-normalize@1.10.0":
```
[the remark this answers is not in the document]
flopez7: Already done, yarn.lock is up to date.

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.module.ts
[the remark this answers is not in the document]
dnechay: @flopez7 could you please point me to that doc/article?
From what I see/tried, it's not necessary for guards if you use them as constructor token (i.e. `@UseGuards(AbuseSlackAuthGuard)`), nest should be able to resolve dependencies in this case

yarn.lock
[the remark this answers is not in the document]
dnechay: @flopez7 maybe something changed after merging develop, because when I run `yarn install` locally from your branch - it changes `yarn.lock` when it shouldn't

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.slack-bot.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      await expect(
+        abuseSlackBot.sendAbuseNotification({
+          abuseId,
+          chainId,
+          escrowAddress,
+          manifestUrl,
+        }),
+      ).resolves.not.toThrow();
+
+      expect(mockHttpService.post).toHaveBeenCalledWith(
```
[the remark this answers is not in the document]
flopez7: IMO this way is better, as you said we avoid to test some already tested methods.

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.module.ts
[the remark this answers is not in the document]
flopez7: Guards in Nest are treated as providers, they all have the @injectable decorator.Also, if a guard has dependencies, it must be listed in the providers array so Nest can inject those dependencies properly. I saw it in official docs:
- [Providers](https://docs.nestjs.com/providers)
- [Guards](https://docs.nestjs.com/guards)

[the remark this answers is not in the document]
dnechay: @flopez7 I suppose that you don't have to list the guard in `providers` list in case you use it somewhere in `@UseGuards` decorator, because that decorator will take care of DI (they have an example with `useGlobalGuards` vs `provide: APP_GUARD`). In our case we inject only `SlackConfigService` which is global, so no DI should be done properly, In case this guard had some other DI from the module - then yes, we would need to use it as provider in order to properly inject dependencies.

I would try removing it from providers list and try triggering the endpoint with dumb data to see if guard is called. If DI is not done - app won't even run.

Anyways, this comment is not blocking, so up to you if you want to check this

dnechay (approved):

packages/apps/reputation-oracle/server/src/modules/abuse/abuse.module.ts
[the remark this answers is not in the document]
flopez7: You're right. If the guard only depends on globally provided services, you don't necessarily have to list it in the providers array. However, I would keep it as is and register it in the providers array. Even if it only depends on a global service right now, registering it ensures proper DI and makes future changes easier

Dzeranov:
@flopez7 looks good.
Just one thing to remember - this abuse system is going to work only with fortune for now since we've yet to implement new webhooks for cvat oracles.

Dzeranov (approved):

PR #3229, opened 2025-03-25
feat: enable operator functionality

### LINKED ISSUE #3216: [Human App][Frontend] Add enable operator functionality
**Description**
Add enable operator functionality

**Motivation**
This functionality is missing on client and marked as [todo](https://github.com/humanprotocol/human-protocol/blob/develop/packages/apps/human-app/frontend/src/modules/operator/profile/components/profile-enable-button.tsx).

Linked with #3215

dnechay:
## Issue tracking
Closes #3216 

## How has this been tested?
- [x] locally e2e; sign in as operator; make it active, then deactivate

## Release plan
As part of parent PR w/o extra steps

After FE runs enable/disable operation, it does refresh of auth tokens because  jwt stores `operator_status`. When adding `operator_status` to JWT, we first read its value from KV store, so there is a risk of reading a stale value from subgraph in many cases. Same applies to scenario when you e.g. deactivate operator and immediately attempt to activate it: RepO has a validation on operator status before calling `KV.set`, so operation may fail.

In order to alleviate this replication lag, after activating/deactivating operator we manually update its data in browser memory in order to display valid status. However if refreshed token has stale value and user refreshes the browser page - they will still see outdated value. In case we find an issue in it - let's address it in separately.


packages/apps/human-app/frontend/src/modules/operator/profile/components/profile-disable-button.tsx
```diff
@@ -23,7 +23,7 @@ export function ProfileDisableButton() {
   const {
     mutate: disableOperatorMutation,
     isError: isDisableOperatorError,
-    isPaused: isDisableOperatorPending,
+    isPending: isDisableOperatorPending,
```
dnechay, line 26: This is to fix "loading spinner" on "Deactivate" button; right now it's missing

Dzeranov (approved):

PR #3231, opened 2025-03-27
[HUM-172]: fix: refactor bug weird redirect behaviour

mpblocky:
## Issue tracking

HUM-172

## Context behind the change

- removed unnecessary context
- added usage of registered oracles endpoint in `useIsAlreadyRegistered`
- moved redirect for unverified workers from profile page to require component

## How has this been tested?

- checked registration in oracle redirect: if it redirects if user is already registered 
- checked if redirect for inactive user works properly

## Release plan

normal

n/a


mpblocky:
@dnechay there is something that need to be discussed regarding the routing You [mentioned earlier](https://github.com/humanprotocol/human-protocol/pull/3014#pullrequestreview-2640556099).

In the router line 115 catches all routes that are not defined. This also catches `/worker/registration-in-exchange-oracle/` without address as param. This causes redirect to profile. To avoid this, we would need to define route just for this or dig deeper to change behaviour for that page. I think it's a bit overkill, especially that probability of scenario when a user uses this URL without address is low. WDYT?

dnechay:
@mpblocky 
Agree that it's not highly likely to happen so we can omit this scenario
However if we open `worker/registration-in-exchange-oracle/<address>` and use is already registered - would be nice to get redirected to the page with list of jobs for that oracle

packages/apps/human-app/frontend/src/modules/auth-web3/context/web3-auth-context.tsx
```diff
@@ -21,7 +21,7 @@ const web3userDataSchema = z.object({
   reputation_network: z.string(),
   operator_status: z.nativeEnum(OperatorStatus),
   exp: z.number(),
-  status: z.string().nullable().optional(),
```
mpblocky, line 24: @dnechay are those correct statuses for users?

packages/apps/human-app/frontend/src/modules/auth/context/auth-context.tsx
```diff
@@ -14,7 +14,7 @@ const extendableUserDataSchema = z.object({
   site_key: z.string().optional().nullable(),
   kyc_status: z.string().optional().nullable(),
   wallet_address: z.string().optional().nullable(),
-  status: z.string().optional().nullable(),
```
mpblocky, line 17: @dnechay same question as above

packages/apps/human-app/frontend/src/modules/auth-web3/providers/require-web3-auth.tsx
```diff
@@ -22,6 +24,16 @@ export function RequireWeb3Auth({ children }: { children: JSX.Element }) {
...
+  if (web3Auth.user.status !== 'active' || web3Auth.status !== 'success') {
+    return (
+      <Navigate
+        replace
+        state={{ from: location }}
+        to={routerPaths.operator.connectWallet}
+      />
+    );
+  }
+
```
mpblocky, line 36: Can this redirect happen in those scenarios?

mpblocky:
@dnechay ready for review, can You also check my comments?

packages/apps/human-app/frontend/src/modules/auth/context/auth-context.tsx
dnechay: For worker user it can be `active`, `pending` and `inactive`

dnechay: @mpblocky yes, for worker auth these 3 are valid

packages/apps/human-app/frontend/src/modules/auth-web3/providers/require-web3-auth.tsx
dnechay: @mpblocky good question!
if user status is not active - then I don't thing we should redirect to this page

For the failed web3 auth - yup, might be useful redirect

mpblocky: If user is inactive - should we redirect at all?

dnechay: There is a change in progress where we won't allow to sign-in if user has `inactive` status, so I suppose you can remove this redirect yes for web3 users

mpblocky:
@dnechay ready for review

packages/apps/human-app/frontend/src/main.tsx
```diff
@@ -44,9 +43,7 @@ createRoot(root).render(
                   <AuthProvider>
                     <DisplayModal />
                     <JWTExpirationCheck>
-                      <RegisteredOraclesProvider>
```
dnechay, line 47: Nice!

packages/apps/human-app/frontend/src/modules/worker/oracle-registration/hooks/use-oracle-registration-flow.ts
```diff
@@ -2,11 +2,10 @@ import { useOracleInstructions } from './use-oracle-instructions';
 import { useOracleRegistration } from './use-oracle-registration';
 
 export function useOracleRegistrationFlow(
-  address: string,
```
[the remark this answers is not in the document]
mpblocky: @dnechay I'm not sure if I understand what do You mean, could You explain it a bit more?

BTW I think we also can extract `useOracleInstructions` and `useOracleRegistration` from `useOracleRegistrationFlow` for simplicity.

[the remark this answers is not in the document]
dnechay: @mpblocky 
> BTW I think we also can extract useOracleInstructions and useOracleRegistration from useOracleRegistrationFlow for simplicity

Good suggestion!

> could You explain it a bit more?

Sure! Right now we pass `address` to `useOracleRegistration` hook and it's explicit: when I look at it I understand that `useOracleRegistration` hook needs specific oracle address that it will register. When you remove `address` from params - I have to open the hook in order to understand it's implicit logic and how it decided which oracle to register. ATM this "implicit" logic is "hidden" in registration form fields: when you click "submit" form calls `handleRegistration` with `oracle_address` and `h_captcha_token`. As a result I should go back and forth between files in order to understand how it works. I would like to simplify it.

We already know `oracleAddress`, so no need to have it as part of the form, the form itself is needed only to get the `h_captcha_token`, so I would do the next set of things:
- get rid of `useOracleRegistrationFlow`, by extracting `useOracleInstructions` from it and `useOracleRegistration`  hook logic
- `useOracleRegistration` is just som wrapper now around `useExchangeOracleRegistrationMutation` and there is no need to have it, we can move `useExchangeOracleRegistrationMutation` usage directly to `RegistrationForm`
- then in `RegistrationForm` component do something like
```
const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
  void methods.handleSubmit((formData: { h_captcha_token: string }) => {
    registerInOracle({
      h_captcha_token: formData.h_captcha_token,
      oracle_address: address
    });
  })(event);
};
```

This way I open `RegistrationForm` and it's immediately clear what's going on.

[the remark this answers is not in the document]
mpblocky: Thanks :) The use registration flow hook with just instructions was implicit/confusing as well, but I left it to Your review - now everything should look much better :)

dnechay (approved):

PR #3239, opened 2025-03-28
[HUM-175]: fix: fix table header props ref warning

mpblocky:
## Issue tracking

HUM-175

## Context behind the change

- modified table header cell component
- adjusted column definitions

## How has this been tested?

- checked if filters work and look the same (only desktop)

## Release plan

normal

n/a

### DIFF
--- a/packages/apps/human-app/frontend/src/shared/components/ui/table/table-header-cell.tsx
+++ b/packages/apps/human-app/frontend/src/shared/components/ui/table/table-header-cell.tsx
@@ -42,29 +42,13 @@ export function TableHeaderCell({
 
   const getHeader = () => {
     if (!iconType) {
-      return (
-        <div
-          role="button"
-          onClick={handleClick}
-          tabIndex={0}
-          onKeyDown={(e) => {
-            if (e.key === 'Enter' || e.key === ' ') handleClick(e);
-          }}
-        >
-          {headerText}
-        </div>
-      );
+      // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+      return <div onClick={handleClick}>{headerText}</div>;
     }
 
     return (
-      <div
-        role="button"
-        onClick={handleClick}
-        tabIndex={0}
-        onKeyDown={(e) => {
-          if (e.key === 'Enter' || e.key === ' ') handleClick(e);
-        }}
-      >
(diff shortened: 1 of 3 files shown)


packages/apps/human-app/frontend/src/shared/components/ui/table/table-header-cell.tsx
```diff
@@ -42,12 +42,31 @@ export const TableHeaderCell = forwardRef<
...
-      return <td onClick={handleClick} {...rest} />;
+      return (
+        <div
+          role="button"
+          onClick={handleClick}
+          tabIndex={0}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter' || e.key === ' ') handleClick(e);
+          }}
+        >
```
mpblocky, line 53: This is necessary to omit React errors with interactive elements. We also could use MUI Button, but that changes the look of the header:
- makes it higher
- other non-interactive (no sorted/filtered) headers will have to be changed as well 

WDYT?

mpblocky:
@dnechay ready with a question

packages/apps/human-app/frontend/src/shared/components/ui/table/table-header-cell.tsx
dnechay: @mpblocky Could you please share some context on why we need to change it from `td` to `div`?

IMU we don't need to introduce it and interactive stuff here: you don't know what exact header cell you render, but they all have `tabIndex=0` which is a bit confusing

I would keep header as simple as possible: simples HTML tag + just onClick handler

mpblocky: It showed warning `validateDOMNesting(...):` where a td cannot be a child of `div`. I also can wrap it in a table like so:

```
      <table>
        <tbody>
          <tr>
            <td onClick={handleClick}>
              <TextHeaderWithIcon iconType={iconType} text={headerText} />
            </td>
          </tr>
        </tbody>
      </table>
```

dnechay: @mpblocky I see now, thank you

For me it looks good. Even if there is a better approach to handle that, it might introduce some extra refactoring to `TableHeaderCell` & `TextHeaderWithIcon` and I don't think we need to do it while fixing `ref` warning (aka in this PR)

dnechay (approved):

PR #3249, opened 2025-03-31
[Reputation Oracle] refactor: storage module

dnechay:
## Issue tracking
Part of https://github.com/humanprotocol/human-protocol/issues/3084

## Context behind the change
Refactored `storage` module:
- moved out some static utils to be in `http` utils file
- moved encryption-related stuff to be in corresponding module
- moved business logic on job solutions to be a part of `payout` module for now

Also thought about retrieving PGP public key for RepO from its private key & passphrase in order to avoid going to KV for that, but sdk doesn't provide such functionality and implementing it in service would be an overhead.

## How has this been tested?
- [x] locally: create fake public endpoint to call `uploadJobResults` with fake data, make sure data encrypted and uploaded to local minio
- [x] locally: use same fake endpoint to call `downloadJsonLikeData` for the file from test above, make sure data downloaded and decrypted
- [x] unit tests

## Release plan
Just merge

Something unexpected in affected flows, should be easily defined by logging.


packages/apps/reputation-oracle/server/src/modules/encryption/encryption.module.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Web3Module } from '../web3/web3.module';
+
+import { PgpEncryptionService } from './pgp-encryption.service';
+
+@Module({
+  imports: [Web3Module],
+  providers: [PgpEncryptionService],
+  exports: [PgpEncryptionService],
+})
+export class EncryptionModuleModule {}
```
Dzeranov, line 12:
```suggestion
export class EncryptionModule {}
```

dnechay: Fixed in https://github.com/humanprotocol/human-protocol/pull/3249/commits/8a10ca865fd8b839897f54d3d5e601c0550c1f81

Dzeranov (approved):

PR #3276, opened 2025-04-10
[HUM-164]: feat: add new http api client

mpblocky:
## Issue tracking

HUM-164

## Context behind the change

- added new http api clients: authorized, unauthorized, auth service
- migrated all endpoints to use it
- removed dead code related to fetcher

## How has this been tested?

- tested every endpoint to ensure it works as before

## Release plan

normal

Monitor incorrect requests in case there was something that was omitted during testing


mpblocky:
@dnechay This one is ready. I know it's huge but I tried to commit every piece separately for easier review. LMK if You need something!

mpblocky:
@dnechay small fixes for my jobs needed, will let You know

mpblocky:
@dnechay ready

packages/apps/human-app/frontend/src/modules/worker/profile/hooks/use-start-kyc-mutation.ts
```diff
@@ -1,32 +1,17 @@
...
-        const result = await apiClient(apiPaths.worker.kycStart.path, {
-          successSchema: kycStartSchema,
-          authenticated: true,
-          withAuthRetry: apiPaths.worker.kycStart.withAuthRetry,
-          options: {
-            method: 'POST',
-          },
-        });
-        return result;
+        return await profileService.startKyc({ provider: 'veriff' });
```
dnechay, line 14: Probably we can just call `startKyc` w/o `try..catch` here? I don't get why we need to catch errors and refresh

packages/apps/human-app/frontend/src/modules/operator/profile/services/profile.service.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.httpApiClient = new HttpApiClient(env.VITE_API_URL);
+    this.authService = new AuthService(this.httpApiClient);
+
+    this.authorizedHttpApiClient = new AuthorizedHttpApiClient(
+      env.VITE_API_URL,
+      this.authService
+    );
+  }
+
+  async enableOperator(data: { signature: string }) {
```
dnechay, line 28:
```suggestion
  async enableOperator(signature: string) {
```

packages/apps/human-app/frontend/src/modules/operator/profile/services/profile.service.ts
```diff
@@ -0,0 +1,87 @@
...
+    } catch (error) {
+      if (error instanceof ApiClientError) {
+        throw error;
+      }
+
+      throw new Error('Failed to enable operator');
+    }
+  }
+
+  async disableOperator(data: { signature: string }) {
```
dnechay, line 47:
```suggestion
  async disableOperator(signature: string) {
```

packages/apps/human-app/frontend/src/modules/operator/profile/hooks/use-enable-operator.ts
```diff
@@ -1,28 +1,20 @@
...
-        options: {
-          method: 'POST',
-          body: JSON.stringify({ signature }),
-        },
+      const result = await operatorProfileService.enableOperator({
+        signature,
       });
 
-      await refreshAccessTokenAsync({ authType: 'web3' });
+      await operatorProfileService.refreshAccessToken();
```
dnechay, line 16: Why not calling authService directly here?

packages/apps/human-app/frontend/src/modules/operator/profile/hooks/use-disable-operator.ts
```diff
@@ -1,28 +1,20 @@
...
-        options: {
-          method: 'POST',
-          body: JSON.stringify({ signature }),
-        },
+      const result = await operatorProfileService.disableOperator({
+        signature,
       });
 
-      await refreshAccessTokenAsync({ authType: 'web3' });
+      await operatorProfileService.refreshAccessToken();
```
dnechay, line 16: Why not calling authService directly here?

packages/apps/human-app/frontend/src/modules/worker/jobs/schemas.ts
```diff
@@ -1,7 +1,7 @@
-/* eslint-disable camelcase -- api response*/
+/* eslint-disable camelcase */
 import { z } from 'zod';
 import { createPaginationSchema } from '@/shared/helpers/pagination';
-import { MyJobStatus, UNKNOWN_JOB_STATUS } from '../types';
+import { MyJobStatus, UNKNOWN_JOB_STATUS } from './types';
 
 const myJobSchema = z.object({
   assignment_id: z.string(),
```
dnechay, line 7: Should this schema take into account that `MyJobStatus` is now lowercased?
If I understand code correctly, `z.nativeEnum(MyJobStatus).parse(value.toUpperCase())` will fail now for any value from response

packages/apps/human-app/frontend/src/modules/worker/jobs/services/jobs.service.ts
```diff
@@ -0,0 +1,155 @@
...
+      return result;
+    } catch (error: unknown) {
+      if (error instanceof ApiClientError) {
+        throw error;
+      }
+      throw new Error('Failed to refresh jobs');
+    }
+  }
+
+  async getUiConfig() {
```
dnechay, line 136: It shouldn't be part of the jobs service. Right now it has only `chainIdsEnabled`, but in general it can have some other params used here and there, so should be general

packages/apps/human-app/frontend/src/modules/worker/jobs/hooks/use-get-ui-config.ts
```diff
@@ -1,28 +1,10 @@
...
-    successSchema: uiConfigSchema,
-    options: {
-      method: 'GET',
-    },
-  });
-  return response;
-};
+import { jobsService } from '../services/jobs.service';
 
 export function useGetUiConfig() {
```
dnechay, line 4: Same for this hook: it should be general, not part of `jobs`

packages/apps/human-app/frontend/src/modules/worker/hooks/use-get-oracles.ts
```diff
@@ -51,52 +36,45 @@ const H_CAPTCHA_ORACLE: Oracle = {
...
+export function useGetOracles() {
+  const { selected_job_types } = useJobsTypesOraclesFilterStore();
 
-  if (env.VITE_FEATURE_FLAG_JOBS_DISCOVERY) {
-    const queryParams = selected_job_types.length
-      ? `?${stringifyUrlQueryObject({ selected_job_types })}`
-      : '';
+  return useQuery({
+    queryFn: async () => {
+      try {
```
dnechay, line 44: Probably we can hide all this logic under `oraclesService.getOracles`?

packages/apps/human-app/frontend/src/modules/signup/worker/hooks/use-sign-up-mutation.ts
```diff
@@ -1,26 +1,25 @@
...
   const navigate = useNavigate();
 
   return useMutation({
-    mutationFn: signUpMutationFn,
+    mutationFn: async (data: Omit<SignUpDto, 'confirmPassword'>) => {
+      return signupService.workerSignUp({
+        email: data.email,
+        password: data.password,
+        // eslint-disable-next-line camelcase
+        h_captcha_token: data.h_captcha_token,
```
dnechay, line 20: Probably now you can use `hCaptchaToken` everywhere and pass it snake_cased inside of `signupService.workerSignUp`

packages/apps/human-app/frontend/src/modules/operator/profile/services/profile.service.ts
dnechay: And in all other similar places where you don't expect any return value from API

packages/apps/human-app/frontend/src/modules/worker/email-verification/hooks/resend-email-verification.ts
```diff
@@ -1,27 +1,25 @@
...
 
 async function resendEmailVerificationMutationFn(
-  data: ResendEmailVerificationDto,
-  isAuthenticated: boolean
+  _data: ResendEmailVerificationDto,
+  isAuthenticated: boolean,
+  userEmail: string | undefined
 ) {
-  if (!isAuthenticated) {
+  if (!isAuthenticated || !userEmail) {
```
dnechay, line 15: Is it possible that user is authenticated but does not have email?
Also from what I see in API schema, there is no need to send `email` for this endpoint

packages/apps/human-app/frontend/src/modules/signup/operator/views/edit-existing-keys-success.page.tsx
```diff
@@ -35,7 +32,8 @@ export function EditExistingKeysSuccessPage() {
     isPending: web3SignUpPending,
   } = useWeb3SignUp();
 
-  const createSignature = async (data: SignatureData) => {
+  const createSignature = async () => {
```
dnechay, line 35: Would be nice to have a different name, something like `handleWeb3Signup`?

packages/apps/human-app/frontend/src/modules/signup/operator/hooks/use-web3-signup.ts
```diff
@@ -1,35 +1,32 @@
...
+        });
+        return response;
+      } catch (error) {
+        if (error instanceof ApiClientError) {
+          throw new Error(String(error.data));
+        }
+
+        throw new Error('Failed to sign up operator with web3');
+      }
+    },
```
dnechay, line 29: It's already part of `signupService.operatorWeb3SignUp`

dnechay (commented):
HTTP clients & auth handling lgtm as for now
Some comments below

1. You don't have to create an instance of `HttpApiClient` or `AuthorizedHttpApiClient` in every specific service, they all have the same base api url, so you can create these instances once and simply use them later w/o wrapping in any classes, e.g.
```
import { humanAppApiClient } from '@/api';
...
const result = await humanAppApiClient.post<SignatureData>(
  apiPaths.web3.prepareSignature,
  {
    body: { ...data },
    successSchema: prepareSignatureSuccessSchema,
  }
);

return result;
```
2. Deployment fails due to build errors, please check them

packages/apps/human-app/frontend/src/api/http-api-client.ts
```diff
@@ -0,0 +1,145 @@
...
+      );
+
+      throw new ApiClientError(errorMessage, response.status, responseBody);
+    }
+
+    if (successSchema) {
+      try {
+        successSchema.parse(responseBody);
+      } catch (error) {
+        throw new Error(`Error parsing response body: ${String(error)}`);
```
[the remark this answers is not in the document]
mpblocky: Yes, we can throw `ApiClientError` with Zod message and then throw generic Error on any other error, WDYT?

packages/apps/human-app/frontend/src/api/authorized-http-api-client.ts
```diff
@@ -0,0 +1,51 @@
...
+      };
+
+      return await super.makeRequest(method, path, _config);
+    } catch (error) {
+      if (
+        error instanceof ApiClientError &&
+        error.status === 401 &&
+        retryOnFailedAuth
+      ) {
+        await this.authProvider.refreshAccessToken();
```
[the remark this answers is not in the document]
mpblocky: Yes! Thanks, I will fix that, a lot of code moved around. Good catch 👀

packages/apps/human-app/frontend/src/api/authorized-http-api-client.ts
```diff
@@ -0,0 +1,51 @@
...
+        return await this.makeRequest(method, path, config, false);
+      }
+
+      if (error instanceof ApiClientError) {
+        throw new ApiClientError(error.message, error.status, error.data);
```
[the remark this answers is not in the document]
mpblocky: Yes, we can remove it

packages/apps/human-app/frontend/src/modules/homepage/hooks/use-web3-signin.ts
```diff
@@ -23,24 +22,24 @@ export function useWeb3SignIn() {
   const { address, chainId, signMessage } = useWalletConnect();
   const { signIn } = useWeb3Auth();
   const navigate = useNavigate();
+  const { mutateAsync } = usePrepareSignature(PrepareSignatureType.SIGN_IN);
+
   return useMutation({
-    mutationFn: async (body: PrepareSignatureBody) => {
-      const dataToSign = await prepareSignature(body);
+    mutationFn: async () => {
+      const data = await mutateAsync();
```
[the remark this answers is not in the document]
mpblocky: I thought it looks cleaner 😅 Will change

packages/apps/human-app/frontend/src/modules/worker/profile/hooks/use-start-kyc-mutation.ts
mpblocky: It was in previous impl: I thought there is some implicit handling, will remove that

packages/apps/human-app/frontend/src/modules/operator/profile/services/profile.service.ts
mpblocky: Will fix

packages/apps/human-app/frontend/src/modules/operator/profile/hooks/use-enable-operator.ts
mpblocky: Seemed cleaner for me, will fix :)

packages/apps/human-app/frontend/src/modules/worker/jobs/schemas.ts
mpblocky: I will double check that

packages/apps/human-app/frontend/src/modules/worker/email-verification/hooks/resend-email-verification.ts
mpblocky: I wasn't sure if that can happen - if not, then I will remove it. It also was previously implemented

packages/apps/human-app/frontend/src/api/http-api-client.ts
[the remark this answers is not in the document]
dnechay: @mpblocky 
If we throw `ApiClientError` from here - it will eventually bubble up to the user in different places (e.g. in alerts) where we take message from error.

Normally - we shouldn't get to this `catch` block at all because response schema either should be backward compatible or changed together with FE. So I would just log the error for visibility and throw some simple like
```
console.log('Invalid response schema, {
  path,
  error,
});
throw new Error('Invalid response schema`);
```

This way we can see what's going on and at the same time use will see meaningful error message from corresponding service, something like `Failed to load jobs`

packages/apps/human-app/frontend/src/modules/worker/profile/hooks/use-start-kyc-mutation.ts
dnechay: @mpblocky 
Then it was missed in the past, let's clean it up :)

packages/apps/human-app/frontend/src/modules/worker/email-verification/hooks/resend-email-verification.ts
dnechay: @mpblocky probably there might be some edge case, but it should be handled in onSubmit handler of ResendVerificationEmailForm component to avoid using this mutation at all

packages/apps/human-app/frontend/src/api/http-api-client.ts
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        url.searchParams.append(key, String(value));
+      });
+    }
+
+    const response = await fetch(url, {
+      method,
+      headers: {
+        'Content-Type': 'application/json',
```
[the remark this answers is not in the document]
mpblocky: aren't they case-insensitive? 🤔  I will change that anyway

packages/apps/human-app/frontend/src/modules/homepage/hooks/use-web3-signin.ts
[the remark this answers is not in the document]
mpblocky: actually @dnechay it is used in multiple places, making and it indeed looks cleaner - we don't have to call `useWalletConnect` to get the address every time. I know it's called here anyway, but then we would need to import `auth.service.ts` from `auth` module, which we shouldn't export as we talked about services before. WDYT?

packages/apps/human-app/frontend/src/api/http-api-client.ts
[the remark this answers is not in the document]
dnechay: @mpblocky They are case-insensitive, which means you can pass them as `header-name` or `HeAdEr-NaMe`, but if you have multiple values in the headers object - results might be surprising depending on the LB/server.

For example, you can open https://httpbin.org/ and execute next code in the browser console:
```
let res = await fetch('https://httpbin.org/headers', {
    headers: {
        'Content-Type': 'first',
        'content-type': 'second',
    }
});
let body = await res.json()
console.log(body);
```

packages/apps/human-app/frontend/src/modules/homepage/hooks/use-web3-signin.ts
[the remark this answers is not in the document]
dnechay: @mpblocky oh, I missed the part with `useWalletConnect` while reviewing it, then it indeed makes sense to keep it as is! I would just give it a better name than `mutateAsync` :)

packages/apps/human-app/frontend/src/modules/operator/profile/hooks/use-enable-operator.ts
mpblocky: Looking at it again, it looks kind of better for me: we don't need to initialize new `authService` for that hook just to refresh token, hook uses the service as intended, are You sure we need to change that or even export the authService from `@api`?

mpblocky: Looking at it again, it looks kind of better for me: we don't need to initialize new `authService` for that hook just to refresh token, hook uses the service as intended, are You sure we need to change that or even export the authService from `@api`?

packages/apps/human-app/frontend/src/modules/operator/profile/hooks/use-disable-operator.ts
mpblocky: Continue [in here](https://github.com/humanprotocol/human-protocol/pull/3276#discussion_r2050472418)

packages/apps/human-app/frontend/src/modules/operator/profile/hooks/use-enable-operator.ts
mpblocky: Alright, just wanted to be sure that You want me to actually export authService :) Will do

mpblocky:
@dnechay everything addressed

packages/apps/human-app/frontend/src/modules/worker/jobs/schemas.ts
dnechay: @mpblocky it indeed works, but in a tricky way.
From server it gets `status` as lowercased value, then it validates it using `myJobSchema` schema by applying the transformation. When transformation is applied there is error due to `value.toUpperCase()` usage, but this error is caught by `catch` and `UNKNOWN_JOB_STATUS` is return as `transform` result. Reason why it shows actual values on UI is hidden in a way how schema validation runs in `HttpApiClient`:
```
try {
  successSchema.parse(responseBody);
}
```
it won't fail because error is caught in `transform` itself and also you don't return the result of schema parsing you use the response as is.

So there are two items to fix:
1) After doing `successSchema.parse` - parsed value should be return, so transformation and `refine` are applied if used
2) `MyJobStatus` validation

After you fix these two - you won't need `textTransform: 'uppercase'`

dnechay (commented):
Looks better!

Two comments from previous review are still actual: [№1](https://github.com/humanprotocol/human-protocol/pull/3276#discussion_r2052576482),  [№2](https://github.com/humanprotocol/human-protocol/pull/3276#discussion_r2050459766) and some minor items to cleanup

1. "Reset Password" functionality is not working due to absence of `h_captcha_token` in payload
2. After registering address for worker, tokens are refreshed, but `Register` button is still there. Seems like some extra step need for updating user state when doing `authService.refreshAccessToken`

packages/apps/human-app/frontend/src/api/http-api-client.ts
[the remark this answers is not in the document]
mpblocky: Sorry, I misunderstood 😓

dnechay (commented):
LGTM

1. Please resolve merge conflicts
2. On example of `OperatorProfileService.enableOperator`, could you please make this and similar methods where we expect "fire & forget" to look like
```
async enableOperator(signature: string) {
  try {
    await authorizedHumanAppApiClient.post(
      apiPaths.enableOperator,
      {
        body: { signature },
      }
    );
  } catch (error) {
    ...
  }
```
so it's clear that it's not expected to handle any return value from it

dnechay (approved):

PR #3278, opened 2025-04-11
feat: add tests for utils and helpers

mpblocky:
## Issue tracking

HUM-165

## Context behind the change

- added tests for utilities

## How has this been tested?

test run

## Release plan

normal

n/a


mpblocky:
@dnechay tests ready

packages/apps/human-app/frontend/src/modules/signup/operator/utils/__tests__/staked-amount-formatter.test.ts
```diff
@@ -0,0 +1,43 @@
+import { describe, expect, it, vi } from 'vitest';
+import { stakedAmountFormatter } from '../staked-amount-formatter';
+
+vi.mock('i18next', () => ({
+  t: (key: string) => (key === 'inputMasks.humanCurrencySuffix' ? 'HMT' : key),
```
dnechay, line 5: Do we need to mock it? Maybe we just can init it?

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/helpers/__tests__/is-hcaptcha-oracle.test.ts
```diff
@@ -0,0 +1,40 @@
+import { describe, expect, it, vi } from 'vitest';
+import { env } from '@/shared/env';
+import { isHCaptchaOracle } from '../is-hcaptcha-oracle';
+
+vi.mock('@/shared/env', () => ({
```
dnechay, line 5: You can have test env defined, not mocked

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/helpers/__tests__/is-hcaptcha-oracle.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  it('should return true when the address matches the hCaptcha oracle address', () => {
+    const result = isHCaptchaOracle(
+      '0x1234567890abcdef1234567890abcdef12345678'
+    );
+    expect(result).toBe(true);
+  });
+
+  it('should return false when the address does not match the hCaptcha oracle address', () => {
+    const result = isHCaptchaOracle(
+      '0xabcdef1234567890abcdef1234567890abcdef12'
```
dnechay, line 21: Should be randomly generated

packages/apps/human-app/frontend/src/modules/smart-contracts/__tests__/check-network.test.ts
```diff
@@ -0,0 +1,65 @@
+import { describe, expect, it, vi } from 'vitest';
+import type { Network } from 'ethers';
+import { env } from '@/shared/env';
+import { MainnetChains, TestnetChains } from '@/modules/smart-contracts/chains';
+import { checkNetwork } from '../check-network';
+
+vi.mock('i18next', () => ({
```
dnechay, line 7: Should be just init, not mocked

packages/apps/human-app/frontend/src/modules/smart-contracts/__tests__/check-network.test.ts
```diff
@@ -0,0 +1,65 @@
...
+  t: (key: string, _options: Record<string, unknown>) => key,
+}));
+
+vi.mock('@/shared/env', () => ({
```
dnechay, line 11: Shouldn't be mocked, should have test env

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/helpers/__tests__/should-navigate-to-registration.test.ts
```diff
@@ -0,0 +1,95 @@
+import { describe, expect, it } from 'vitest';
+import { shouldNavigateToRegistration } from '../should-navigate-to-registration';
+import { type Oracle } from '../../../hooks';
+
+describe('shouldNavigateToRegistration Helper', () => {
```
dnechay, line 5: Please use faker to generate data.
Also you can DRY oracle data generation to some function in this test file

dnechay (commented):
Let's keep tests files next to what they test instead of `__tests__` folders, i.e.
```
- date.ts
- date.spec.ts
```
It's just common approach we have in other places

packages/apps/human-app/frontend/src/shared/helpers/__tests__/date.test.ts
```diff
@@ -0,0 +1,53 @@
+import { describe, expect, it } from 'vitest';
+import { formatDate, parseDate, getTomorrowDate } from '../date';
+
+describe('Date Helper Functions', () => {
+  describe('formatDate', () => {
+    it('should correctly format a date string', () => {
+      const dateString = '2023-01-15T12:30:45Z';
+      const formattedDate = formatDate(dateString);
+      expect(formattedDate).toMatch(/^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}$/);
```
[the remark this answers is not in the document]
mpblocky: Actually - `parseDate` takes into account the time zone. Should we still test the string? IMHO regex is good enough.

packages/apps/human-app/frontend/src/modules/worker/jobs/my-jobs/utils/__tests__/get-chip-status-color.test.ts
```diff
@@ -0,0 +1,82 @@
+import { describe, expect, it } from 'vitest';
+import type { ColorPalette } from '@/shared/styles/color-palette';
+import { getChipStatusColor } from '../get-chip-status-color';
+import { MyJobStatus, UNKNOWN_JOB_STATUS } from '../../../types';
+
+describe('getChipStatusColor Function', () => {
+  const mockColorPalette: ColorPalette = {
```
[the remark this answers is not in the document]
mpblocky: We don't :)

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/helpers/__tests__/is-hcaptcha-oracle.test.ts
mpblocky: I think mocking is better:
- we don't need to define every variable
- we clearly see what are we testing against
- easily can set variables

packages/apps/human-app/frontend/src/modules/smart-contracts/__tests__/get-network-name.test.ts
```diff
@@ -0,0 +1,49 @@
+import { describe, expect, it, vi } from 'vitest';
+import { env } from '@/shared/env';
+import {
+  AllTestnetsChains,
+  AllMainnetChains,
+} from '@/modules/smart-contracts/chains';
+import { getNetworkName } from '../get-network-name';
+
+vi.mock('@/shared/env', () => ({
```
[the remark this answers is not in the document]
mpblocky: As in [here](https://github.com/humanprotocol/human-protocol/pull/3278#discussion_r2044118550)

packages/apps/human-app/frontend/src/modules/smart-contracts/__tests__/check-network.test.ts
mpblocky: Lets continue in [here](https://github.com/humanprotocol/human-protocol/pull/3278#discussion_r2044118550)

packages/apps/human-app/frontend/src/shared/helpers/__tests__/date.test.ts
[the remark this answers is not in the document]
dnechay: What if by mistake `formatDate` alters the date and formats it in invalid way, like `2025-30-30 ....`? This test will pass
I would say that this test should do an exact match.

In case you would be generating random date for this test - then regexp should be fine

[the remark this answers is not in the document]
dnechay: @mpblocky What if by mistake `formatDate` alters the date and formats it in invalid way, like `2025-30-30 ....`? This test will pass
I would say that this test should do an exact match.

In case you would be generating random date for this test - then regexp should be fine

packages/apps/human-app/frontend/src/modules/worker/jobs/my-jobs/utils/__tests__/get-chip-status-color.test.ts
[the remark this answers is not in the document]
dnechay: So let's not mock it :)

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/helpers/__tests__/is-hcaptcha-oracle.test.ts
dnechay: @mpblocky agree with you on the above. In this specific case - it's fine to have `@/shared/env` mocked, because for this kind of tests we can assume that `@/shared/env` is correct and we test a different thing.

At the same time I would like to be sure that variables provided by `@/shared/env` are correct (i.e. we can test env schema validation separately). So it should be a separate test, probably using [vi.stubEnv](https://vitest.dev/api/vi.html#vi-stubenv)

mpblocky: I will add these tests

mpblocky: I'm not sure if it's something we should focus on; we are using `zod` to validate envs, so testing it would result in just testing `zod` library. Do You want me to test `env.ts` file?

packages/apps/human-app/frontend/src/shared/helpers/__tests__/date.test.ts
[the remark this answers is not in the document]
dnechay: Ideally we should ensure that this utility only formats the date and doesn't alter its value.
It can be easily done by:
```
1. generate some random date
2. call `randomDate.setMilliseconds(0)`
3. Then you format it
4. get a date from `new Date(formattedDate)`
5. compare `randomDate.toISOString() === formattedDateObj.toISOString()`
```

At the same time - random + regexp is fine to ensure it corresponds to a format. So up to you how specific you would like to be about this test

packages/apps/human-app/frontend/src/modules/worker/jobs-discovery/helpers/__tests__/is-hcaptcha-oracle.test.ts
dnechay: Not now, I suppose that we might come up with some common approach for all services later, so was just thinking out loud

dnechay (commented):
LGTM

Please wrap-up [this discussion](https://github.com/humanprotocol/human-protocol/pull/3278#discussion_r2044280150) and it's good to go

dnechay (approved):

PR #3282, opened 2025-04-14
[HUMAN App] fix: operator role values

dnechay:
## Issue tracking
N/A

## Context behind the change
Atm on UI we have it's own list of human-readable options, when in KV store we rely on snake-cased value. Fixing the form to take actual values from sdk

## How has this been tested?
- [ ] update operator role while sign-up, read KV store and make sure snake-cased value is written

## Release plan
Just merge

No


KirillKirill (approved):
lgtm

PR #3284, opened 2025-04-16
chore: remove boilerplate for auth contexts

mpblocky:
## Issue tracking

HUM-142

## Context behind the change

- reduced boilerplate from contexts

## How has this been tested?

- logged in both ways and clicked around

## Release plan

normal

n/a

### DIFF
--- a/packages/apps/human-app/frontend/src/modules/auth/context/auth-context.tsx
+++ b/packages/apps/human-app/frontend/src/modules/auth/context/auth-context.tsx
@@ -1,14 +1,7 @@
-/* eslint-disable camelcase -- ... */
-import { useState, createContext, useEffect } from 'react';
-import { jwtDecode } from 'jwt-decode';
+/* eslint-disable camelcase */
+// web2-auth.tsx
 import { z } from 'zod';
-import { useQueryClient } from '@tanstack/react-query';
-import { browserAuthProvider } from '@/shared/contexts/browser-auth-provider';
-import {
-  ModalType,
-  useModalStore,
-} from '@/shared/components/ui/modal/modal.store';
-import { type AuthTokensSuccessResponse } from '@/shared/schemas';
+import { createAuthProvider } from '@/shared/contexts/generic-auth-context';
 
 const extendableUserDataSchema = z.object({
   site_key: z.string().optional().nullable(),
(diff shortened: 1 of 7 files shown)


mpblocky:
@dnechay ready!

mpblocky:
I found some more improvements, will let You know soon!

mpblocky:
@dnechay changes added

packages/apps/human-app/frontend/src/modules/auth/context/auth-context.tsx
```diff
@@ -22,133 +15,14 @@ const userDataSchema = z
     email: z.string(),
     user_id: z.number(),
     reputation_network: z.string(),
-    email_notifications: z.boolean().optional(), // TODO that should be verified when email notifications feature is done
+    email_notifications: z.boolean().optional(),
     exp: z.number(),
   })
   .merge(extendableUserDataSchema);
```
dnechay, line 21: Why merging these two and not just `userDataSchema`?

packages/apps/human-app/frontend/src/shared/contexts/generic-auth-context.tsx
```diff
@@ -0,0 +1,152 @@
...
+        const currentAuthType = browserAuthProvider.getAuthType();
+        const savedUserData = browserAuthProvider.getUserData();
+
+        if (!accessToken || currentAuthType !== authType) {
+          setAuthState({ user: null, status: 'idle' });
+          return;
+        }
+        const userData = jwtDecode(accessToken);
+        const userDataWithSavedData = savedUserData.data
+          ? { ...userData, ...savedUserData.data }
```
dnechay, line 80: Why do we need to merge these two? Shouldn't jwt payload have latest valid data to use?

packages/apps/human-app/frontend/src/shared/contexts/generic-auth-context.tsx
```diff
@@ -0,0 +1,152 @@
...
+        });
+      }
+    };
+
+    const signIn = (signInSuccess: AuthTokensSuccessResponse) => {
+      browserAuthProvider.signIn(signInSuccess, authType);
+      handleSignIn();
+    };
+
+    const signOut = ({
```
dnechay, line 102: This function has new signature now and because of that build fails since it's called with invalid args in different places.
Don't forget to provide proper defaults when you fix it

dnechay (commented):
Looks better then it was :)

packages/apps/human-app/frontend/src/modules/auth/context/auth-context.tsx
mpblocky: Missed that while moving code, thanks!

packages/apps/human-app/frontend/src/shared/contexts/generic-auth-context.tsx
```diff
@@ -0,0 +1,152 @@
...
+}
+
+export function createAuthProvider<T extends UserData | Web3UserData>(config: {
+  authType: AuthType;
+  schema: ZodType<T>;
+}) {
+  const { authType, schema } = config;
+  const AuthContext = createContext<
+    AuthenticatedUserContextType<T> | UnauthenticatedUserContextType<T> | null
+  >(null);
```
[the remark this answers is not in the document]
mpblocky: Yep, sounds good

packages/apps/human-app/frontend/src/shared/contexts/generic-auth-context.tsx
mpblocky: It actually was a previous implementation, I thought it had it's reasons to be there 😅 I will remove it and test if everything is alright, seemed weird to me too

dnechay: If there will be surprises - we will be able to detect them and either fix or drop some comments 👌🏻

packages/apps/human-app/frontend/src/shared/contexts/generic-auth-context.tsx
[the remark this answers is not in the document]
mpblocky: Actually, we have checks for null values in contexts, e.g. in `useAuth` so I don't think we should change anything here. WDYT?

mpblocky:
@dnechay all addressed

packages/apps/human-app/frontend/src/shared/contexts/generic-auth-context.tsx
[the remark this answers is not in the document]
dnechay: IMO it's just easier to digest when you have specific defaults rather than `null`, but since it's common thing - let's keep it as is

dnechay (approved):

Issue #2803, opened 2024-11-18
[Reputation Oracle] Revision/Refactoring


Dzeranov:
**Description**
Need to do a revision of the Reputation Oracle. This is complex and ongoing process. All the necessary actions will be described below.

**Action points**
- [x] #2928
- [x] #3064
- [x] #3065
- [x] #3084
- [x] Revisit snake case transformers (keys return in camelcase in validation errors)
- [x] Revisit Github Actions: make sure jobs names are clear and necessary caching ([PR](https://github.com/humanprotocol/human-protocol/pull/3316))
- [x] Avoid loading the full user from DB for every request when doing auth ([PR](https://github.com/humanprotocol/human-protocol/pull/3322))
- [x] Get rid of HUMAN App in DB ([PR](https://github.com/humanprotocol/human-protocol/pull/3339))
- [x] Setup up `CASCADE` on DB relations for proper deletions ([PR](https://github.com/humanprotocol/human-protocol/pull/3340))
- [x] ~Split user & operator to different tables~ (icebox: no viable profit for performance, but it's heavy lifting)
- [x] ~Change primary ids of entities to be `uuid` instead of `SERIAL`~ (icebox: we have space for maneuver here; if we are close to the limits - change column type to `bigint` and work on it; `tokens` table is number 1 candidate)
- [x] Add TTL on refresh tokens in DB ([PR](https://github.com/humanprotocol/human-protocol/pull/3343))

PR #3322, opened 2025-05-07
[Reputation Oracle] Remove fetching user from a db in a guard

Dzeranov:
## Issue tracking

Part of #2803 
## Context behind the change

Removed user fetching from JwtHttpStrategy. `userId` is being passed instead and handled by services accordingly
## How has this been tested?

- [x] unit tests
- [x] deployed locally and tested affected endpoints

## Release plan

N/A

Just in case, need to monitor affected endpoints

### DIFF
--- a/packages/apps/reputation-oracle/server/src/common/types/request.ts
+++ b/packages/apps/reputation-oracle/server/src/common/types/request.ts
@@ -1,3 +1,3 @@
 export interface RequestWithUser extends Request {
-  user: any;
+  user: number;
 }
--- a/packages/apps/reputation-oracle/server/src/common/types/request.ts
+++ b/packages/apps/reputation-oracle/server/src/common/types/request.ts
@@ -1,3 +1,3 @@
 export interface RequestWithUser extends Request {
-  user: number;
+  user: { id: number };
 }
--- a/packages/apps/reputation-oracle/server/src/modules/auth/auth.service.ts
+++ b/packages/apps/reputation-oracle/server/src/modules/auth/auth.service.ts
@@ -422,7 +422,9 @@ export class AuthService {
   }
 
   async resendEmailVerification(userId: number): Promise<void> {
-    const user = await this.userRepository.findOneById(userId);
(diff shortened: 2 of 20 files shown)


packages/apps/reputation-oracle/server/src/common/types/request.ts
```diff
@@ -1,3 +1,3 @@
 export interface RequestWithUser extends Request {
-  user: any;
+  user: number;
```
dnechay, line 2: Maybe we have it as
```
user: { id: number }
```
?
Current naming a bit confusing tbh when you use it

packages/apps/reputation-oracle/server/src/modules/auth/auth.service.ts
```diff
@@ -421,8 +421,10 @@ export class AuthService {
     await this.tokenRepository.deleteOne(tokenEntity);
   }
 
-  async resendEmailVerification(user: Web2UserEntity): Promise<void> {
-    if (user.status !== UserStatus.PENDING) {
+  async resendEmailVerification(userId: number): Promise<void> {
+    const user = await this.userRepository.findOneById(userId);
```
dnechay, line 425:
```suggestion
    const user = (await this.userRepository.findOneById(userId)) as Web2UserEntity;
```

and remove `as string` below for `user.email`

packages/apps/reputation-oracle/server/src/modules/user/user.service.spec.ts
```diff
@@ -267,13 +278,17 @@ describe('UserService', () => {
 
     it('should throw if same address already exists', async () => {
       const user = generateWorkerUser();
+      const existingUser = generateWorkerUser({
+        privateKey: generateEthWallet().privateKey,
+      });
```
dnechay, line 283:
```suggestion
      const existingUser = generateWorkerUser();
```

dnechay (approved):
LGTM
Some suggestions to discuss

dnechay (approved):

Issue #3321, opened 2025-05-06
Bump `yarn` version for the repository


dnechay:
In our monorepo we have fixed `yarn` version - 1.19.0 - [that was hardcoded long time ago](https://github.com/humanprotocol/human-protocol/pull/347). Since then many things changed and latest yarn version is 4.9.1. It got a lot of improvements to workspaces and dependency installation process. In our GitHub Actions that have "install dependencies" step we spend ~2 minutes on yarn "Linking dependencies" step and probably the only way to improve it is to bump yarn itself.

**Describe the solution you'd like**
Attempt to bump `yarn` to the latest version - 4.9.1. If by some reason we can't in our repo - bump to the version that won't have issues with our setup

PR #3324, opened 2025-05-08
chore: upgrade yarn to latest

dnechay:
## Issue tracking
#3321 

> [!TIP]
> In case you get `Duplicate workspace name` error when running yarn scripts - please remove `dist` folders for `reputation-oracle/server` and `human-app/server`. It's fixed as part of this PR but you have to manually clean existing builds on your machine 
 
- in root `package.json` file should be only those dependencies that are used for the whole project (i.e. root workspace), but we don't have such, so all them removed
- in docker we do not use `--immutable` flag because it fails to build; the reason is that in all dockerfiles we copy only necessary `package.json` files (in order to avoid excessive deps & changes and get better caching) and `yarn.lock` gets modified because no other workspaces present; it's fine because in CI we still use `--immutable`
- `postinstall` script for root workspace runs only once: when you do a fresh install of dependencies (i.e. no `node_modules` and stuff), so decided to replace it with `yarn build:libs` script; it also remove the "burden" of building `core` and `sdk` on any deps install
- build process for `core` and `sdk` is split now; it's done in order to improve DX, because we don't touch `core` package a lot and you can build it once, later rebuilding just `sdk` or any other necessary package; but they both still referenced as `workspace:*`, i.e. apps use version from repo instead of what published in npm registry; we might want to change it later and use fixed version installation in apps
- `node_modules` linked is used instead of "Plug-n-Play" in order to get things updated with minimal risk of breaks; some packages might not support it, so it's kind of "common practice"
- `package.json` usage in healthchecks removed because if used - then `package.json` is included in `dist` folder after build in `yarn@4` it's not possible to exclude something in globs, so it find `dist/package.json` as well and considers it as "duplicate" of workspace, failing to run any script
- no need in root `tsconfig.json` because we don't have root project with references; replaces it with `tsconfig.base.json`; probably later we can also remove some other stuff like `eslint`, `jest` and etc.
- added `clean` scripts for convenience
- fixed lint errors in `gcv`
- `yarn.lock` got fully recreated, so it might have some side-effects

## How has this been tested?
- [x] `yarn install` on clean repo finishes w/o errors
- [x] go to every workspace and run `lint`, `test` and `build` commands
- [x] run locally w/ Docker 
- [x] all CI checks are green

## Release plan
Merge & release as a separate PR/change

⚠️ We will need to update build command on render from
```
yarn install; yarn workspace @human-protocol/sdk build; yarn build
```
to
```
yarn install; yarn workspace human-protocol build:libs; yarn build
```

Potential side-effects in different workflows due to dependencies bump. Critical flows (like payouts) use exact `ethers` version, so should be good. For anything else we should be able to notice it on STG if happens.

Also dependabot might not work 100% because they still adding full support of yarn v4, but we don't use any unusual feature, so probably we are fine here.

> [!IMPORTANT]
> ~`Dependency Review` fails to run due to very big result. We might want to open a follow-up PR with some package bump later in order to get a valid scorecard for new packages.~ Fixed by upgrading GA and changing configuration to not display scorecard (we don't check it atm anyways)


Dzeranov (approved):
lgtm

portuu3 (dismissed):

flopez7 (commented):
Great job, code looks good. However, since it's a major refactor, I think it would be a good idea to regenerate `yarn.lock` to ensure all dependencies are up-to-date and no residual packages remain

dnechay:
@flopez7 I would suggest upgrading packages in a different PR because it may cause issues in build and how they work.
E.g. once I got `class-validator` updated from `0.14.1` to `0.15.0` - validation tests got broken in some services. Also there might be incompatible types between packages and peer dependencies: e.g. `hardhat` strictly tied to `ethers@6.13.5`, `6.14` is not compatible in types and we can't build some sdk

So I would suggest delivering this first, make sure it works fine - and then we are free to do any updates as a follow-up, wdyt?

flopez7:
Sure, let's do it that way!

PR #3326, opened 2025-05-12
[Job Launcher] Refactor error handling

flopez7:
## Issue tracking
N/A

## Context behind the change
Refactor error handling across services to use custom error classes:
- Replaced instances of ControlledError with specific error classes (ValidationError, NotFoundError, ConflictError, ServerError) in QualificationService, RateService, RoutingProtocolService, SendGridService, StorageService, Web3Service, WebhookService, and their respective test files.
- Removed the storage.errors.ts file as its functionality is now integrated into the common error handling.
- Updated tests to reflect changes in error handling and ensure proper error types are thrown and caught.

## How has this been tested?
- Deployed locally and called some endpoints to see if custom errors were thrown. 
- Ran unit tests

## Release plan
None

Check that the new errors are not instance of ControlledError


packages/apps/job-launcher/server/src/common/utils/index.ts
```diff
@@ -76,7 +75,7 @@ export const parseUrl = (
     }
   }
 
-  throw new ControlledError('Invalid URL', HttpStatus.BAD_REQUEST);
+  throw new ValidationError('Invalid URL');
```
dnechay, line 78: Maybe just `new Error` here?

packages/apps/job-launcher/server/src/common/utils/storage.ts
```diff
@@ -30,28 +30,19 @@ export function generateBucketUrl(
     storageData.provider != StorageProviders.GCS &&
     storageData.provider != StorageProviders.LOCAL
   ) {
-    throw new ControlledError(
-      ErrorBucket.InvalidProvider,
-      HttpStatus.BAD_REQUEST,
-    );
+    throw new ValidationError(ErrorBucket.InvalidProvider);
```
dnechay, line 33: Are all these `ValidationError` http-related? Maybe just `new Error`?

packages/apps/job-launcher/server/src/modules/content-moderation/gcv-content-moderation.service.ts
```diff
@@ -379,10 +376,7 @@ export class GCVContentModerationService implements IContentModeratorService {
 
       const [files] = await bucket.getFiles({ prefix: bucketPrefix });
       if (!files || files.length === 0) {
-        throw new ControlledError(
-          ErrorContentModeration.NoResultsFound,
-          HttpStatus.NOT_FOUND,
-        );
+        throw new NotFoundError(ErrorContentModeration.NoResultsFound);
```
dnechay, line 379: Maybe just `new Error`?

packages/apps/job-launcher/server/src/modules/storage/storage.service.ts
```diff
@@ -54,9 +53,9 @@ export class StorageService {
       return Buffer.from(data);
     } catch (error) {
       if (error.response?.status === HttpStatus.NOT_FOUND) {
-        throw new FileNotFoundError(url);
+        throw new NotFoundError(`${ErrorStorage.NotFound}: ${url}`);
```
dnechay, line 56: It's confusing to have `NotFoundError` here because it might end up to 404, when it's not, it's about processing error, which is more about 422 or 500

packages/apps/job-launcher/server/src/modules/webhook/webhook.service.ts
```diff
@@ -81,7 +84,7 @@ export class WebhookService {
 
     // Check if the request was successful.
     if (status !== HttpStatus.CREATED && status !== HttpStatus.OK) {
```
dnechay, line 86: `httpService.post` will throw `AxiosError` in case it's not `2xx` code, so it makes more sense to handle it like [this](https://github.com/humanprotocol/human-protocol/blob/8261d897ce2b9a4f0288a4ab68f547592cdf69b2/packages/apps/reputation-oracle/server/src/modules/webhook/webhook-outgoing.service.ts#L81-L95) and throwing just `Error`, because it's not http-facing service

packages/apps/job-launcher/server/src/modules/webhook/webhook.service.ts
```diff
@@ -158,10 +160,7 @@ export class WebhookService {
     );
 
     if (!jobEntity) {
-      throw new ControlledError(
-        ErrorWebhook.InvalidEscrow,
-        HttpStatus.BAD_REQUEST,
-      );
+      throw new NotFoundError(ErrorWebhook.InvalidEscrow);
```
dnechay, line 163: Same for this and above: it can be just `new Error` because these are called in async flows, not in http

dnechay (commented):
Looks cleaner! Left some questions and suggestions

- in all places that are not directly used by http layer (e.g. `common/utils` or `storage.service`) we probably should not use HTTP-related Errors, it's a bit confusing to see things like `ConflictError` when you validate web3 signature and not intuitive for developer from the future on how to pick some specific error class from all of these
- I noticed that you changed "http-response code" for some cases (e.g. in `job.service` some places have 409/422 vs 400 before), can it affect something on UI?
-

packages/apps/job-launcher/server/src/common/guards/whitelist.auth.ts
```diff
@@ -16,14 +11,14 @@ export class WhitelistAuthGuard implements CanActivate {
     const user = request.user;
 
     if (!user) {
```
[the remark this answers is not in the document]
flopez7: I believe the current implementation is better. A guard should be generic and return the appropriate error code, like 401 in this case, to clearly indicate an authorization issue. While the stack trace can help identify the source of the error, having an specific error code makes debugging and understanding the issue much easier.

If we remove the user check and let it fail with a 500 error, it could lead to confusion and make the guard less intuitive. A 500 error typically indicates a server-side issue, not an authorization problem, which might mislead developers. So, I suggest keeping the user check as it is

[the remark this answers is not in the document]
dnechay: @flopez7 
But if we use this guard on some route that doesn't have jwt authorization - it's not authorization issue (401), it's about wrong implementation on our end, which is 500 and we should know about it.

As a simple example, imagine we have some metric on http status code, e.g. notify us in slack when number of 5xx is more than N. In case we just throw here or return 500 - we will know about wrong guard usage with such alert, but if it's 401 - we won't. It's simply because you don't want to get alerted on 4xx codes, because it's user error

packages/apps/job-launcher/server/src/common/pipes/validation.ts
```diff
@@ -1,27 +1,23 @@
...
-import { ControlledError } from '../errors/controlled';
+import { ValidationError } from '../errors';
 
 @Injectable()
 export class HttpValidationPipe extends ValidationPipe {
   constructor(options?: ValidationPipeOptions) {
     super({
-      exceptionFactory: (errors: ValidationError[]): ControlledError => {
+      exceptionFactory: (errors: ValidError[]): ValidationError => {
         const errorMessages = errors
```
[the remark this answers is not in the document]
flopez7: IMO, we must leave forbidNonWhitelisted and forbidUnknownValues, that way we make sure that the data processed is exactly what we have already defined as DTOs.

packages/apps/job-launcher/server/src/common/utils/decimal.ts
```diff
@@ -16,10 +15,7 @@ export function div(a: number, b: number): number {
   const decimalB = new Decimal(b);
 
   if (decimalB.isZero()) {
-    throw new ControlledError(
-      'Division by zero is not allowed.',
-      HttpStatus.CONFLICT,
-    );
+    throw new ConflictError('Division by zero is not allowed.');
```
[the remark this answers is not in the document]
flopez7: We're using it in methods like createJob and createFiatPayment, which are directly called by the user. For this reason, I prefer to return a validation error here instead of just a generic Error. This way, the user will receive the proper 400 exception

packages/apps/job-launcher/server/src/common/pipes/validation.ts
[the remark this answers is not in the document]
dnechay: Ok, that makes sense here
In RepO we accept webhooks from third-parties and don't need all data from them for processing, so had to disable it in use proper projections. If JL has full control over its data - it fine to have strict validation

packages/apps/job-launcher/server/src/common/utils/index.ts
flopez7: It is being used to verify a url coming from a DTO, I will leave it as a ValidationError

packages/apps/job-launcher/server/src/common/utils/storage.ts
flopez7: The generateBucketUrl method is used when generating manifests, which are created directly from user input. In this context, the ValidationError is used to ensure that invalid user input is caught early and returned as a proper 400 response, which is more informative than a generic 500 error. Using new Error would make it harder to distinguish between validation issues and unexpected server errors

dnechay: Taking into account that `ValidationError` might be considered not only as "http-error" - sounds reasonable, yes

packages/apps/job-launcher/server/src/modules/webhook/webhook.service.ts
flopez7: Even though these methods are called in async flows, they are originated from the createWebhook endpoint, which is HTTP-based. I believe we should return the appropriate error, like ValidationError or InvalidEscrow, to ensure the calling system (in this case, the oracle, but could be anyone) can handle it properly. Using a generic Error would make it harder to distinguish the nature of the issue and could complicate debugging or error handling on the caller's side.

packages/apps/job-launcher/server/src/common/guards/whitelist.auth.ts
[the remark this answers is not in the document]
flopez7: Since it's a guard, I believe it should always return the appropriate error (401) to clearly indicate an authentication issue. Returning a 500 could confuse users, as it suggests a server-side problem rather than an authorization failure. We can add internal logs to catch misuses instead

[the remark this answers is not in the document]
dnechay: Let's add log and keep it as is then 👍🏻

packages/apps/job-launcher/server/src/common/utils/signature.ts
```diff
@@ -47,9 +43,6 @@ export function recoverSigner(
   try {
     return ethers.verifyMessage(message, signature);
   } catch (_error) {
-    throw new ControlledError(
-      ErrorSignature.InvalidSignature,
-      HttpStatus.CONFLICT,
-    );
+    throw new ConflictError(ErrorSignature.InvalidSignature);
```
dnechay, line 46: Might it be better to have it as `ValidationError` as well?

dnechay (dismissed):

dnechay:
Please double-check that client-facing (I mean FE app) functionality is not broken in places where status codes changed

flopez7:
> Please double-check that client-facing (I mean FE app) functionality is not broken in places where status codes changed

@dnechay In FE we use simple try -catch without specifying error code, so should be fine. Only place we're checking error code is to refresh auth token and we didn't change there the error returned

portuu3 (approved):

PR #3342, opened 2025-05-15
[Recording Oracle] Implement custom error handling

flopez7:
## Issue tracking
N/A

## Context behind the change
- Replaced all HTTP Exceptions with custom error classes (AuthError, ValidationError, NotFoundError, ConflictError, ServerError) across various guards, services, and repositories for improved error management.
- Introduced a centralized error handling mechanism with ExceptionFilter to standardize HTTP responses for different error types.
- Removed some usused code as IsValidEthereumAddress decorator and GlobalExceptionFilter

## How has this been tested?
- Ran unit tests
- Deployed locally and created some webhooks for tests

## Release plan
None

Check that the returned message in case of error is different, it must have statusCode, timestamp, message and path. Also, now it's being logged with the exception received


packages/apps/fortune/recording-oracle/src/common/exceptions/exception.filter.ts
```diff
@@ -0,0 +1,60 @@
...
+    const status = this.getStatus(exception);
+    const message = exception.message || 'Internal server error';
+
+    this.logger.error(
+      `Exception caught: ${message}`,
+      exception.stack || 'No stack trace available',
+    );
+
+    response.status(status).json({
+      statusCode: status,
```
dnechay, line 54:
```suggestion
      status_code: status,
```
This is not transformed to snake case due to order of middlewares in nest app

packages/apps/fortune/recording-oracle/src/common/guards/signature.auth.ts
```diff
@@ -42,11 +38,10 @@ export class SignatureAuthGuard implements CanActivate {
       if (isVerified) {
         return true;
       }
-    } catch (error) {
-      // eslint-disable-next-line no-console
-      console.error(error);
+    } catch (error: any) {
+      throw new ValidationError(error.message);
```
dnechay, line 42: Probably it shouldn't be 400: imagine there is a network error while getting escrow data, it's 500, not 400 then
Would be better to log (with common logger) and re-throw or even remove try-catch at all

packages/apps/fortune/recording-oracle/src/common/interceptors/transform-enum.interceptor.ts
```diff
@@ -72,7 +72,7 @@ export class TransformEnumInterceptor implements NestInterceptor {
     // Validate the transformed data
     const validationErrors = validateSync(transformedInstance);
     if (validationErrors.length > 0) {
-      throw new BadRequestException('Validation failed');
+      throw new AuthError('Validation failed');
```
dnechay, line 75: Should be `ValidationError`?

packages/apps/fortune/recording-oracle/src/common/utils/webhook.ts
```diff
@@ -1,11 +1,20 @@
...
+import { Logger } from '@nestjs/common';
+import { AxiosError } from 'axios';
+import { firstValueFrom } from 'rxjs';
+import { WebhookDto } from '../../modules/webhook/webhook.dto';
 import { HEADER_SIGNATURE_KEY } from '../constants';
 import { CaseConverter } from './case-converter';
-import { WebhookDto } from '../../modules/webhook/webhook.dto';
+import { signMessage } from './signature';
+
+export function formatAxiosError(error: AxiosError) {
```
dnechay, line 10: Maybe move it to some common utils?

dnechay (approved):

PR #3394, opened 2025-06-16
[Human App] Implement JWT authentication guard and strategy to validate the token

flopez7:
## Context behind the change
- Added JwtAuthGuard to protect routes and handle JWT validation.
- Created JwtHttpStrategy for extracting and validating JWT tokens.
- Introduced RequestWithUser interface to standardize request objects with user data and token.
- Updated various controllers to utilize the new JWT authentication mechanism, ensuring token extraction from requests.
- Refactored existing endpoints to replace direct token handling with the new request structure.
- Enhanced tests to accommodate changes in controller methods and request handling.
- Updated `jest.config.ts` to improve test usage 

## How has this been tested?
Deployed locally, generated a new JWT and call some open and closed endpoints

## Release plan
None

Check if all jwt endpoints required still working.


packages/apps/human-app/server/src/integrations/kv-store/kv-store.gateway.ts
```diff
@@ -88,4 +97,45 @@ export class KvStoreGateway {
       return jobTypes;
     }
   }
+
+  async getReputationOraclePublicKey(
```
dnechay, line 101: This should only get the pubkey from kv and not have implicit caching. Let's just fetch it here and cache in-memory where used (i.e. in guard itself) w/o any TTL because we don't have even rotation for this value, so something like:
```
const keysPerChain = new Map<number, string>();

let pubkey: string;
if (keysPerChain.has(chainId)) {
  pubkey = keysPerChain.get(chainId);
} else {
  pubkey = await kvgateway.getReputationOraclePublicKey(args);
  keysPerChain.set(chainId, pubkey)
}
```
should be enough

packages/apps/human-app/server/src/common/guards/jwt.auth.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Injectable()
+export class JwtAuthGuard extends AuthGuard('jwt-http') implements CanActivate {
+  constructor(private readonly reflector: Reflector) {
+    super();
+  }
+
+  public async canActivate(context: ExecutionContext): Promise<boolean> {
+    // Check for public routes first
+    const isPublic = this.reflector.getAllAndOverride<boolean>('isPublic', [
```
dnechay, line 19: We have this check, but don't have such decorator/metadata.
Let's define this decorator and use it in `getAllAndOverride` (see RepO for example)

packages/apps/human-app/server/src/integrations/kv-store/kv-store.gateway.ts
flopez7: I have added the cache because other methods also do this here too

dnechay: @flopez7 I see, let's be consistent then

packages/apps/human-app/server/src/common/guards/strategy/jwt.http.ts
```diff
@@ -0,0 +1,73 @@
...
+      exp: number;
+    },
+  ): Promise<JwtUserData> {
+    console.log('JWT payload in validate:', payload);
+
+    if (!payload.email) {
+      throw new UnauthorizedException('Invalid token: missing email');
+    }
+
+    return {
```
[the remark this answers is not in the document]
dnechay: @flopez7 and that's the tricky part: JWT payload is different based on the user type, e.g. operator used doesn't have `qualifications` at all.

It is better to keep it lean, so we don't have to remember about this types and maintain it when remove\add some unnecessary fields

dnechay (approved):

Issue #3478, opened 2025-07-24
[SDK] Allow manifest to be a json string


portuu3:
**Description**
Allow manifest to be a json string

**Motivation**
Allow passing manifest data instead of a url to a file

**Implementation details** 
- Remove url validation in setup
- Rename parameter name
- Rename getManifestUrl to getManifest

PR #3479, opened 2025-07-24
[SDK] Allow manifest to be a json string 

flopez7:
## Issue tracking
#3478 

## Context behind the change
Rename manifest_url to manifest and update related methods (Setup and getManifest) in SDK

## How has this been tested?
- Ran unit tests
- Deployed one escrow and set it up using SDK

## Release plan
Deploy new SDK version
Deploy new subgraph

None


portuu3 (dismissed):
Rename also subgraph entities

dnechay (commented):
SDK changes lgtm, but we still need to apply them in Reputation Oracle and Dashboard Server

flopez7:
> ~There is another place that needs rename of `manifestUrl` to `manifest`: [ref](https://github.com/humanprotocol/human-protocol/blob/1a479d54496a86c8833c1de6c1fa4745852b6c84/packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts#L131)~ UPD: above not actual, I just missed the pull
> 
> I would suggest to export `IEscrow` from sdk, so in test we can create a generator with proper typing that will help us easily find such places in the future (now TS can't detect invalid types due to `any` usage)

`IEscrow` is already exported [https://github.com/humanprotocol/human-protocol/blob/develop/packages/sdk/typescript/human-protocol-sdk/src/interfaces.ts](here)

dnechay:
@flopez7 is it finalized?

flopez7:
@dnechay Yes, @portuu3 was deploying it in subgraph before merging it

dnechay (approved):

PR #3507, opened 2025-08-13
feat: improve logs & services config

dnechay:
## Issue tracking
Freestyle

## Context behind the change
1. added `version` to all services logs in order to easily filter by version in DD
2. adjusted migrations commands and config for JL & Fortune to avoid unnecessary builds
3. removed unused "serve-static" in JL, because it's not necessary for swagger
4. updated `logger` lib with some convenience tooling
5. noisy cron logs from `info` to `debug` as we don't need them in DD (at least now)
6. added ability to override log level in services via `LOG_LEVEL` env variable

Decided to not apply tsconfig & eslint updates for now because it's going to cause way too many changes.

## How has this been tested?
- [x] run service, check that logs output has proper `version` binding now
- [x] run RepO & HUMAN App, check that `/health/ping` returns `version`
- [x] run JL and open swagger, make sure it works
- [x] run JL and Fortune migrations commands, make sure they work
- [x] run RepO with `LOG_LEVEL=info yarn start`, make sure that cron jobs are not logged

## Release plan
- [x] update start command for services on Render to include `GIT_HASH=$RENDER_GIT_COMMIT`
- [x] release new `logger` version

No

### DIFF
--- a/packages/apps/dashboard/server/src/logger/index.ts
+++ b/packages/apps/dashboard/server/src/logger/index.ts
@@ -1,13 +1,27 @@
-import { createLogger, NestLogger, LogLevel } from '@human-protocol/logger';
+import {
+  createLogger,
+  NestLogger,
+  LogLevel,
+  isLogLevel,
+} from '@human-protocol/logger';
 
 import Environment from '../common/utils/environment';
 
 const isDevelopment = Environment.isDevelopment();
 
+const LOG_LEVEL_OVERRIDE = process.env.LOG_LEVEL;
+
+let logLevel = LogLevel.INFO;
+if (isLogLevel(LOG_LEVEL_OVERRIDE)) {
+  logLevel = LOG_LEVEL_OVERRIDE;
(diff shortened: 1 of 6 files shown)


packages/apps/job-launcher/server/src/app.module.ts
```diff
@@ -82,13 +80,6 @@ import { WebhookModule } from './modules/webhook/webhook.module';
     WebhookModule,
     StatisticModule,
     QualificationModule,
-    ServeStaticModule.forRoot({
-      rootPath: join(
-        __dirname,
-        '../../../../../../',
-        'node_modules/swagger-ui-dist',
-      ),
-    }),
```
flopez7, line 91: Not sure if we can delete this. I remember we had to add it because when deploying the server in Vercel, it caused problems because it couldn't find some Swagger files, and we had to specify the path manually. We can try to see if that error has been fixed already

flopez7 (approved):
LGTM 👍

Issue #3444, opened 2025-07-03
[HUMAN App] Use the backend to get information of governance proposals


portuu3:
**Description**
Instead of fetching the governance logs from the frontend, use the backend to get information for governance proposals banner.

**Motivation**
Using the backend we can use paid RPCs without disclosing API keys

**Implementation details** 
Cache the proposal info with a 15 min TTL

PR #3509, opened 2025-08-14
[Human app] Add proposals endpoint and update frontend

flopez7:
## Issue tracking
#3444 

## Context behind the change

- Add governance module with new endpoint for getting active and pending proposals and cache them.
- Update frontend to query new endpoint instead of querying RPC.
- Show pending proposals in banner too.

## How has this been tested?
Deployed locally on testnet

## Release plan

- Add new `GOVERNANCE_RPC_URL` and `GOVERNOR_ADDRESS` to Server
- Remove `GOVERNOR_ADDRESS` from Frontend

None


packages/apps/human-app/server/src/modules/governance/governance.service.ts
```diff
@@ -0,0 +1,175 @@
...
+import { ProposalResponse } from './model/governance.model';
+
+@Injectable()
+export class GovernanceService {
+  constructor(
+    private readonly configService: EnvironmentConfigService,
+    @Inject(CACHE_MANAGER) private readonly cacheManager: Cache,
+  ) {}
+
+  public async getProposals(): Promise<ProposalResponse[]> {
```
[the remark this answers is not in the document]
flopez7: Done it in new commits. Btw, voteStart and voteEnd contains timestamp, not blockNumber. We're using timestamp inside governance contracts

dnechay (commented):
Looks much better now!

1. please check [logger usage examples](https://github.com/humanprotocol/human-protocol/blob/develop/packages/libs/logger/README.md#usage-examples)
2. there is still a case w/ 2+ concurrent requests where you access cache from them and can get invalid state
3. one [comment from prev review](https://github.com/humanprotocol/human-protocol/pull/3509/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#r2281856310)

packages/apps/human-app/frontend/src/modules/governance-banner/components/governance-banner.tsx
```diff
@@ -40,15 +43,19 @@ export function GovernanceBanner() {
...
   if (!isVerificationCompleted || isLoading || isError || !data) {
     return null;
   }
 
-  const forVotes = parseFloat(data.forVotes) || 0;
-  const againstVotes = parseFloat(data.againstVotes) || 0;
-  const abstainVotes = parseFloat(data.abstainVotes) || 0;
+  const now = Math.floor(Date.now() / 1000);
+  const isPending = now < data.voteStart;
+  const isActive = data.voteStart <= now && now < data.voteEnd;
```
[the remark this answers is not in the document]
dnechay: @flopez7 the thing is that you calculate it in two different places: in `fetchProposal` and in the banner itself, so it is easy to miss it if this logic has to be updated, so would be nice to have it in a single place and options are:
- backend
- `fetchProposal` function, so you get data is as from BE and map it to have all extras before you use it (in filters below and in banner)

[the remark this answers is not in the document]
dnechay: @flopez7 added utils look good, however `getProposalStatus` is not a "time" util, would be nice to have another place for it

packages/apps/human-app/frontend/src/shared/utils/time.ts
```diff
@@ -0,0 +1,22 @@
+export type ProposalStatus = 'pending' | 'active';
+
+export function getProposalStatus(
```
dnechay, line 3: It's better to have this util in some other place because it doesn't have anything in common with `time`.

packages/apps/human-app/frontend/src/shared/utils/time.ts
```diff
@@ -0,0 +1,22 @@
...
+  voteStartMs: number,
+  voteEndMs: number,
+  now: number = Date.now()
```
dnechay, line 6:
```suggestion
  proposal: Proposal
```
declare `now` inside and use as `proposal.voteStart <= now`, because `now` is part of this function logic and you don't have to pass it as argument.

Also passing just a proposal is simpler for the caller of this util

packages/apps/human-app/server/src/common/utils/cache.ts
```diff
@@ -0,0 +1,3 @@
+export const generateCacheKey = (...parts: (string | number)[]): string => {
```
dnechay, line 1: Sorry for confusion in last comment: I meant that implementation is nice, didn't mean to move it to util (esp. taking into account that it can't be shared due to `governance` prefix, so let's please move it back

packages/apps/human-app/frontend/src/shared/utils/time.ts
```diff
@@ -0,0 +1,22 @@
...
+): ProposalStatus {
+  if (voteStartMs <= now && now < voteEndMs) return 'active';
+  return 'pending';
+}
+
+export function formatCountdown(diffMs: number): string {
```
dnechay, line 12: If you pass `targetDate` (as a `Date` or `ms`) - the called won't need to get `now` every time, so usage of this util would be cleaner

packages/apps/human-app/frontend/src/modules/governance-banner/components/governance-banner.tsx
```diff
@@ -5,51 +5,41 @@ import { useTranslation } from 'react-i18next';
...
 import { useColorMode } from '@/shared/contexts/color-mode';
 import { useWorkerIdentityVerificationStatus } from '@/modules/worker/profile/hooks';
-import { useActiveProposalQuery } from '../hooks/use-active-proposal-query';
+import { useProposalQuery } from '../hooks/use-proposal-query';
+import { formatCountdown, getProposalStatus } from '../../../shared/utils/time';
 
 export function GovernanceBanner() {
   const { t } = useTranslation();
-  const { data, isLoading, isError } = useActiveProposalQuery();
+  const { data, isLoading, isError } = useProposalQuery();
```
dnechay, line 13:
```suggestion
  const { data: proposal, isLoading, isError } = useProposalQuery();
```

dnechay (commented):
Looks way better and easier to understand now!

Some minor comments left.

Please also implement some safety-belt for `getProposals` to run only once in case of concurrent requests. Ideally - it should be a cron job to scan logs for new proposals and actualize data in cache and `getProposals` just to retrieve data from cache to return to the client

packages/apps/human-app/server/src/modules/governance/governance.service.ts
```diff
@@ -0,0 +1,181 @@
...
+        this.logger.warn('Failed to fetch state for proposal', {
+          error: err,
+          proposalId: proposal.proposalId,
+        });
+        continue;
+      }
+
+      if (state !== ProposalState.PENDING && state !== ProposalState.ACTIVE) {
+        continue;
+      }
```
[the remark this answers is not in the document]
flopez7: I'd keep the explicit comparisons here. They avoid creating a new array on each iteration and are a little faster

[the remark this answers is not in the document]
dnechay: @flopez7 It's believed that this kind of microoptimization is negligible in real apps that are not microbenchmarks: (it's going to be garbage-collected pretty quick; + such array is small, so creating it is lightweight in Node.js heap), so it's usually better to have a code with better readability.

If there was a heavy-loop, like with 1m iterations - then it would definitely make sense to have regular comparison for small number of values and for big number would be better to have a pre-created set of valid values and use `set.has`

Anyway, it is just a style suggestion, so up to you

packages/apps/human-app/frontend/src/shared/utils/time.ts
dnechay: @flopez7 actually it could be simple to just add proper status in `fetchProposal`, like:
```
const activeProposals = list.filter(
  (p) => p.voteStart <= now && now < p.voteEnd
);
if (activeProposals.length > 0) {
  const earliestEndingProposal = activeProposals.sort((a, b) => a.voteEnd - b.voteEnd)[0];
  return {
    ...earliestEndingProposal,
    status: 'active',
  };
}

const pendingProposals = list.filter((p) => now < p.voteStart);
if (pendingProposals.length > 0) {
  const earliestUpcomingProposal = activeProposals.sort((a, b) => a.voteEnd - b.voteEnd)[0];
  return {
    ...earliestUpcomingProposal,
    status: 'pending',
  };
}

return null;
```
So you don't even need this util

dnechay (commented):
Code lgtm

@flopez7 what about concurrent requests that will call `getProposals`?

flopez7:
Concurrent requests are fine. It’s read-only and cached; worst case we duplicate work and last write wins. No correctness risk (merge by proposalId, only PENDING/ACTIVE). We only advance lastScannedBlock after successful event fetch. Anyway, this is just an optimization, previously we did per-user requests on frontend and now it’s cached on server-side.

dnechay (approved):

PR #3524, opened 2025-08-27
 [Human App] Add caching for proposals endpoint and increase cache duration

flopez7:
## Issue tracking
N/A

## Context behind the change

- Add caching for proposals endpoint 
- Increase cache duration

## How has this been tested?
Deployed locally

## Release plan
None

None

### DIFF
--- a/packages/apps/human-app/server/src/modules/governance/governance.service.ts
+++ b/packages/apps/human-app/server/src/modules/governance/governance.service.ts
@@ -10,12 +10,10 @@ import { ProposalResponse } from './model/governance.model';
 
 const N_BLOCKS_LOOKBACK = 100000;
 
-type Proposal = {
-  proposalId: string;
-  voteStart: number;
-  voteEnd: number;
+type Block = {
+  timestamp: number;
+  number: number;
 };
-
 @Injectable()
 export class GovernanceService {
   private readonly logger = new Logger(GovernanceService.name);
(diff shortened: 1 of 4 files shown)


dnechay (commented):
Instead of caching using another key, we can change the logic a bit:
1. rename existing logic e.g. to `syncProposals`
2. change the logic to store just `keptProposals` in Redis
and then
```
let lastSyncedAt = 0;
getProposals() {
  if (Date.now() - this.configService.cacheTtlProposals > lastSyncedAt`) {
     await syncProposals();
     lastSyncedAt = Date.now();
  }
  const proposals = this.cacheManager.get('ketpProposals');
  return proposals;
}
```

packages/apps/human-app/server/src/modules/governance/governance.service.ts
```diff
@@ -10,12 +10,10 @@ import { ProposalResponse } from './model/governance.model';
...
-  voteStart: number;
-  voteEnd: number;
+type Block = {
+  timestamp: number;
+  number: number;
 };
-
 @Injectable()
 export class GovernanceService {
   private readonly logger = new Logger(GovernanceService.name);
```
dnechay, line 19: Sorry, didn't notice in original PR: we should be using our logger now
```suggestion
  private readonly logger = logger.child({ context: GovernanceService.name });
```
and
```
import logger from '../../logger';
```

It works with current implementation because we override nest logger, but the correct way is to use our custom one

packages/apps/human-app/server/src/modules/governance/governance.service.ts
```diff
@@ -34,48 +47,37 @@ export class GovernanceService {
...
-
-    const cachedLastScannedBlock =
-      (await this.cacheManager.get<number>(lastScannedBlockKey)) ?? 0;
+    const currentBlock = await provider.getBlock('latest');
+    if (!currentBlock) {
+      this.logger.error('No latest block from RPC');
+      throw new Error('Blockchain node unavailable');
+    }
 
     const fromBlock =
```
dnechay, line 56:
```suggestion
    const cachedLastScannedBlockNumber = cachedLastScannedBlock?.number ?? 0;
    const fromBlock = cachedLastScannedBlockNumber > 0
        ? cachedLastScannedBlockNumber + 1
        : Math.max(0, currentBlock.number - N_BLOCKS_LOOKBACK);
```

packages/apps/human-app/server/src/modules/governance/governance.service.ts
```diff
@@ -34,48 +47,37 @@ export class GovernanceService {
...
+        : currentBlock.number - N_BLOCKS_LOOKBACK;
 
-    let proposalList: Proposal[] = [];
+    let allProposals: ProposalResponse[] = [];
     try {
       const newProposals = await this.getProposalCreatedEvents(
         contract,
         fromBlock,
-        currentBlock,
+        currentBlock.number ?? 0,
```
dnechay, line 66:
```suggestion
        currentBlock.number,
```
Shouldn't number always be here if block is retrieved above?

dnechay (commented):
LGTM overall, few minor comments

dnechay (approved):

Issue #3516, opened 2025-08-25
[SDK] Fixed Gas Price on Aurora Networks


portuu3:
**Description**
Currently, the HUMAN Protocol SDK relies on the default provider gas estimation when interacting with Aurora (both mainnet and testnet). However, Aurora operates with a fixed gas price of 0.07GWei, which makes gas estimation unnecessary and their RPCs do not support it.

We should add some kind of a middleware that ensures a fixed gas price is always used when sending transactions through the SDK on Aurora networks.

**Motivation**
Avoid extra gas cost on transactions

**Implementation details** 
- Detect if the current chainId corresponds to Aurora mainnet (1313161554) or Aurora testnet (1313161555).
- Automatically set the gas price to 0.07Gwei before sending the transaction.
- Expose a configuration option to override the fixed gas price if needed.

PR #3526, opened 2025-08-28
[SDK] Fixed Gas Price on Aurora Networks

flopez7:
## Issue tracking
#3516

## Context behind the change
Enhance transaction handling by applying specific defaults for Aurora network

## How has this been tested?
Launched some transactions with both SDKs and checked gas price

## Release plan
Deploy new SDK version

Check if gas is being properly applied to Aurora network


packages/sdk/typescript/human-protocol-sdk/src/base.ts
```diff
@@ -21,4 +23,19 @@ export abstract class BaseEthersClient {
...
+
+  /**
+   * Internal helper to enrich transaction overrides with network specific defaults
+   * (e.g. fixed gas price for Aurora networks) while preserving user provided
+   * fee parameters.
+   */
+  protected applyTxDefaults(txOptions: Overrides = {}): Overrides {
+    if (
+      this.networkData.chainId === ChainId.AURORA_TESTNET &&
+      txOptions.gasPrice === undefined
```
Dzeranov, line 35: A couple questions:
1. Are we planning to include only aurora testnet support?
2. I'm not sure that we need to check if the gasPrice is undefined here. If I remember correctly, every transaction on aurora will cost 0.07 gwei so I don't think we need to give other users to set the gasPrice. In case it's lower that 0.07 transaction might fail, if it's higher than it's an unnecessary spending. @portuu3 please correct me if I'm wrong.

portuu3: 1. Aurora mainnet is not in the SDK yet, we will add it later when we deploye the contracts in mainnet
2. You are right. Actually if the user wants to spend is up to him, but also makes sense to just force the static value

packages/sdk/python/human-protocol-sdk/human_protocol_sdk/utils.py
```diff
@@ -319,3 +324,25 @@ def validate_json(data: str) -> bool:
...
+    :return: Mutated tx options with enforced defaults
+    """
+    opts: TxParams = dict(tx_options) if tx_options else {}
+    try:
+        if w3.eth.chain_id == ChainId.AURORA_TESTNET.value:
+            opts["gasPrice"] = DEFAULT_AURORA_GAS_PRICE
+            opts.pop("maxFeePerGas", None)
+            opts.pop("maxPriorityFeePerGas", None)
+    except Exception:
+        pass
```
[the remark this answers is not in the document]
flopez7: You're right, removed the try/except because the only thing that could throw there was reading chain_id, and that’s already validated when client are created

packages/sdk/python/human-protocol-sdk/example_kvstore_set.py
Dzeranov, line 1: @flopez7 is this file included by a mistake?

Issue #3483, opened 2025-07-29
[SDK] spike: find a better way of publishing docs to GitBook


dnechay:
**Is your feature request related to a problem? Please describe.**
Atm when publishing new SDK version we have to create a new branch that corresponds to version tag, create a new "doc collection" in GitBook space, then fully sync GitHub repo from the created branch to created collection and set it to latest version. This process is not only doesn't have any instruction, but also implicit by it's nature, seems to be heavy-weight and also pollutes our repository with unnecessary branches (we don't use them for overall releases, i.e. we don't checkout from these branches for hotfixes on old versions, we use just to publish docs).

**Describe the solution you'd like**
The expected solution should be simpler. Ideally syncing version tag and docs, but not sure it's possible, so this spike is to investigate possible solutions and then we pick one that is a good fit.

**Describe alternatives you've considered**
TBD by the spike engineer

**Additional context**
Might make sense to work on this together w/ https://github.com/humanprotocol/human-protocol/issues/3424 as it's kinda related

dnechay:
@portuu3 
> Just I wonder if it can be placed inside SDK

What exactly you mean? Docs folder? If so, I suppose it should be possible, because you simply provide a path to directory for GitBook, so it can be placed anywhere

> Also with UI you cannot choose a tag as ref, we should check if it is possible via API
That's true it's not possible via UI, but it's possible via API: I've already done it for `4.3.0` release (you can check that we have docs, but there is no branch), the format is next (`POST: https://api.gitbook.com/v1/spaces/space_id/git/import`):
```json
{
    "url": "https://github.com/humanprotocol/human-protocol",
    "ref": "refs/tags/sdk@4.3.0",
    "repoProjectDirectory": "./"
}
```
and include api key (generated on UI) in "bearer" auth

portuu3:
@dnechay 
> What exactly you mean? Docs folder? If so, I suppose it should be possible, because you simply provide a path to directory for GitBook, so it can be placed anywhere
I mean inside SDK site as new version of the docs and also set the new version as default.

dnechay:
[...]

Issue #3424, opened 2025-06-27
[SDK] Change the mechanism for publishing new package version


dnechay:
**Is your feature request related to a problem? Please describe.**
When `sdk` is published - it has `human-protocol/core` version specified as `x` (analogue of `*`) which creates different issues like:
- when you install some specific version of `sdk`, e.g. 3.0.8, you still get latest version of core, not one that is supposed to work
- once you have `sdk` installed, `core` package dependency is "locked" and not getting updated when you bump `sdk`

In order to solve the above issues, we have to use the exact version of `core` package in `sdk`, which means changing `"@human-protocol/core": "workspace:x"` to `"@human-protocol/core": "workspace:*"` in `package.json` of `sdk`, so when `yarn pack` the package it will resolve workspace reference to exact `core` version from its `package.json` (check [docs for details](https://yarnpkg.com/features/workspaces#cross-references)). In order to implement such change - we have to keep `version` field in `package.json` updated for all publishable packages, but such change is going to break the CI/CD process we have now.

**Describe the solution you'd like**
Right now our CD process for `human-protocol/core` and `human-protocol/sdk` process looks like:
1. Implement changes and merge them as PRs to `develop`
2. Create release PR and merge it to `main`
3. Create a new GitHub Release w/ proper version tag that triggers `core` and `sdk` to be published with the same version.

The proposal is to change it to the next one:
1. Implement changes and merge them as PRs to `develop`
2. Have a release PR (as we do now)
3. When preparing a release, revisit all changes that are made for all  publishable packages and bump `version` field in their `package.json` via separate PR (so we can review if the version is correct)
4. Merge updated version to `develop`, then merge release PR to `main`
5. Once you have packages with proper versions merged to `main` branch - it should trigger GitHub action that will publish all necessary packages in the proper order (e.g. `sdk` depends on `core`, so we have to make sure that `core` is published first`)

Such workflow should be retryable, i.e. if check should present on "if version is already published/no changes - do nothing".

**Describe alternatives you've considered**
We considered using an automated tools, e.g. `semantic-release`, but agreed that it is an overhead for now and we can think about it in case we publish new package version more often and simple but a bit manual process become too complicated.

**Additional context**
PRs that fixed broken dependency resolution and have some more details: https://github.com/humanprotocol/human-protocol/pull/3419, https://github.com/humanprotocol/human-protocol/pull/3422

PR #3539, opened 2025-09-03
feat: introduce changesets for package publishing

dnechay:
## Issue tracking
#3424 

### What's in
1. Refactored monorepo to have `eslint` and `prettier` per app/lib, so we have cleaner config files for these tools, ability to refactor them separately and can have some root-level scripts that are not affected by root-level lint/prettier. Also removed global `jest` config because it's not used (note: `.gitignore` changes in apps related to it because `.ignore` config of tools relies on it). Resulting configs are the same except for `gcv` app that has a small change in prettier related to `trailingComma`
2. Updated `yarn` version to latest stable in order to have some latest pieces that are necessary for publishing process (`--json` flag in `yarn npm publish` + fixed `pre-` version; some other items you can find in their changelog)
3. Added `.local` to `.gitignore`, so developers can have their `.local` folder with any stuff for local tests. Also added `package.tgz` to avoid potential commit of packed packages
4. Set `npmPublishAccess` to `public` by default since we publish our packages with that
5. Added `prepack` lifecycle hook to packages, so they are built automatically when `publish` is triggered. Also updated `build` command to have `clean` included to avoid potential garbage when doing builds on local (you can still rely on incremental TS builds if build by `tsc`, not via `package.json` scripts)
6. Actualized existing packages versions to what we have published in npm
7. Introduced `changesets` for version management + new GitHub Action to publish packages and create tags.

### Drill down into the new publish process
The whole idea as same as in original issue, only version management relies on `changesets` tooling to simplify process for us.
> [!WARNING]
> We rely on `changesets` only to easily track changes and properly set new versions. We still do publishing relying on `yarn workspaces`  we have and not on some fancy tooling provided by `changesets` only because [under the hood](https://github.com/changesets/changesets/blob/ca0be26207d74e40abe9edc82219a66a2d09853c/packages/cli/src/commands/publish/npm-utils.ts#L68-L88) it runs `npm publish` instead of `yarn npm publish` and because of that `.yarnrc` is not respected + it causes some other issues with published packages (e.g. like mentioned in [this issue](https://github.com/changesets/changesets/issues/1454))

As part of a new process, engineers will have to create `changeset` file as part of their PR in case they have some changes for publishable packages. Before somebody opens a PR - they should run `yarn changeset` command in their branch from the repo root, select packages with changes and what version part (major, minior, patch) expected for bump according to their changes. This will generate a changeset file in `.changeset` folder that should be part of PR and also reviewed and merged to `develop`.

Let's loot at an example

**I. Work on packages changes w/o worrying about exact version number, only about [semver](https://semver.org/) "weight"**

1. Engineer A does a `patch` version change for `sdk`, then merges PR to `develop`. Nothing happens at this point
2. Engineer B does `minor` change for `logger` package and merges PR to `develop`. Nothing happens at this point
3. Engineer C changes both `sdk` and `logger` with `patch`, merges PR to `develop`. Nothing happens at this point

**II. Pre-leases: do actual versions update**

When we are ready to release, engineer that facilitates the release will need to open a new PR from `develop` to do a version bump. To do that - run `yarn changeset version` from the project root, commit changes and open a PR. Then we review that new versions are correct and merge it to `develop`

> [!NOTE]
> In our example we will get a `patch` version change for `sdk` because all changes from last version are `patch` (i.e. `1.2.3` -> `1.2.4`); For logger - it's going to be a `minor` version bump (i.e. `1.2.3` -> `1.3.0`), because `patch` can be a part of `minor` version, but not vice-versa and all changes go in a batch for next release.

**III. Publish packages with proper versions**
There is a GitHub action that will automatically publish all packages that have new versions. It's triggered by pushed on `main` branch, so in order to release packages - just merge `develop` into `main` with new proper versions set up.
This action will also create git tags for published packages, such as `@human-protocol/logger@1.1.0`. 

> [!IMPORTANT]
> Since we rely on `yarn workspaces` to publish packages and have proper package versions in `package.json` - `yarn` will resolve `worksapce:` links for us correctly while doing publishing in topological order, e.g. `core` will be published first, then `sdk`

## How has this been tested?
- [x] run `yarn changeset` to get package changes tracked
- [x] run `yarn changeset version` to make sure versions updated correctly
- [x] run `publish` command from GitHub Action to make sure packages are built, packed and can be published
- [x] grabbed and example log file from publishing a synthetic package to personal registry and tested tagging script on it
- [x] called `yarn pack` w/ and w/o `SKIP_PREPACK` to make sure `prepack` script works and `SKIP_PREPACK` is respected
- [x] run at least on service (used RepO) that relies on all packages in docker and use healthcheck to verify it works

## Release plan
1. Make sure that there are no pending packages for release (because once we get new flow merged to `develop` - new github action will be in place)
2. Merge this PR and release it separately in order to e2e test the new flow.

- it might be blocking for some packages that have "pending" release (i.e. changes merged to develop, but not published yet)
- GitHub action is not tested, because we have to put it into `main` for it, so some things might be failing and we will have to fix them separately

## What's next
- we might want to set up some GitHub action that will be releasing `alpha/beta` versions with tag `next` instead of `latest` from the specific branch you provide in input arguments
- as part of #3483 , we can have another GitHub action that will be triggered on new tag creation and will sync changes to GitBook: it will need to create a new space for new sdk version + trigger sync from git tag
- discuss possibility to also commit artifacts of `core` package build using [Git LFS](https://git-lfs.com/), in order to speed up builds in CI/CD systems


.changeset/config.json
```diff
@@ -0,0 +1,13 @@
+{
```
dnechay, line 1: Ref on config: https://github.com/changesets/changesets/blob/main/docs/config-file-options.md

dnechay:
Let's hold the merge till we have all the pending items released

PR #3569, opened 2025-09-18
Invalid property name in API validation messages

### LINKED ISSUE #3381: Invalid property name in API validation messages
**Describe the bug**
All our APIs have snake case schema, but everywhere in the code we use camel case, and when performing API schema validation - casing is already transformed, so validation messages contain camel-cased values instead of snake-case, which is confusing

**To Reproduce**
1. Go to any app and find an endpoint that uses some compound param name (e.g. `escrow_address`)
2. Send a request to such endpoint with invalid data
3. Check the validation error message

**Expected behavior**
API schema to be consistent in params names

**Screenshots**
Image

**Note**
Probably the only option to fix that is to remove case transformer for incoming request data and do such transformations in controllers.

flopez7:
## Issue tracking
Close #3381

## Context behind the change
- Refactor validation error handling to include detailed error messages for all oracles 
- Improve class validation response structure for JL, Exchange Oracle and Reputation Oracle

## How has this been tested?
Deployed locally and made some calls to make sure all error messages are in snake_case

## Release plan
Deploy again all services

None

### DIFF
--- a/packages/apps/reputation-oracle/server/src/common/pipes/validation.ts
+++ b/packages/apps/reputation-oracle/server/src/common/pipes/validation.ts
@@ -7,7 +7,7 @@ import {
   type ValidationPipeOptions,
 } from '@nestjs/common';
 
-import { camelToSnake } from '../../utils/case-converters';
+import { camelToSnake } from '@/utils/case-converters';
 
 @Injectable()
 export class HttpValidationPipe extends ValidationPipe {
@@ -27,20 +27,12 @@ export class HttpValidationPipe extends ValidationPipe {
   }
 
   private formatErrorsSnakeCase(errors: ValidationError[]): string[] {
-    const rawMessages = this.flattenValidationErrors(errors);
-    return rawMessages.map((msg) => {
-      const firstSpace = msg.indexOf(' ');
-      if (firstSpace === -1) return msg;
-
-      const rawPath = msg.slice(0, firstSpace);
-      const rest = msg.slice(firstSpace + 1);
-
-      const transformedPath = rawPath
-        .split('.')
-        .map((segment) => segment.replace(/^[A-Za-z0-9_]+/, camelToSnake))
-        .join('.');
-
-      return `${transformedPath} ${rest}`;
-    });
+    return errors
+      .flatMap((error) => this.mapChildrenToValidationErrors(error))
+      .flatMap((error) =>
+        Object.values(error.constraints || {}).map((msg) =>
+          msg.replace(error.property, camelToSnake(error.property)),
+        ),
+      );
   }
 }
--- a/packages/apps/reputation-oracle/server/src/common/pipes/validation.ts
+++ b/packages/apps/reputation-oracle/server/src/common/pipes/validation.ts
@@ -7,6 +7,8 @@ import {
   type ValidationPipeOptions,
 } from '@nestjs/common';
 
+import { camelToSnake } from '../../utils/case-converters';
+
 @Injectable()
 export class HttpValidationPipe extends ValidationPipe {
   constructor(options?: ValidationPipeOptions) {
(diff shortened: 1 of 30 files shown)


packages/apps/reputation-oracle/server/src/common/pipes/validation.ts
```diff
@@ -24,4 +23,31 @@ export class HttpValidationPipe extends ValidationPipe {
       ...options,
     });
   }
+
+  private formatErrorsSnakeCase(
```
dnechay, line 27: Can we use `[mapChildrenToValidationErrors](https://github.com/nestjs/nest/blob/9bff3c98e49a256b37cbb520676cd0deec444487/packages/common/pipes/validation.pipe.ts#L291C13-L291C42)` to flatten errors and handle only case changes on our end?
It would help to avoid potential bugs if something changes in nest and also make code more readable

portuu3:
> In genera lgtm, but I would like to discuss something first before we move on with it.
> 
> Suggested approach is based on the fact that `class-validator` library returns errors in format like
> 
> ```
> $property must <constraint details>
> ```
> 
> ([docs ref](https://github.com/typestack/class-validator?tab=readme-ov-file#validation-errors)) and we assume that the very first token (aka "word" in sentence) in the error message (or earch `constraint` of `ValidationError`) is the property name, so we take it and change its casing.
> 
> However, I'm not entirely sure that:
> 
> 1. library has this format for all possible decorators (even though it should be 99% true)
> 2. library won't ever change it
>    Even though 2 above are not highly likely to happen and lead to invalid error messages again, there is another case that takes place atm.
> 
> It's possible to provide custom error message to any decorator provided by the library + when you write your custom decorators. In this case format might be completely different from default and we have to remember to stick to it in order to avoid getting weird errors messages: we get `_provided password must be at least 8 characters long` right now with two issues:
> 
> 1. first word in error message had capital letter, so incorrectly got `_` at the beginning
> 2. `password` property name is not on the first place, so its casing won't be changed
> 
> I carefully checked `class-validator` documentation and code of `ValidationPipe` and don't think we can put something into our validation pipe that will always transform property names correctly, it's possible to do only if we stick to specific message format.
> 
> So I see 3 possible options on how we deal with it:
> 
> 1. everywhere we stick to the error message format like `$property <error details>`, so suggested might be applied (however, we will have to remember about it and take it into account in code reviews)
> 2. for camelCased properties we already have to mention proper `name` in the `ApiProperty` decorator so it's correct in swagger (there is no way to do it automatically), so we could also write custom error messages in place for it, w/o a need to do any extra transform in validation pipe
> 3. we remove case transformers and use our "controller layer" to map request/response data accordingly; this way we don't have to remember about naming in swagger or about error messages at all and have proper separation of concerns (e.g. if we change API schema - we don't have to change service classes as it would be needed now because we pass DTOs directly)
> 
> I personally prefer №3 because it's straightforward, clean and comes with 0 extra maintenance cost (we faced some issues already on converting case for `request.query` after upgrading express because it's read-only object now), but we will have to plan this work as a separate refactoring then
> 
> wdyt @portuu3 @Dzeranov @flopez7

I agree that option 3 might be the best approach. But I wouldn't take care of it now if the solution implemented in this PR is working for now. @dnechay

Dzeranov:
@dnechay I agree with 3rd approach because we won't have to maintain a bit weird looking code.
However, I suggest we come back to it later because implementing it will take some time and we can't afford to do that now.

dnechay:
@flopez7 could you please check for custom message then, get rid of them, so we can expect this common format of error messages?

Then you can probably use [mapChildrenToValidationErrors](https://github.com/nestjs/nest/blob/9bff3c98e49a256b37cbb520676cd0deec444487/packages/common/pipes/validation.pipe.ts#L291C13-L291C42) to unwrap all errors and replace property with something like:
```
error.constraints.map((c) => c.replace(error.property, camelToSnake(error.property)))
```

It should be good for long and later we can change to №3 when we have lull in tasks

dnechay (dismissed):
LGTM

One missing error message in RepO in `IsValidPassword` ([ref](https://github.com/humanprotocol/human-protocol/blob/66a95e61dda1a0235c4058c23da997eaf3b2aba5/packages/apps/reputation-oracle/server/src/common/validators/index.ts#L46)). It can be refactored to `@Length(8, 128)` instead of `MinLength` and `MaxLength`

Other than that - all nice

PR #3586, opened 2025-09-29
Enhance validation for HTTP urls

### LINKED ISSUE #3275: Invalid validation for HTTP urls
**Describe the bug**
Right now in different places we rely on `new URL(str)` constructor in order to validate if some string is valid HTTP URL, but it's not 100% correct

**To Reproduce**
In any place where we validate http urls, try to paste `http:invalid` or e.g. `http://test` and you will see you can pass validation

**Expected behavior**
Should return `false` for valid URI that is not valid HTTP URL.

**Suggested solution**
Either replace it with some library that does good validation (e.g. [validatorjs](https://github.com/validatorjs/validator.js)) or if we don't need to be 100% strict we can add simple domain validation for http-protocols/

flopez7:
## Issue tracking
Closes #3275

## Context behind the change
Add validator package and enhance URL validation logic for JL, Reputation Oracle and SDK

## How has this been tested?
Ran unit tests with different url test cases to check if everything worked as expected

## Release plan
- Create new minor change for SDK
- Deploy services

Check that the URLs that we're currently using are not failing with new validations.

### DIFF
--- a/packages/sdk/typescript/human-protocol-sdk/src/utils.ts
+++ b/packages/sdk/typescript/human-protocol-sdk/src/utils.ts
@@ -47,14 +47,10 @@ export const throwError = (e: any) => {
  * @returns
  */
 export const isValidUrl = (url: string): boolean => {
-  if (typeof url !== 'string' || url.trim() === '') return false;
-
-  const localhostRegex = /^https?:\/\/localhost(?::\d{1,5})?(?:\/|$)/i;
-  if (localhostRegex.test(url)) return true;
-
   return isURL(url, {
     require_protocol: true,
     protocols: ['http', 'https'],
+    require_tld: false,
   });
 };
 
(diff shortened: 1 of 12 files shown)


packages/sdk/typescript/human-protocol-sdk/src/utils.ts
```diff
@@ -45,13 +46,16 @@ export const throwError = (e: any) => {
...
-    new URL(url);
-    return true;
-  } catch (err) {
-    return false;
-  }
+export const isValidUrl = (url: string): boolean => {
+  if (typeof url !== 'string' || url.trim() === '') return false;
+
+  const localhostRegex = /^https?:\/\/localhost(?::\d{1,5})?(?:\/|$)/i;
+  if (localhostRegex.test(url)) return true;
```
dnechay, line 53: Let's remove this and use `require_tld: false` settings instead: when we run services in docker we use service names in URLs (e.g. `http://minio:9000/bucket_name`. `http://reputation-oracle:5000/health/ping), so if we don't accept these - it will be broken

dnechay (commented):
Let's add unit tests in SDK, RepO & JL that:
- will ensure we do not support invalid URL anymore
- help to ensure that we keep supporting `localhost` and similar local aliases like docker service names (`http://<service>:<port>`

dnechay (approved):
lgtm

Nice work on improving unit tests coverage for sdk utils

PR #3598, opened 2025-10-07
[SDK] publish docs to GitBook

### LINKED ISSUE #3483: [SDK] spike: find a better way of publishing docs to GitBook

flopez7:
## Issue tracking
Close #3483

## Context behind the change
- Add GitHub Actions workflow 
- Add script for publishing SDK documentation to GitBook
- Unify the first page of documents

## How has this been tested?
Launched several times via act 

## Release plan
Add some github secrets:
`GITBOOK_API_TOKEN=
GITBOOK_ORG_ID=
GITBOOK_PARENT_SPACE_ID=
GITBOOK_SITE_ID=
MAKE_DEFAULT=true`

None

### DIFF
--- a/docs/sdk/README.md
+++ b/docs/sdk/README.md
@@ -33,6 +33,6 @@ pip install human-protocol-sdk
 
 - Content is imported from the repository at release time.
 
-## Changelock
+## Changelog
 
 - [CHANGELOG](changelog.md) — release notes and changes.
--- a/docs/sdk/README.md
+++ b/docs/sdk/README.md
@@ -1,0 +1,38 @@
+# HUMAN Protocol SDK
+
+Welcome to the official documentation for HUMAN Protocol SDK.
+
+## Overview
+
+- Covers both TypeScript and Python SDKs used to build on HUMAN Protocol.
+- Use the sidebar to browse guides and API references; search to quickly find classes and functions.
+
+## Install
+
+### Typescript
+
+```bash
+npm install @human-protocol/sdk
+# or
+yarn add @human-protocol/sdk
+```
+
+### Python
+
+```bash
+pip install human-protocol-sdk
+```
+
+## Links
+
+- [NPM package](https://www.npmjs.com/package/@human-protocol/sdk)
+- [PyPI package](https://pypi.org/project/human-protocol-sdk/)
+- [GitHub repository](https://github.com/humanprotocol/human-protocol)
+
+## Versioning
+
+- Content is imported from the repository at release time.
+
+## Changelock
+
+- [CHANGELOG](changelog.md) — release notes and changes.
--- a/scripts/gitbook-sync-sdk.mjs
+++ b/scripts/gitbook-sync-sdk.mjs
@@ -12,7 +12,7 @@ const {
   GITBOOK_SITE_ID,
   GIT_REF,
   SPACE_VISIBILITY = "public",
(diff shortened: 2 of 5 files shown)


docs/sdk/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+- [NPM package](https://www.npmjs.com/package/@human-protocol/sdk)
+- [PyPI package](https://pypi.org/project/human-protocol-sdk/)
+- [GitHub repository](https://github.com/humanprotocol/human-protocol)
+
+## Versioning
+
+- Content is imported from the repository at release time.
+
+## Changelock
```
dnechay, line 36:
```suggestion
## Changelog
```

scripts/gitbook-sync-sdk.mjs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const {
+  GITBOOK_API_TOKEN,
+  GITBOOK_ORG_ID,
+  PARENT_SPACE_ID,
+  GITBOOK_SITE_ID,
+  GIT_REF,
+  SPACE_VISIBILITY = "public",
+  SPACE_ICON_URL = "https://firebasestorage.googleapis.com/v0/b/gitbook-x-prod.appspot.com/o/spaces%2FULrmzCkJk7WEl6FnhbTZ%2Ficon%2FvYf8cnyycn97DBMRLNJr%2Fhp-sdk-docs-logo.png?alt=media&token=24fd7851-0806-4528-b912-e714c43e5152",
```
dnechay, line 15: Just in case: maybe use base64 version of it? Using token in the url doesn't look sage

scripts/gitbook-sync-sdk.mjs
```diff
@@ -0,0 +1,115 @@
...
+      return JSON.parse(text);
+    } catch (_) {
+      return { status: res.status, body: text };
+    }
+  }
+  return { status: res.status, body: text };
+};
+
+async function createSpace() {
+  const body = { title: `v${version}`, visibility: SPACE_VISIBILITY };
```
dnechay, line 58: Should it be always public?

```suggestion
  const body = { title: `v${version}`, visibility: 'public' };
```

If by some reason we want it configurable - add `SPACE_VISIBILITY` to release plan

dnechay (commented):
LGTM overall
Couple suggestions to discuss

scripts/gitbook-sync-sdk.mjs
flopez7: Gitbook API expects a URI and has length limits, and base64 can easily exceed them if you want to use any other image. This way, using the URL, we wouldn't have problem

scripts/gitbook-sync-sdk.mjs
flopez7: For example, I used it for some tests. It's added as public by default. That's why I didn't add it to release plan

dnechay (approved):

dnechay:
Nice work

Issue #3619, opened 2025-10-15
[Docs] Staking on Human App


Dzeranov:
**Description**
Create a document describing required changes for Human App/Reputation Oracle to implement staking requirement on Human App

**Context**
1. Jobs discovery should be unavailable unless user has staked on either CEX or staking contract (list of supported exchanges will be defined later).
2. Users should be able to securely store their read-only api keys for CEX (need to think whether it's needed to store more than 1 exchange)
3. Extra rewards for users who staked more than X amount
4. Staking requirement should be 'pluggable' (so we could enable/disable it at any time without changing the codebase)

Issue #3560, opened 2025-09-15
Write a staking tutorial


portuu3:
Write a staking tutorial in docs.humanprotocol.org at `HUMAN Tech Docs/Tutorials/Staking`

PR #3632, opened 2025-10-24
Integrate staking amount requirement

### LINKED ISSUE #3618: [Human App] Require staking for accessing jobs on Human App
**Description**
Need to implement a mechanism that will prevent users from participating in jobs on Human App unless they've staked at least 1k HMT on either staking contract or some CEX.
As the next step users should be rewarded if they've staked more than X amount.

**Implementation details**
1. Tech doc with required changes
2. TBD based on 1
(see sub-issues for more details)

flopez7:
## Issue tracking
Close #3618, #3560, #3619, #3633

## Context behind the change
Reputation Oracle:
- Added ExchangeApiKeysModule for managing exchange API keys.
- Added validation for exchange names using ExchangeNameValidator.
- Integrated AES encryption for storing API keys securely.
- Implemented exchange-specific clients (GateExchangeClient, MexcExchangeClient) for interacting with respective APIs.
- Enhanced AuthService to check staking eligibility based on exchange balances.
- Added configuration services for encryption and staking parameters.
- Created migration for the exchange_api_keys table in the database.

Human App:
- Added new stake_allowed property to JWT
- Check in all the necessary controllers if we're allowed to interact based on this stake property.

## How has this been tested?
In progress

## Release plan
Add new `AES_ENCRYPTION_KEY` and `STAKING_ELIGIBILITY_ENABLED` in Reputation Oracle
Add `VITE_STAKING_DASHBOARD_URL` in Human App FE

Not yet defined

### DIFF
--- a/packages/apps/human-app/frontend/src/modules/worker/profile/components/staking-info.tsx
+++ b/packages/apps/human-app/frontend/src/modules/worker/profile/components/staking-info.tsx
@@ -31,7 +31,7 @@ export function StakingInfo() {
     refetch,
     isRefetching,
   } = useGetStakingSummary();
-  const { data: uiConfig } = useGetUiConfig();
+  const { data: uiConfig, isLoading: isUiConfigLoading } = useGetUiConfig();
 
   const isConnectButtonDisabled =
     !!exchangeApiKeyData?.exchange_name || isExchangeApiKeyLoading;
@@ -46,7 +46,7 @@ export function StakingInfo() {
     isError;
 
   const isStaked =
-    isLoading || isStakingError
+    isLoading || isStakingError || isUiConfigLoading
       ? false
       : stakedAmount >= Number(uiConfig?.minThreshold || '0');
 


packages/apps/reputation-oracle/server/src/modules/encryption/fixtures/index.ts
```diff
@@ -0,0 +1,4 @@
+export const mockEncryptionConfigService = {
+  // 32-byte key for AES-256-GCM tests
+  aesEncryptionKey: '12345678901234567890123456789012',
```
dnechay:
```suggestion
  aesEncryptionKey: faker.string.sample(32),
```

packages/apps/reputation-oracle/server/src/modules/exchange-api-keys/exchange-api-key.entity.ts
```diff
@@ -0,0 +1,43 @@
...
+  @BeforeInsert()
+  protected beforeInsert(): void {
+    this.createdAt = new Date();
+    this.updatedAt = this.createdAt;
+  }
+
+  @BeforeUpdate()
+  protected beforeUpdate(): void {
+    this.updatedAt = new Date();
+  }
```
dnechay:
```suggestion
```
These two are part of `BaseEntity` here

packages/apps/reputation-oracle/server/src/modules/exchange/mexc-exchange.service.ts
```diff
@@ -0,0 +1,109 @@
...
+    const controller = new AbortController();
+    const timeout = setTimeout(
+      () => controller.abort(),
+      this.stakingConfigService.timeoutMs,
+    );
+    try {
+      const res = await fetch(url, {
+        method: 'GET',
+        headers: { 'X-MEXC-APIKEY': creds.apiKey },
+        signal: controller.signal,
```
dnechay: You can use just `AbortSignal.timeout(ms)`

packages/apps/reputation-oracle/server/src/modules/exchange/mexc-exchange.service.ts
```diff
@@ -0,0 +1,109 @@
...
+      } as RequestInit);
+      return res.ok;
+    } catch (error) {
```
dnechay: Probably we should distinguish somehow if we got access error or some sort of network error, so we can distinguish between those and let the end use know

packages/apps/reputation-oracle/server/src/modules/exchange/gate-exchange.service.ts
```diff
@@ -0,0 +1,175 @@
...
+  return createHmac('sha512', secret).update(payload).digest('hex');
+}
+
+@Injectable()
+export class GateExchangeService extends ExchangeService {
+  readonly id = 'gate' as const;
+  private readonly logger = logger.child({ context: GateExchangeService.name });
+  private readonly apiBaseUrl =
+    process.env.GATE_API_BASE_URL && process.env.GATE_API_BASE_URL.length > 0
+      ? process.env.GATE_API_BASE_URL
```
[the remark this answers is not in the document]
flopez7: Was an error, it wasn't an env

packages/apps/reputation-oracle/server/src/modules/exchange-api-keys/exchange-api-keys.controller.ts
```diff
@@ -0,0 +1,119 @@
...
+      allOf: [{ $ref: getSchemaPath(EnrolledApiKeyDto) }],
+    },
+  })
+  @Get('/')
+  async retrieveEnrolledApiKeys(
+    @Req() request: RequestWithUser,
+  ): Promise<EnrolledApiKeyDto | null> {
+    const userId = request.user.id;
+
+    return this.exchangeApiKeysService.retrievedEnrolledApiKey(userId);
```
dnechay: nit: maybe 404 in case if not found?

packages/apps/reputation-oracle/server/src/modules/exchange-api-keys/exchange-api-keys.service.ts
```diff
@@ -0,0 +1,117 @@
...
+    ]);
+
+    return {
+      exchangeName: entity.exchangeName,
+      apiKey: decryptedApiKey.toString(),
+      secretKey: decryptedSecretKey.toString(),
+    };
+  }
+
+  async retrievedEnrolledApiKey(
```
dnechay: nit:
It's the same as `retrieved`, but just doesn't return `secretKey` because used in controller. So maybe use `retrieve` there and and just delete `secretKey` before sending a response?

packages/apps/reputation-oracle/server/src/modules/encryption/aes-encryption.service.ts
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+type EncryptionOutput = {
+  encrypted: Buffer;
+  iv: Buffer;
+  authTag: Buffer;
+};
+
+@Injectable()
+export class AesEncryptionService {
```
dnechay: Please add unit tests for this service

packages/apps/human-app/server/src/modules/oracle-discovery/oracle-discovery.controller.ts
```diff
@@ -22,6 +27,7 @@ import Environment from '../../common/utils/environment';
 import { ChainId } from '@human-protocol/sdk';
 
 @ApiTags('Oracle-Discovery')
+@ApiBearerAuth()
```
dnechay: Good catch

packages/apps/reputation-oracle/server/src/modules/exchange/mexc-exchange.client.ts
```diff
@@ -0,0 +1,98 @@
...
+      statusText: res.statusText,
+    });
+    return false;
+  }
+
+  async getAccountBalance(asset: string): Promise<number> {
+    const path = '/account';
+    const timestamp = Date.now();
+    const query = `timestamp=${timestamp}&recvWindow=${this.recvWindow}`;
+    const signature = this.signQuery(query);
```
dnechay: nit: this all part with query signature can be DRYed as well

packages/apps/reputation-oracle/server/src/modules/exchange/mexc-exchange.client.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    const res = await fetchWithHandling(
+      this.id,
+      url,
+      { 'X-MEXC-APIKEY': this.apiKey },
+      this.logger,
+      this.timeoutMs,
+    );
+    if (!res.ok) {
+      this.logger.warn('MEXC balance fetch failed', {
```
dnechay:
```suggestion
      this.logger.debug('MEXC balance fetch failed', {
```
I'm not sure that we want to log any non-2xx code.
Let's log only errors in `fetchWithHandling` (so we get major code errors or sensible network errors only), if somebody doesn't have proper access or exchange is 500/502 it will be just noise for us if we log it other than debug

dnechay (dismissed):

packages/apps/reputation-oracle/server/src/modules/auth/auth.service.ts
```diff
@@ -265,6 +277,51 @@ export class AuthService {
...
+      inspectedStakeAmount < this.stakingConfigService.minThreshold &&
+      userEntity.evmAddress
+    ) {
+      const onChainStake = await this.web3Service.getStakedBalance(
+        userEntity.evmAddress,
+      );
+      inspectedStakeAmount += onChainStake;
+    }
+
+    return inspectedStakeAmount >= this.stakingConfigService.minThreshold;
```
dnechay: @flopez7 
If there is a network error/hiccup, then `inspectedStakeAmount` might be 0 and user on UI will see that they don't have correct stake.

Would it be better to fail auth in this case? Or maybe have some other way to handle that taking into account that info is written in JWT and also used by oracles

packages/apps/reputation-oracle/server/src/modules/exchange-api-keys/exchange-api-keys.module.ts
```diff
@@ -0,0 +1,18 @@
+import { Module } from '@nestjs/common';
+
+import { EncryptionModule } from '@/modules/encryption';
+import { ExchangeModule } from '@/modules/exchange/exchange.module';
+import { Web3Module } from '@/modules/web3';
+
+import { ExchangeApiKeysController } from './exchange-api-keys.controller';
+import { ExchangeApiKeysRepository } from './exchange-api-keys.repository';
+import { ExchangeApiKeysService } from './exchange-api-keys.service';
+import { UserModule } from '../user';
```
dnechay:
```suggestion
import { UserModule } from '@/modules/user';
```

packages/apps/reputation-oracle/server/src/modules/staking/staking.service.ts
```diff
@@ -0,0 +1,112 @@
...
+        userId,
+        error,
+      });
+    }
+
+    if (user.evmAddress) {
+      try {
+        summary.onChainStake = (
+          await this.getOnChainStakedBalance(user.evmAddress)
+        ).toLocaleString(undefined, {
```
dnechay: Why not `ethers.formatUnits`?

packages/apps/reputation-oracle/server/src/modules/staking/staking.service.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+  const actual = jest.requireActual('@human-protocol/sdk');
+  return {
+    ...actual,
+    StakingClient: {
+      build: jest.fn(),
+    },
+  };
+});
+
+const { StakingClient } = jest.requireMock('@human-protocol/sdk');
```
dnechay: Since you already mocked it - you can import it directly as
```
import { StakingClient } from '@human-protocol/sdk';
```
and then below
```
const mockedStakingClient = jest.mocked(StakingClient);
```
so you can use typed `mockedStakingClient` everywhere you need to mock staking client behavior

You can check for examples in some other modules we have

packages/apps/reputation-oracle/server/src/modules/staking/staking.service.ts
flopez7: The value returned by getOnChainStakedBalance is already in ETH. I only need to control how it’s displayed by avoiding scientific notation like 1e-18 and showing the full decimal string

packages/apps/reputation-oracle/server/src/utils/stake.ts
```diff
@@ -0,0 +1,6 @@
+export const formatStake = (value: number) =>
+  new Intl.NumberFormat(undefined, {
+    maximumFractionDigits: 18,
```
portuu3: this might not be always 18 for some exchanges

portuu3 (changes requested):

dnechay (approved):
LGTM

packages/apps/reputation-oracle/server/src/modules/staking/staking.service.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    const summary: StakeSummaryData = {
+      exchangeStake: '0',
+      onChainStake: '0',
+      minThreshold: this.stakingConfigService.minThreshold.toString(),
+    };
+
+    try {
+      summary.exchangeStake = formatStake(
```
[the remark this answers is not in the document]
flopez7: I think it’s better to send it already formatted. That way, if any other service needs this value, it can use it as is without having to handle the formatting again

dnechay (approved):

docker-setup/docker-compose.dev.yml
```diff
@@ -91,7 +91,7 @@ services:
       timeout: 5s
       retries: 5
     volumes:
-      - graph-node-db-data:/var/lib/postgresql/data:Z
+      - graph-node-db-data:/var/lib/postgresql:Z
```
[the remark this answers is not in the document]
flopez7: Good catch, fix pushed

dnechay (dismissed):

PR #3648, opened 2025-11-04
Create changeset for python sdk and bump packages versions

portuu3:
## Issue tracking
NA

## Context behind the change
- Create changeset for python sdk and bump packages versions
- Bump packages versions
- Update subgraph ids for mainnets


dnechay (approved):
Do we need to also re-generate the docs or they already re-generated on develop?

PR #3690, opened 2025-11-21
Fix cancelation refunds

portuu3:
## Issue tracking
NA

## Context behind the change
Cancelation refund events where not properly tracked in Job Launcher because when the cancelation happen via Cancel() (escrow contract) it didn't emit CancelationRefund event.

## How has this been tested?

- [x] Local tests
- [x] Testnet

## Release plan
- [x] Upgrade testnet contracts
- [x] Upgrade mainnet contracts

NA

### DIFF
--- a/.changeset/tasty-shrimps-grow.md
+++ b/.changeset/tasty-shrimps-grow.md
@@ -1,0 +1,5 @@
+---
+"@human-protocol/core": patch
+---
+
+Add CancelationRefund event to any cancelation
--- a/packages/core/.openzeppelin/bsc-testnet.json
+++ b/packages/core/.openzeppelin/bsc-testnet.json
@@ -2956,6 +2956,163 @@
         },
         "namespaces": {}
       }
+    },
+    "addfe6d6ad3f78f6ccf914624fcf7eba6c970f5ba3a20a52f0184991b3d172e1": {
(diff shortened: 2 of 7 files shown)


packages/apps/job-launcher/server/src/modules/job/job.service.ts
```diff
@@ -923,24 +925,96 @@ export class JobService {
...
+    const provider = signer.provider;
+    const contract: Escrow = Escrow__factory.connect(escrowAddress!, provider);
+    const fromBlock = 79278120; //This issue started at this block
+    const toBlock = 'latest';
+    const cancelledFilter = contract.filters?.Cancelled?.();
+    const cancelledLogs = await contract.queryFilter(
+      cancelledFilter,
+      fromBlock,
+      toBlock,
+    );
```
dnechay: You can just double-check contract status and:
- throw if it's not `cancelled`
- if it's `cancelled` - query `Transfer` events and then compare transfer data with launcher address

That would make implementation much simpler, because now it's hard to get why you need to filter cancellation and use it as ref logs

Dzeranov (approved):

dnechay (approved):

dnechay (dismissed):

dnechay (approved):

PR #3722, opened 2025-12-18
refactor: dashboard chains usage and HUMAN App errors

dnechay:
## Issue tracking
Freestyle

## Context behind the change
1. Got alert from DD about unhandled exception and 500 error when checking operator stats and operator doesn't have url set in KV store. Changed the logic to properly handle that in order to return correct status code
2. Got alert from DD about `ERR_INVALID_URL`, that happens in case if somebody tries to use chain id that is not configured (e.g. 1338 - localhost). Addressed that by refactoring the way of working with chain ids, so only expected values allowed, not every chain supported by SDK
3. Similar to above: got alert on chain id usage on JL and fixed it

## How has this been tested?
- [x] lint & test for human app passes
- [x] run dashboard locally, hit every changed endpoint with testnet data via swagger, make sure data is returned 

## Release plan
1. Merge & deploy
2. Remove `WEB3_ENV` var for dashboard server

Should be none

### DIFF
--- a/packages/apps/dashboard/server/src/common/constants/chains.ts
+++ b/packages/apps/dashboard/server/src/common/constants/chains.ts
@@ -2,18 +2,24 @@ import { ChainId as SdkChainId } from '@human-protocol/sdk';
 
 import Environment from '../utils/environment';
 
-export enum ProductionChainId {
-  POLYGON_MAINNET = SdkChainId.POLYGON,
-  BSC_MAINNET = SdkChainId.BSC_MAINNET,
-  ETHEREUM = SdkChainId.MAINNET,
-}
+export const ProductionChainId = {
+  POLYGON_MAINNET: SdkChainId.POLYGON,
+  BSC_MAINNET: SdkChainId.BSC_MAINNET,
+  ETHEREUM: SdkChainId.MAINNET,
+} as const satisfies Record<string, SdkChainId>;
 
-export enum DevelopmentChainId {
-  POLYGON_AMOY = SdkChainId.POLYGON_AMOY,
-  BSC_TESTNET = SdkChainId.BSC_TESTNET,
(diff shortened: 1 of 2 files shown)


packages/apps/dashboard/server/src/common/constants/chains.ts
```diff
@@ -0,0 +1,22 @@
...
+  BSC_MAINNET = SdkChainId.BSC_MAINNET,
+  ETHEREUM = SdkChainId.MAINNET,
+}
+
+export enum DevelopmentChainId {
+  POLYGON_AMOY = SdkChainId.POLYGON_AMOY,
+  BSC_TESTNET = SdkChainId.BSC_TESTNET,
+  SEPOLIA = SdkChainId.SEPOLIA,
+  LOCALHOST = SdkChainId.LOCALHOST,
+}
```
flopez7:
```suggestion
export const ProductionChainId = {
  POLYGON_MAINNET: SdkChainId.POLYGON,
  BSC_MAINNET: SdkChainId.BSC_MAINNET,
  ETHEREUM: SdkChainId.MAINNET,
} as const satisfies Record<string, SdkChainId>;

export type ProductionChainId =
  (typeof ProductionChainId)[keyof typeof ProductionChainId];

export const DevelopmentChainId = {
  POLYGON_AMOY: SdkChainId.POLYGON_AMOY,
  BSC_TESTNET: SdkChainId.BSC_TESTNET,
  SEPOLIA: SdkChainId.SEPOLIA,
  LOCALHOST: SdkChainId.LOCALHOST,
} as const satisfies Record<string, SdkChainId>;

export type DevelopmentChainId =
  (typeof DevelopmentChainId)[keyof typeof DevelopmentChainId];
```

We can define objects as const instead of enum. This way it uses SdkChainId internally, so we don't have to specify `chainId as number` on each SDK call

Issue #3716, opened 2025-12-15
[SDK] Add ability to specify `timeout` for transactions wait


dnechay:
**Is your feature request related to a problem? Please describe.**
Sometimes transactions can stuck in the pool and `wait` operation keeps open connection until apps is shut down. In order to gracefully handle timeout-cancellations, would be good to have an ability to pass `timeout` param to `tx.wait` (docs [ref](https://docs.ethers.org/v6/api/contract/#ContractTransactionResponse-wait)).

**Describe the solution you'd like**
Accept `timeout` option (same format as in `ethers` - milliseconds) in method options where we pass gas price, nonce, and etc. Also consider option of adding `confirms` params as well

**Describe alternatives you've considered**
Using `Promise.race` - not enough because it just throws, original `wait` promise still runs in the background.

**Additional context**
N/A

PR #3744, opened 2026-01-21
[SDK] Enhance transaction handling with wait options

flopez7:
## Issue tracking
#3716 

## Context behind the change
- Updated EscrowClient, KVStoreClient and StakingClient to utilize TransactionOverrides for transaction options, allowing for customizable wait parameters (confirmations and timeout).
- Refactored methods to ensure proper handling of transaction confirmations and timeouts during contract interactions.
- Added tests to verify that wait options are correctly passed to transaction wait.

## How has this been tested?
Created some transactions manually with different configurations to make it fail.
Ran unit tests.

## Release plan
Deploy new SDK version. 

Python SDK has been significantly changed to introduce confirmation logic. If we don't specify confirmations, it should work as usual.

### DIFF
--- a/packages/sdk/typescript/human-protocol-sdk/src/escrow.ts
+++ b/packages/sdk/typescript/human-protocol-sdk/src/escrow.ts
@@ -217,7 +217,7 @@ export class EscrowClient extends BaseEthersClient {
     }
 
     try {
-      const result = await this.sendTransaction(
+      const result = await this.sendTxAndWait(
         (overrides) =>
           this.escrowFactoryContract.createEscrow(
             tokenAddress,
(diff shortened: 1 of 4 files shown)


packages/sdk/typescript/human-protocol-sdk/src/escrow.ts
```diff
@@ -1098,7 +1110,7 @@ export class EscrowClient extends BaseEthersClient {
     finalResultsHash: string,
     payoutId: string,
     forceComplete = false,
-    txOptions: Overrides = {}
+    txOptions: TransactionOverrides = {}
```
dnechay: nit: we don't send tx in this method
```suggestion
    txOptions: Overrides = {}
```

Issue #3726, opened 2025-12-18
Publish releases on GitHub when new lib gets published


dnechay:
**Description**
Some time ago we changed the way how we do SDK releases. Since then we only push git tags attached to package version, but do not publish GitHub release. However we would like to have releases to be published as well.

**Motivation**
To have public release notes and use them e.g. for docs.

**Implementation details** 
Taking into account that we might be publishing different versions of npm packages in the same actions (hence having different tags for them) it's better to create new GitHub Action that will be triggered on tag creation and craft a release for this specific package version and tag.
To publish release - either need to use GitHub API directly or use action like [softprops/action-gh-release](https://github.com/marketplace/actions/gh-release) or similar

PR #3763, opened 2026-01-27
[CI] add GitHub Actions workflow for creating package releases

flopez7:
## Issue tracking
#3726

## Context behind the change
Created a new GitHub Actions workflow that runs whenever a JS or Python package tag is pushed, checks if a GitHub Release already exists for that tag, and if not automatically creates one using softprops/action-gh-release with auto‑generated release notes from recent commits

## How has this been tested?
Ran locally

## Release plan
None

None


dnechay:
@portuu3 taking into account that we prepare a version-release PR anyway that removes all changesets, crafting changelog file manually for each package won't be a big deal, but we can also try to enable automatic changelogs by `changesets`. Both work for me if we have separate changelogs that we can ref in action from this PR 👍🏻

Issue #3772, opened 2026-02-02
Switch monorepo to npm core package


flopez7:
**Description**
Update monorepo to consume `@human-protocol/core` from npm instead of `workspace:*`, and split contract deployment + release workflows from the current `cd-packages.yaml` so npm publishing (and its changesets) runs before merging to `develop`.

**Motivation**
Using the published npm package removes the per-build core compilation step, cutting build times (previously ~15 minutes) for several packages. We also keep the repo history intact and can still switch back to `workspace:*` when local contract changes are needed

**Implementation details** 

- Replace all `@human-protocol/core: workspace:*` dependencies across services/packages with the npm latest version.
- Decouple contract deployment and releases from `cd-packages.yaml` into a dedicated workflow that can run earlier.
- Create a separate `changeset` configuration for the core package release flow, and run it before changes land in `develop`.
- Ensure CI/tests use the published package by default, with an option to switch back to workspace:* for local dev.

PR #3773, opened 2026-02-04
Switch monorepo to npm core package

flopez7:
c## Issue tracking
#3772

## Context behind the change
This PR updates the monorepo so most packages consume `@human-protocol/core` from npm (instead of resolving it from the local workspace) to avoid rebuilding core on every CI/job. This removes a major build-time and keeps the repo history intact, while still allowing local core development when needed.

**Main changes**:

- Replaced workspace-based core resolution with npm resolution by updating consumers to use `@human-protocol/core: "npm:^5.0.1"`.
- Extracted core build invocations from workflows/scripts that were implicitly rebuilding core, so they now rely on the published npm package artifacts by default.
- Split release/deploy responsibilities: created a dedicated workflow for contract deployments/releases (decoupled from the old `cd-packages.yaml` flow), so publishing and deployment steps can run earlier and independently.
- Created a separate Changesets configuration for the core package release flow to support running core publishing before changes land in develop.

**Changesets notes**
To enable independent contract deployments, core needed its own Changesets flow. That introduced friction with Changesets dependency validation:
- Keeping all dependents with a normal range like `^5.0.1` caused Changesets errors when the root Changesets config ignored `@human-protocol/core` but non-ignored packages (SDK) depend on it.
- Also tried excluding core from Yarn workspaces to prevent local resolution; however, to make Changesets fully “independent” in that mode, core effectively needs to live in a separate repo (Changesets’ model and validation assumes a single, coherent workspace graph).
- Final approach: keep SDK depending on `@human-protocol/core` via `npm:^5.0.1` and the rest of the packages using `^5.0.1`. This produces a warning during yarn changeset / yarn changeset version, and when running changeset version dependency ranges get updated for most packages but not for the SDK package (because of how Changesets applies dependency updates/validation in this split setup).

## How has this been tested?
Deployed all services and verified dependency resolution behavior: consumers resolve `@human-protocol/core` from npm (not local) under the default Yarn config.
Generated some changesets for testing purpose. 
Built all docker images.

## Release plan
- Modify all packages deployment on render to use the new focused commands
-Core publishing is handled via the core-specific Changesets flow/workflow and should run before merging to develop when a core release is required
- Contract deployments now run via the dedicated deploy workflow (decoupled from the old combined pipeline), so deployments can be triggered independently of general package publishing.
- If local contract changes are required for development, switch back to workspace resolution and run the core-local scripts as needed.
- Set up [trusted publishing](https://docs.npmjs.com/trusted-publishers) for this action

Check services build time to make sure it is decreased considerably. 
Check that all actions works fine.

### DIFF
--- a/.github/workflows/ci-test-subgraph.yaml
+++ b/.github/workflows/ci-test-subgraph.yaml
@@ -23,6 +23,6 @@ jobs:
       - name: Generate manifest for Polygon for tests
         run: NETWORK=polygon yarn workspace @tools/subgraph generate
       - name: Build packages (scoped)
-        run: yarn workspaces foreach -Rpt --from @tools/subgraph run build
+        run: yarn workspaces foreach -Rpt --from @tools/subgraph --topological-dev run build
       - name: Run subgraph test
         run: yarn workspace @tools/subgraph test


packages/sdk/python/human-protocol-sdk/scripts/build-contracts.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 set -eux
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../../../../.." && pwd)"
+
 rm -rf artifacts
-cp -r ../../../core/artifacts .
+cp -r "${REPO_ROOT}/node_modules/@human-protocol/core/artifacts" .
```
dnechay: Taking into account that python sdk depends on js package and has its own `package.json`, maybe it makes sense to install core for it? This way when building python sdk we can also install updated local core using `workspaces` protocol just for python sdk when needed

Currently this `cp` line looks very implicit

packages/apps/human-app/server/src/main.ts
```diff
@@ -8,6 +8,23 @@ import { Cache } from 'cache-manager';
 import { AppModule } from './app.module';
 import { EnvironmentConfigService } from './common/config/environment-config.service';
 import logger, { nestLoggerOverride } from './logger';
+import fs from 'node:fs';
+import path from 'node:path';
+
+const entry = require.resolve('@human-protocol/core/typechain-types');
+let dir = path.dirname(entry);
+
+while (!fs.existsSync(path.join(dir, 'package.json'))) {
```
dnechay: Suppose we can clean it up

PR #3782, opened 2026-02-11
[SDK] Handle subgraph errors properly 

### LINKED ISSUE #3745: Exclude noisy subgraph errors from being logged as part of "unhandled exception" logging
**Is your feature request related to a problem? Please describe.**
In web services we have a global exception filter that logs "unhandled exception". When subgraph is laggy and throws "bad indexers" error - such errors are being logged as unhandled and hence we get alerted. We can try specifying some indexer that is more stable, but it can lead to the same situation, so better would be to just ignore such errors from being logged, so we aren't alerted with this noisy error.

Errors we noticed so far:
- bad indexers
- 521
- 502
- just a `fetch failed` from underlying `undici` client in graph sdk

**Describe the solution you'd like**
Ideally we should have some custom error class in SDK and throw it in case subgraph request fails (e.g. `SubgraphRequestFailedError`), then in apps we will be able to ignore the log and return 502 just by checking error instance type.

**Describe alternatives you've considered**
Considered handling errors on apps directly, but that's code duplication and breaks boundaries.

**Additional context**
- you can check Datadog logs e.g. for HUMAN Dashboard in order to get error examples
- it should be implemented in all JS-based web services

flopez7:
## Issue tracking
Close #3745

## Context behind the change
- Added typed subgraph errors in TS (`SubgraphRequestError`, `SubgraphBadIndexerError`)
- Updated exception filters to avoid logging bad indexers as “Unhandled exception”
- Other subgraph failures are still logged, but under a dedicated subgraph error path

## How has this been tested?
Ran unit tests

## Release plan

- Deploy new SDK version. 
- Add new monitor+alert on DataDog based on new logs

None

### DIFF
--- a/.changeset/crisp-buckets-agree.md
+++ b/.changeset/crisp-buckets-agree.md
@@ -1,0 +1,5 @@
+---
+"@human-protocol/sdk": minor
+---
+
+Added typed subgraph errors (SubgraphRequestError, SubgraphBadIndexerError) and wrapped subgraph request failures with these classes
--- a/packages/apps/dashboard/server/src/common/filters/exception.filter.ts
+++ b/packages/apps/dashboard/server/src/common/filters/exception.filter.ts
@@ -28,7 +28,7 @@ export class ExceptionFilter implements IExceptionFilter {
       status = HttpStatus.BAD_GATEWAY;
       responseBody.message = exception.message;
 
-      this.logger.error('Unhandled subgraph error', {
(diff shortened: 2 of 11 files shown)


dnechay (commented):
LGTM overall, one suggestion

Also please add to release plan that we need to add new monitor+alert on DataDog based on new logs

dnechay (dismissed):

portuu3 (approved):

PR #3854, opened 2026-03-31
fix axios to version 1.13.1

flopez7:
## Issue tracking
https://github.com/axios/axios/issues/10604

## Context behind the change
Temporarily pin axios to a known safe version at monorepo root to prevent accidental installation of the compromised v1.14.1 through direct or transitive dependencies. This is a precautionary change while the upstream security incident is being assessed

## How has this been tested?
Ran tests

## Release plan
None

None


flopez7:
Removed Axios v1.14.1 from npm

PR #3876, opened 2026-04-24
[Reputation Oracle] Mark escrow completion as completed before webhook creation

flopez7:
## Issue tracking
Freestyle

## Context behind the change
Persist escrow completion immediately after successful on-chain finalization, so webhook creation failures no longer leave the escrow stuck in an inconsistent DB state.

## How has this been tested?
Ran unit tests locally

## Release plan
None

None

### DIFF
--- a/packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
+++ b/packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
@@ -268,6 +268,9 @@ export class EscrowCompletionService {
           );
         }
 
+        escrowCompletionEntity.status = EscrowCompletionStatus.COMPLETED;
+        await this.escrowCompletionRepository.updateOne(escrowCompletionEntity);
+
         const oracleAddresses: string[] = [
           escrowData.launcher as string,
           escrowData.exchangeOracle as string,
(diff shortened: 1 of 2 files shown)


packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
```diff
@@ -282,19 +285,26 @@ export class EscrowCompletionService {
...
+              escrowCompletionEntityId: escrowCompletionEntity.id,
+              oracleAddress,
+            });
+            continue;
           }
 
           const { webhookUrl } = oracleData;
           if (!webhookUrl) {
-            throw new Error('Webhook url is no set for oracle');
+            this.logger.error('Webhook url is no set for oracle', {
```
portuu3:
```suggestion
            this.logger.error('Webhook url is not set for this oracle', {
```

portuu3 (changes requested):

dnechay (changes requested):
After this change, if webhook creation fails - it will never be retried, making the guarantee of webhook sending "at most once", but we need "at least once" in order to notify other oracle and let the whole flow finish.

If we fail to get operator data or fail to create webhook in DB (which can be only DB error), then we should fail and retry later. If oracle doesn't have webhook url set at all - we fail as well, but can consider a possibility of skipping webhook creation for such oracles

packages/apps/reputation-oracle/server/src/modules/escrow-completion/escrow-completion.service.ts
```diff
@@ -282,19 +285,26 @@ export class EscrowCompletionService {
...
         };
 
-        let allWebhooksCreated = true;
         for (const oracleAddress of oracleAddresses) {
           const oracleData = await OperatorUtils.getOperator(
             chainId,
             oracleAddress,
           );
           if (!oracleData) {
-            throw new Error('Oracle data is missing');
```
dnechay, line 292: It is mainly a safety belt for lagging subgraph, because it shouldn't be possible since at least fee and role must be set

portuu3:
> After this change, if webhook creation fails - it will never be retried, making the guarantee of webhook sending "at most once", but we need "at least once" in order to notify other oracle and let the whole flow finish.
> 
> If we fail to get operator data or fail to create webhook in DB (which can be only DB error), then we should fail and retry later. If oracle doesn't have webhook url set at all - we fail as well, but can consider a possibility of skipping webhook creation for such oracles


@dnechay  We cannot depend on other oracle's proper configuration, if they missconfigure their oracle is their problem and we should not have a wrong status in our database. I don't say we don't implement a retry mechanism, but if they do not configure it properly after some retries the status should change anyway, or we first change the status and just retry the webhook sending.

dnechay:
@portuu3 webhook sending is different process, here we only create one.

@flopez7 So taking all the above into account I would suggest to just skip webhook creation for oracles that don't have it set up and rely on existing retry mechanism in other cases

dnechay (approved):

portuu3 (approved):

PR #3971, opened 2026-07-14
Add an endpoint that accepts an arbitrary JSON manifest

portuu3:
## Issue tracking
NA

## Context behind the change
Add an endpoint that accepts an arbitrary JSON manifest

## How has this been tested?
Staging: https://amoy.polygonscan.com/address/0x31a219aeD1F5C51767c6F5d276e1A0C12Ea9051F

## Release plan
NA

NA

### DIFF
--- a/packages/apps/job-launcher/client/src/types/index.ts
+++ b/packages/apps/job-launcher/client/src/types/index.ts
@@ -107,11 +107,13 @@ export enum JobType {
 }
 
 export enum CvatJobType {
+  IMAGE_LABEL_BINARY = 'image_label_binary',
   IMAGE_POINTS = 'image_points',
   IMAGE_POLYGONS = 'image_polygons',
   IMAGE_BOXES = 'image_boxes',
   IMAGE_BOXES_FROM_POINTS = 'image_boxes_from_points',
   IMAGE_SKELETONS_FROM_BOXES = 'image_skeletons_from_boxes',
+  AUDIO_TRANSCRIPTION = 'audio_transcription',
 }
 
 export type FortuneRequest = {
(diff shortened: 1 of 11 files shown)


packages/apps/job-launcher/client/src/types/index.ts
```diff
@@ -107,11 +107,13 @@ export enum JobType {
 }
 
 export enum CvatJobType {
+  IMAGE_LABEL_BINARY = 'image_label_binary',
```
Dzeranov: We don't have `image_label_binary` so we can remove that.

packages/apps/job-launcher/server/src/modules/job/job.controller.ts
```diff
@@ -128,6 +129,42 @@ export class JobController {
...
+  })
+  @ApiResponse({
+    status: 400,
+    description: 'Bad Request. Invalid input parameters.',
+  })
+  @ApiResponse({
+    status: 401,
+    description: 'Unauthorized. Missing or invalid credentials.',
+  })
+  @Post('/unknown-manifest')
```
Dzeranov:
```suggestion
  @Post('/manifest-quick-launch')
```

I believe it's a bit better than 'unknown manifest'.

packages/apps/job-launcher/client/src/types/index.ts
Dzeranov: @portuu3

portuu3: @Dzeranov do you want to remove it from CVAT oracles also?

Dzeranov: In CVAT oracles it's just a leftover from one of the previous implementations so we can safely remove it from JL.

portuu3: It is ready then

Dzeranov (approved):

PR #3981, opened 2026-07-23
Update cvat payout calculator for the new manifest

portuu3:
## Issue tracking
NA

## Context behind the change
Update cvat payout calculator for the new manifest

## How has this been tested?
Tested using unit tests
Pending to test with CVAT oracles on testnet when ready

## Release plan
NA

NA

### DIFF
--- a/packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
+++ b/packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
@@ -60,14 +60,6 @@ describe('CvatPayoutsCalculator', () => {
       } as unknown as EscrowClient);
     });
 
-    afterEach(() => {
-      jest.resetAllMocks();
-      mockedEscrowClient.build.mockResolvedValue({
-        getIntermediateResultsUrl: mockedGetIntermediateResultsUrl,
-        getReservedFunds: mockedGetReservedFunds,
-      } as unknown as EscrowClient);
-    });
-
     it('throws when invalid annotation meta downloaded from valid url', async () => {
       const intermediateResultsUrl = faker.internet.url();
       mockedGetIntermediateResultsUrl.mockResolvedValueOnce(
@@ -102,7 +94,6 @@ describe('CvatPayoutsCalculator', () => {
       const jobsPerAnnotator = faker.number.int({ min: 1, max: 3 });
       const jobCount = jobsPerAnnotator * annotators.length;
       const payoutPerJob = BigInt(faker.number.int({ min: 1000, max: 100000 }));
-      const reservedFunds = payoutPerJob * BigInt(jobCount);
 
       const annotationsMeta: CvatAnnotationMeta = {
         jobs: Array.from({ length: jobCount }, (_v, index: number) => ({
@@ -139,12 +130,13 @@ describe('CvatPayoutsCalculator', () => {
         annotationsMeta,
       );
 
-      mockedGetReservedFunds.mockResolvedValueOnce(reservedFunds);
-      const manifest = generateCvatManifest();
+      mockedGetReservedFunds.mockResolvedValueOnce(
+        payoutPerJob * BigInt(jobCount),
+      );
       const payouts = await calculator.calculate({
         chainId,
         escrowAddress,
-        manifest: manifest,
+        manifest: generateCvatManifest(),
         finalResultsUrl: faker.internet.url(),
       });
 
@@ -245,12 +237,12 @@ describe('CvatPayoutsCalculator', () => {
         jobs: [
           {
             job_id: faker.number.int(),
-            final_result_id: faker.number.int(),
+            final_result_id: 1,
           },
         ],
         results: [
           {
-            id: faker.number.int(),
+            id: 2,
             job_id: faker.number.int(),
             annotator_wallet_address: faker.finance.ethereumAddress(),
             annotation_quality: faker.number.float(),
--- a/packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
+++ b/packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
@@ -4,7 +4,6 @@ import { faker } from '@faker-js/faker';
 import { createMock } from '@golevelup/ts-jest';
 import { EscrowClient } from '@human-protocol/sdk';
 import { Test } from '@nestjs/testing';
-import { ethers } from 'ethers';
 import _ from 'lodash';
 
 import { CvatAnnotationMeta } from '@/common/types';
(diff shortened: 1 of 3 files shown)


packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.ts
```diff
@@ -41,26 +39,28 @@ export class CvatPayoutsCalculator implements EscrowPayoutsCalculator {
...
+    if (!matchedJobResults.length) {
+      throw new Error('Invalid annotation meta');
+    }
+
+    const jobBountyValue = reservedFunds / BigInt(matchedJobResults.length);
+    const workersBounties = new Map<string, bigint>();
+
+    for (const jobFinalResult of matchedJobResults) {
       // TODO: enable annotation quality validation when ready
       if (
```
dnechay: Already filtered above, so probably we can remove this check on `jobFilanResult`

packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
```diff
@@ -48,14 +47,24 @@ describe('CvatPayoutsCalculator', () => {
...
+      } as unknown as EscrowClient);
+    });
+
+    afterEach(() => {
+      jest.resetAllMocks();
+      mockedEscrowClient.build.mockResolvedValue({
+        getIntermediateResultsUrl: mockedGetIntermediateResultsUrl,
+        getReservedFunds: mockedGetReservedFunds,
       } as unknown as EscrowClient);
     });
```
dnechay: No need in this block
```suggestion
```

packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
```diff
@@ -84,23 +93,22 @@ describe('CvatPayoutsCalculator', () => {
...
       const annotators = [
         faker.finance.ethereumAddress(),
         faker.finance.ethereumAddress(),
       ];
 
       const jobsPerAnnotator = faker.number.int({ min: 1, max: 3 });
-      const tokenDecimals = BigInt(6);
+      const jobCount = jobsPerAnnotator * annotators.length;
+      const payoutPerJob = BigInt(faker.number.int({ min: 1000, max: 100000 }));
+      const reservedFunds = payoutPerJob * BigInt(jobCount);
```
dnechay:
```suggestion
mockedGetReservedFunds.mockResolvedValueOnce(payoutPerJob * BigInt(jobCount));
```

packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
```diff
@@ -48,14 +47,24 @@ describe('CvatPayoutsCalculator', () => {
     const mockedGetIntermediateResultsUrl = jest
       .fn()
       .mockImplementation(async () => faker.internet.url());
-    const mockedGetTokenAddress = jest.fn().mockImplementation(async () => {
-      return faker.finance.ethereumAddress();
-    });
+    const mockedGetReservedFunds = jest
+      .fn()
+      .mockImplementation(async () =>
+        BigInt(faker.number.int({ min: 1000, max: 100000 })),
```
dnechay, line 53: Let's not mock default with some random number, because it can confuse in case of incorrect/failing test suite, it's better to mimic real behavior (i.e. funds not reserved) which returns 0n
```suggestion
        0n,
```

packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
```diff
@@ -130,26 +138,19 @@ describe('CvatPayoutsCalculator', () => {
...
 
+      mockedGetReservedFunds.mockResolvedValueOnce(reservedFunds);
+      const manifest = generateCvatManifest();
       const payouts = await calculator.calculate({
         chainId,
         escrowAddress,
-        manifest,
+        manifest: manifest,
         finalResultsUrl: faker.internet.url(),
       });
```
dnechay:
```suggestion
const payouts = await calculator.calculate({
chainId,
escrowAddress,
        manifest: generateCvatManifest(),
finalResultsUrl: faker.internet.url(),
});
```

packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
```diff
@@ -231,5 +239,33 @@ describe('CvatPayoutsCalculator', () => {
...
+        jobs: [
+          {
+            job_id: faker.number.int(),
+            final_result_id: faker.number.int(),
+          },
+        ],
+        results: [
+          {
+            id: faker.number.int(),
+            job_id: faker.number.int(),
```
dnechay, line 246: Even though we use faker there, there is a very little chance that `faker.number.int()` will generate `final_result_id = result.id`, so for the purpose of this test let's make sure somehow that those ids are 100% different

dnechay (commented):
Lgtm overall, some cleanup comments

packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
portuu3: I dont understand why you want to return 0 here. We use reserved funds to calculate the job bounty

dnechay: @portuu3 
Here you just declare a mock to be used later, so there can define actual mocked values.

By default it should return something, that will make test:
- pass, where reserved funds aren't necessary
- fail, where you have to override

Setting random value by default satisfies both, but in case something goes wrong - e.g. you forget to set `mockResolvedValueOnce` in the the test where it actually matters - you will get failed assertions with weird calculations of job bounty something like `expected 1234.34, got 45834.3`, but in case we set `0` as default, then it would be `expected 1234.34, got 0` because of the math, which makes it clear that we just forgot to mock reserved funds call return value

dnechay (dismissed):

Dzeranov (dismissed):

packages/apps/reputation-oracle/server/src/modules/escrow-completion/payouts-calculation/cvat-payouts-calculator.spec.ts
portuu3: @dnechay check again please

dnechay (approved):

PR #3988, opened 2026-07-29
[Job Launcher] fix: error loading fee with an invalid rpc

portuu3:
## Issue tracking
NA

## Context behind the change
The app used MetaMask’s SDK connector, which used Wagmi’s default Polygon Amoy RPC instead of the RPC configured in the MetaMask extension. That public RPC failed DNS/network requests, leaving jobLauncherFee undefined and causing the Decimal crash.
We switched to MetaMask’s injected browser provider and added loading guards before calculating the fee.

## How has this been tested?
Locally

## Release plan
NA

NA


dnechay (approved):
"Pay now" loading state doesn't work as expected, but not related to this PR

PR #3995, opened 2026-08-04
chore: bump aiohttp to 3.14.3 to patch security vulnerabilities

Dzeranov:
## Issue tracking

N/A
## Context behind the change

Bumped `aiohttp` package to `3.14.3` as per https://github.com/humanprotocol/human-protocol/actions/runs/30901488755/job/91966585685


zhiltsov-max:
The dependency has to be declared in `pyproject.toml` like "aiohttp>=xxx". Updating just the lockfile introduces the risk that the change will be just overwritten by the next automatic lockfile update.

zhiltsov-max (approved):

